PR #5, opened 2015-12-15
D.R.Y. Even in your example. No - especially in your example.

CodeRust:


gaffer01:
Thanks for your interest in the project and for your contribution. We're not yet in a position to accept pull requests as we need to sort out a 3rd party contributor license. Once that is done, and has been signed by contributors, we will be happy to consider pull requests.

However please note that this version of Gaffer (version 1) is no longer under active development. All effort is going into Gaffer 2, which although offering some of the same functionality, has a very different API and different base classes to this version of Gaffer. So it would be worth holding off making contributions to Gaffer 1 and instead waiting for Gaffer 2.

karabijavad:
@gaffer01 when is gaffer2 going to be available?

SamMorrowDrums:
@gaffer01 while the legal implications of having a "3rd party contributor license" that must be signed presumably makes bureaucratic sense internally to GCHQ, externally it is clunky and unfortunate. Is it not possible for you to safely and securely make code review of submitted PRs, and conduct appropriate testing and analysis of submissions, and have the correct licence information that ensures your outright ownership of all contributed code?

If you start this process by putting up barriers (however small) to contribution, the friction you create will ensure that eventually your open source projects will be replaced and ultimately forgotten by history, because open-sourcing in a productive manner is more than simply opening access to source code, and using Github as a release mechanism.

Many great open source projects have good developer guidelines that explain what you expect and demand for contribs to be considered, and if you were able to run it like that, with implicit agreement to the T&Cs of submission it would really make a substantial difference.

gaffer01:
@karabijavad Gaffer2 should be available within the next 2 or 3 weeks.

@SamMorrowDrums Thanks for your thoughts. Having a contributors license is fairly standard - see for example https://code.facebook.com/cla, https://azure.github.io/guidelines.html#cla. But we will consider your points further as we prepare the license.

PR #34, opened 2016-02-10
Adding operations as described in G31,

james010101101:
See Issue G31 for description also 
Removed provided scope from accumulo-core import in the accumulo-store
so the accumulo-store can be imported properly for use in client code.


p013570:
Any chance you could write some unit/integration tests for the new operations and/or their handlers?

accumulo-store/src/main/java/gaffer/accumulostore/utils/Constants.java
```diff
@@ -19,22 +19,25 @@
...
-    //Iterator options
+    // Iterator options
     public static final String VIEW = "View";
     public static final String DATA_SCHEMA = "Data_Schema";
     public static final String STORE_SCHEMA = "Store_Schema";
-    public static final String ENTITY_ONLY = "Entity_Only";
-    public static final String ALL_EDGE_ONLY = "All_Edges_Only";
+    public static final String INCLUDE_ENTITIES = "Include_All_Entities";
+    public static final String INCLUDE_ALL_EDGES = "Include_All_Edges";
+    public static final String NO_EDGES = "All_Edges_Only";
```
p013570: Perhaps All_Edges_Only should be No_Edges

accumulo-store/src/main/java/gaffer/accumulostore/utils/Constants.java
```diff
@@ -43,32 +46,57 @@
...
 
     // Iterator priorities
-    public static final int AGE_OFF_ITERATOR_PRIORITY = 10; // Applied during major compactions, minor compactions and scans.
-    public static final int AGGREGATOR_ITERATOR_PRIORITY = 20; // Applied during major compactions, minor compactions and scans.
-    public static final int BLOOM_FILTER_ITERATOR_PRIORITY = 31; // Applied only during scans.
-    public static final int EDGE_ENTITY_DIRECTED_FILTER_ITERATOR_PRIORITY = 32; // Applied only during scans.
-    public static final int ELEMENT_FILTER_ITERATOR_PRIORITY = 33; // Applied only during scans.
-    public static final int QUERY_TIME_AGGREGATOR_PRIORITY = 34; // Applied only during scans.
-    public static final int TRANSFORM_PRIORITY = 50; // Applied only during scans.
+    public static final int AGE_OFF_ITERATOR_PRIORITY = 10; // Applied during
```
p013570: Looks like your ide reformatted your comments here and formatted them strangely. Perhaps put the comment on 1 line or on the line before or as a javadoc

accumulo-store/src/main/java/gaffer/accumulostore/operation/handler/GetElementsInRangesHandler.java
```diff
@@ -0,0 +1,60 @@
...
+                                store),
+                        store.getKeyPackage().getIteratorFactory()
+                                .getEdgeEntityDirectionFilterIteratorSetting(operation),
+                        store.getKeyPackage().getIteratorFactory().getElementPropertyRangeQueryFilter(operation),
+                        store.getKeyPackage().getIteratorFactory().getQueryTimeAggregatorIteratorSetting(store));
+            } else {
+                ret = new AccumuloRangeIDRetriever(store, operation);
+            }
+        } catch (IteratorSettingException | StoreException e) {
+            throw new RuntimeException(e);
```
p013570: It might be better to throw an OperationException here. This also applies to the other accumulo operation handlers.

accumulo-store/src/main/java/gaffer/accumulostore/operation/impl/GetEdgesBetweenSets.java
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    @Override
+    public boolean isIncludeEntities() {
+        return false;
+    }
+
+    @Override
+    public void setIncludeEntities(final boolean includeEntities) {
+        if (includeEntities) {
```
p013570: To be consistent with how we manage the other fields that cannot be modified, it might be better to allow setIncludeEntities to be called when includeEntities is false.

accumulo-store/src/main/java/gaffer/accumulostore/utils/Constants.java
james010101101: Good spot

accumulo-store/src/test/java/gaffer/accumulostore/operation/handler/GetElementsBetweenSetsHandlerTest.java
```diff
@@ -0,0 +1,325 @@
...
+
+    }
+
+    @Test
+    public void testShouldSummarise() throws StoreException {
+        testShouldSummarise(byteEntityStore);
+        testShouldSummarise(gaffer1KeyStore);
+    }
+
+    public void testShouldSummarise(final AccumuloStore store) throws StoreException {
```
p013570: Could be private

accumulo-store/src/main/java/gaffer/accumulostore/utils/Constants.java
james010101101: Comments generally need updates/rewrites/reformatting, that will probably have to be done on another branch

accumulo-store/src/test/java/gaffer/accumulostore/operation/handler/GetElementsBetweenSetsHandlerTest.java
```diff
@@ -0,0 +1,325 @@
...
+        assertEquals(expectedResults, results);
+    }
+
+    @Test
+    public void testShouldHaveNoIncomingEdges() throws StoreException {
+        testShouldHaveNoIncomingEdges(byteEntityStore);
+        testShouldHaveNoIncomingEdges(gaffer1KeyStore);
+    }
+
+    public void testShouldHaveNoIncomingEdges(final AccumuloStore store) throws StoreException {
```
p013570: could be private

accumulo-store/src/test/java/gaffer/accumulostore/operation/handler/GetElementsBetweenSetsHandlerTest.java
```diff
@@ -0,0 +1,325 @@
...
+            assertTrue(results.contains(expectedResult));
+        }
+
+        //The result size should be 1
+        assertEquals(1, results.size());
+
+        assertEquals(expectedResults, results);
+    }
+
+    public void testShouldSummariseOutGoingEdgesOnly() throws StoreException {
```
p013570: I think this is method is missing the test annotation.

accumulo-store/src/test/java/gaffer/accumulostore/operation/handler/GetElementsWithinSetHandlerTest.java
```diff
@@ -0,0 +1,293 @@
...
+        for (Element expectedResult : expectedResults) {
+            assertTrue(results.contains(expectedResult));
+        }
+       
+        //The result size should be 2
+        assertEquals(2, results.size());
+        
+	}
+
+	private static void setupGraph(final AccumuloStore store, int numEntries) {
```
p013570: I don't think the parameter 'numEntries' is used in this method.

accumulo-store/src/main/java/gaffer/accumulostore/operation/impl/GetEdgesBetweenSets.java
james010101101: This is what this does?
If includeEntities is true here it throws an exception, otherwise the value is false which doesn't need to be assigned, as we always return false.

PR #36, opened 2016-02-10
Adding a maintenence jar to be built with the accumulo store, this jar

james010101101:
As issue G15


accumulo-store/src/main/java/gaffer/accumulostore/utils/AddUpdateTableIterator.java
```diff
@@ -17,14 +17,23 @@
...
 import org.apache.accumulo.core.client.IteratorSetting;
 import org.apache.accumulo.core.client.TableNotFoundException;
 import org.apache.accumulo.core.iterators.IteratorUtil.IteratorScope;
 
+import java.io.IOException;
+import java.nio.file.Path;
+import java.nio.file.Paths;
 import java.util.EnumSet;
 
 public class AddUpdateTableIterator {
```
p013570: It would be helpful to have a javadoc explaining that this class has a main method and when it should be used

Issue #37, opened 2016-02-10
Add Optional Parameter to Rest Service to restrict result size


james010101101:
The rest service doOperation/get calls call should have an optional parameter n, this will restrict the size of any returned iterable of results to n.

p013570:
Merged into develop

PR #55, opened 2016-02-17
gh-37

james010101101:
Adding n paramater to restrict the result set size of any /get operation
on the rest api


rest-api/src/main/java/gaffer/rest/service/SimpleOperationService.java
```diff
@@ -79,39 +82,51 @@ public void addElements(final AddElements operation) {
     }
 
     @Override
-    public Iterable<Element> getElementsBySeed(final GetElementsSeed<ElementSeed, Element> operation) {
-        return execute(operation);
+    public Iterable<Element> getElementsBySeed(final GetElementsSeed<ElementSeed, Element> operation, final Integer n) {
+        Iterable<Element> elements = execute(operation);
```
p013570: I think you can do: return null != n ? Iterables.limit(elements, n) : elements

james010101101: Done in the executeGet method to avoid having to duplicate the tertiary operator on every Call.

PR #70, opened 2016-02-23
G17 accumulo validator filter

p013570:


p013570:
We may need to add an additional section to the data schema for extra input validation. This would allow additional validation (could be computationally expensive) to be carried out when elements are added. However this will require a substantial change to the code will should be investigated as a separate issue gh-72.

james010101101:
Needs an ability to re-add or update the iterator same as the aggregator;
Should have an option to pick if you want to add or update the aggregator or validator as the validator is optional and may not be on the table.

PR #80, opened 2016-02-29
G17 accumulo validator filter

p013570:


p013570:
Need to add integration tests to check age off works in accumulo

p013570:
This needs fully testing on an accumulo instance before being merged in.

a09631:
Pull request approved.  Will be fully tested before push to master.

Issue #78, opened 2016-02-29
Fails to load resources packaged in jars such as in the example


p013570:
It would be better to load the resources (such as dataSchema.json) as resource streams so that assembled jars (such as example simple query) can be run from the command line.

As part of the fix for this issue we also ensure all streams are closed.

p013570:
Merged into develop

PR #83, opened 2016-03-01
gh-78 - ensures streams are closed

p013570:


accumulo-store/src/main/java/gaffer/accumulostore/key/core/AbstractCoreKeyAccumuloElementConverter.java
```diff
@@ -139,8 +140,9 @@ public Properties getPropertiesFromValue(final String group, final Value value)
             return properties;
         }
         final MapWritable map = new MapWritable();
-        try {
-            map.readFields(new DataInputStream(new ByteArrayInputStream(value.get())));
+        try (final InputStream inStream = new ByteArrayInputStream(value.get());
+             final DataInputStream dataStream = new DataInputStream(inStream)) {
+            map.readFields(dataStream);
```
james010101101: Why not add a finally block to these try catches and close these resources there? In this case the streams aren't needed once the objects are created, this could be added in a few of the changed files.

Issue #53, opened 2016-02-16
The filter function 'not' flag does not work if used in the Data Schema


p013570:
This is due to filter functions being cloned before use and the not flag is not maintained when cloned.

p013570:
Probably best to remove the not flag from FilterFunction and create a Not filter function that wraps a FilterFunction and inverts the result.

p013570:
Migration steps:
The not flag on FilterFunction has been removed. So if you were using this in a data schema, store schema or view then you will need to wrap your function inside a Not filter function.

For example:

in json
```
"filter": {
  "functions": [
    {
      "function": {
        "class": "gaffer.function.simple.filter.Not",
        "function": {
          "class": "gaffer.function.simple.filter.IsEqual",
          "value": "user02"
        }
      },
      "selection": [
        {
          "key": "userId"
        }
      ]
    }
  ]
}
```

in java using a view builder
```
.filter(new ElementFilter.Builder()
  .select(Property.USER_ID)
  .execute(new Not(new IsEqual("user02")))
  .build())
```

p013570:
Merged into develop

PR #85, opened 2016-03-02
gh-53 - removed not flag and added Not filter function

p013570:


a09631:
Pull request approved.

Issue #92, opened 2016-03-03
Frequency map aggregator should return null if input is null


p013570:
In the _state method if frequencyMap is null, then it should return null. Otherwise null values are magically converted into new frequency maps.

p013570:
merged into develop

PR #95, opened 2016-03-03
gh-92 - aggregate functions now initialise to null

p013570:


a09631:
Pull request approved.

Issue #91, opened 2016-03-03
Aggregation should skip properties that don't exist


p013570:
It should also skip properties that are null - i.e. null values should be treated as though the property does not exist.

p013570:
merged into master

PR #96, opened 2016-03-03
gh-91 - added null check to aggregator selections

p013570:


a09631:
Pull request approved, will be merged when changes merged with develop.

PR #101, opened 2016-03-16
G46 validate views

p013570:


p013570:
It would be better if users don't have to specify element group properties and their classes. They should only need to specify any transient properties required, the other properties should be looked up in the schema.

p013570:
We need to consider changing the view so you only have to specify transient properties. Standard properties should be read from the schema.

gaffer-core/store/src/main/java/gaffer/store/schema/Schema.java
```diff
@@ -246,6 +289,15 @@ protected void addEntity(final String group, final SchemaEntityDefinition elemen
         super.addEntity(group, elementDef);
     }
 
+    @Override
+    public String toString() {
+        try {
+            return "Schema" + new String(toJson(true), "UTF-8");
```
a09631: If encoding is a project level constant, this needs to be extracted and made available project-wide.

gaffer-core/store/src/test/java/gaffer/store/schema/SchemaElementDefinitionValidatorTest.java
```diff
@@ -27,24 +27,192 @@
...
+        // Then
+        assertTrue(isValid);
+    }
 
     @Test
     public void shouldValidateAndReturnTrueWhenNoPropertiesAggregated() {
         // Given
         final SchemaElementDefinition elementDef = mock(SchemaElementDefinition.class);
         final SchemaElementDefinitionValidator validator = new SchemaElementDefinitionValidator();
+        ;
```
a09631: Remove if unused.

PR #108, opened 2016-03-21
G105 ensure accumulo store operations are json serialisable

james010101101:


james010101101:
Pull request should be to merge into develop not master.

PR #114, opened 2016-03-22
G76 add accumulo split table operation

james010101101:


accumulo-store/src/main/java/gaffer/accumulostore/operation/hdfs/handler/job/SampleDataForSplitPointsJobFactory.java
```diff
@@ -0,0 +1,115 @@
...
+import org.apache.hadoop.mapred.JobConf;
+import org.apache.hadoop.mapreduce.Job;
+import org.apache.hadoop.mapreduce.lib.output.SequenceFileOutputFormat;
+
+import java.io.IOException;
+
+public class SampleDataForSplitPointsJobFactory {
+
+    public static final String PROPORTION_TO_SAMPLE = "proportion_to_sample";
+    public static final String UTF_8_CHARSET = "UTF-8";
```
p013570: There is now a UTF-8 constant in CommonUtils.CommonConstants

Issue #98, opened 2016-03-16
Allow accumulo bulk import to be run in 2 steps


p013570:
The bulk import from HDFS for accumulo runs as 2 jobs: fetch from HDFS and import into accumulo.
If the second job fails it would be useful to be able to rerun the operation but skip the first step.

a09631:
Pull request approved and merged into develop.

james010101101:
# Migration steps

From now on when running a bulk import job, the provided failure directory must already be present, exist on the filesystem and the accumulo user must have read/write permissions on this folder.

PR #123, opened 2016-03-31
gh-98 Enable AccumuloStore Import step to be run in two parts

james010101101:
Added Option & AccumuloStoreConstant to skip import step on AddElementsFromHdfs for the accumulo-store
Added Operation to do the import of keys/values to Accumulo from a directory.
Moved failureDirectory path variable from MapReduceOperation to AddElementsFromHdfs & ImportKeyValueFiles.

### DIFF
--- a/accumulo-store/src/main/java/gaffer/accumulostore/operation/hdfs/handler/AddElementsFromHdfsHandler.java
+++ b/accumulo-store/src/main/java/gaffer/accumulostore/operation/hdfs/handler/AddElementsFromHdfsHandler.java
@@ -16,16 +16,15 @@
 
 package gaffer.accumulostore.operation.hdfs.handler;
 
-import org.apache.hadoop.util.ToolRunner;
-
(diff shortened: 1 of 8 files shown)


accumulo-store/src/main/java/gaffer/accumulostore/operation/hdfs/handler/AddElementsFromHdfsHandler.java
```diff
@@ -36,7 +35,10 @@ public Void doOperation(final AddElementsFromHdfs operation, final Store store)
 
     public void doOperation(final AddElementsFromHdfs operation, final AccumuloStore store) throws OperationException {
         fetchElements(operation, store);
-        importElements(operation, store);
+        String skipImport = operation.getOption(AccumuloStoreConstants.ADD_ELEMENTS_FROM_HDFS_SKIP_IMPORT);
```
a09631: Mark parameter final to remain consistent.

PR #126, opened 2016-04-07
G112 cache operation

p013570:


p013570:
Need to up documentation to state that the cache is maintained per single request only, it cannot be used across multiple separate operation requests. So, it must be used inside a single operation chain.

p013570:
Updated javadoc as per previous comment.

Issue #154, opened 2016-04-20
Register Operation Handlers in accumulostore


james010101101:
Handlers for the 
SplitTable, SampleDataForSplitPoints and ImportAccumuloKeyValueFiles

Operations are not registered in the accumulostore.

a09631:
Pull request approved and merged with develop.

PR #156, opened 2016-04-20
gh-154 Add handlers to the accumulostore for new operations

james010101101:


p013570:
We could really do with a test to check all the accumulo operations have registered handlers. 

We do a similar thing in test StoreTest.shouldCreateStoreWithValidSchemasAndRegisterOperations, however, this only tests the core operations have been registered. We need a similar method in AccumuloStoreTest, to check the extra operations.

Issue #155, opened 2016-04-20
AddElementsFromHdfs.Builder inputDirectory methods are set to protected


james010101101:
The builder methods should be public not protected.

p013570:
Merged into release-0.3.4

PR #158, opened 2016-04-20
gh-155 Set AddElementsFromHdfs builder methods to public

james010101101:


p013570:
We have had a few problems with builders now - perhaps we should add the following method to the OperationTest interface:
```java
@Test
void shouldBuildOpeartion();
```

Then we should write a test for all operation builders to check all the builder methods are available, public, return the correct builder and set the correct field on the operation.

accumulo-store/src/main/java/gaffer/accumulostore/operation/hdfs/impl/SampleDataForSplitPoints.java
```diff
@@ -116,17 +117,17 @@ public Builder proportionToSample(final float proportionToSample) {
         }
 
         @Override
-        protected Builder inputPaths(final List<Path> inputPaths) {
+        protected Builder inputPaths(final List<String> inputPaths) {
```
p013570: Should be changed to public

accumulo-store/src/main/java/gaffer/accumulostore/operation/hdfs/impl/SampleDataForSplitPoints.java
```diff
@@ -116,17 +117,17 @@ public Builder proportionToSample(final float proportionToSample) {
...
             return (Builder) super.inputPaths(inputPaths);
         }
 
         @Override
-        protected Builder addInputPaths(final List<Path> inputPaths) {
+        protected Builder addInputPaths(final List<String> inputPaths) {
```
p013570: Should be changed to public

accumulo-store/src/main/java/gaffer/accumulostore/operation/hdfs/impl/SampleDataForSplitPoints.java
```diff
@@ -116,17 +117,17 @@ public Builder proportionToSample(final float proportionToSample) {
...
             return (Builder) super.addInputPaths(inputPaths);
         }
 
         @Override
-        protected Builder addInputPath(final Path inputPath) {
+        protected Builder addInputPath(final String inputPath) {
```
p013570: Should be changed to public

james010101101:
Will be merged with pull request gh-161

PR #177, opened 2016-04-29
G146 Add methods to expose all supported Operations

a09631:


p013570:
The changes look good. We could also really do with similar methods in the REST API within GraphConfigurationService to allows users to call your new graph methods via the REST API. This can either be done as part of this issue or we can create a separate issue for the work.

PR #194, opened 2016-05-10
Gh 186 user auths

p013570:


gaffer-core/operation/src/main/java/gaffer/user/User.java
```diff
@@ -0,0 +1,149 @@
...
+            user.dataAuths.addAll(dataAuths);
+            return this;
+        }
+
+        public Builder dataAuth(final String dataAuth) {
+            user.addDataAuth(dataAuth);
+            return this;
+        }
+
+        public Builder lock() {
```
a09631: Should a user not be locked when it's built?

gaffer-integration-tests/src/test/java/gaffer/integration/AbstractStoreIT.java
```diff
@@ -289,4 +292,8 @@ protected void addToMap(final Edge element, final Map<EdgeSeed, Edge> edges) {
     protected void addToMap(final Entity element, final Map<EntitySeed, Entity> entities) {
         entities.put(ElementSeed.createSeed(element), element);
     }
+
+    protected User getUser() {
```
a09631: Do we not want to use this in addDefaultElements?

Issue #193, opened 2016-05-10
Creation of Gaffer Docker files


ng12developer:
A Docker version of Gaffer.

p013570:
Nearly done - just need to update the docker documentation and create some basic scripts to make building and running the containers easier.

p013570:
Some of the docker files still need improving but that will be done separately. Also we need to create a docker image for the rest api - this will be done in issue gh-261

p013570:
Merged into master in gaffer-docker repository.

PR #176, opened 2016-04-28
gh-168-Initial-Commit-for-Gaffer-Docker

ng10developer:
Unable to assign issue 168 
Implemented proposed fix

PR #195, opened 2016-05-10
gh-193-Commit of Gaffer Docker files

ng12developer:
New pull request for Docker, please ignore pull request 176.


p013570:
See comments from old pull request gh-176.

gaffer01:
Please remove the file LICENSE from all of the subfolders. Remove all the commented out lines from the file ssh_config (but leave the copyright and Apache license notice).

ng12developer:
LICENSES removed and ssh_config edited.

p013570:
In the last release of Gaffer there were a few modifications that mean unfortunately some changes need to be made to the docker scripts to get them working again with Gaffer 0.3.5.

- Java version now needs updating in the base docker script to 1.8.
- The gaffer examples have been refactored, so gaffer/setup_files/configure_store_prop.sh will need to be updated to use the correct path for the store properties.
- The test at the end of the guide will need to be updated to use the slightly modified film example.

Also, we could do with 'yum install -y tar' being added to the top of Dockerfile.centos6.gaffer and Dockerfile.hadoop.centos6.

ng12developer:
I will adapt the examples to get in line with 0.3.5.

Can I clarify that its is Java 1.8 that is needed on the base image? The Gaffer code base for 0.3.5 and the develop branch still reference 1.7 and the Gaffer image will inherit from the base. There is also an issue of not having tested the Hadoop version is fully compatible with 1.8, whose image also inherits from the base.

p013570:
Brilliant, thanks. 

The core framework of Gaffer still requires Java 1.7, however the TinkerPop implementation that was added as a module in 0.3.5 requires 1.8. So in order to build the entire project Java 1.8 is required. If the hadoop testing is a problem, for the time being you could leave Java at 1.7 and exclude the tinkerpop module from the maven build. If using maven with version > 3.2.1 then you can use:
```bash
mvn clean install -pl \!tinkerpop
```

ng12developer:
Adapted examples for version 0.3.5.
tinkerpop module excluded from maven build of master code base.
tar commands added.

PR #205, opened 2016-05-12
Gh 191 standardise format of accumulo store unit tests

a09631:


accumulo-store/src/test/java/gaffer/accumulostore/key/core/impl/AbstractAccumuloElementConverterTest.java
```diff
@@ -93,24 +108,34 @@ public void TestAccumuloKeyConverterCFCQPropertydEdge() throws SchemaException,
 
     @Test
     public void TestAccumuloKeyConverterCFCQPropertydEntity() throws SchemaException, AccumuloElementConversionException, IOException {
```
p013570: Method names should start with a lower case letter. If possible test methods should use the format 'shouldReturnXWhenY'.

accumulo-store/src/test/java/gaffer/accumulostore/operation/handler/GetElementsBetweenSetsHandlerTest.java
```diff
@@ -48,39 +51,47 @@
...
-        EXPECTED_SUMMARISED_EDGE.putProperty(AccumuloPropertyNames.COUNT, 23 * 3);
-        EXPECTED_SUMMARISED_EDGE.putProperty(AccumuloPropertyNames.TIMESTAMP, TIMESTAMP);
-    }
+    private final List<EntitySeed> seedsA = Arrays.asList(new EntitySeed("A0"));
+    private final List<EntitySeed> seedsB = Arrays.asList(new EntitySeed("A23"));
+
+    private View defaultView;
+    private AccumuloStore byteEntityStore;
+    private AccumuloStore gaffer1KeyStore;
+    private Element expectedEdge1 = new Edge(TestGroups.EDGE, "A0", "A23", true);
```
p013570: It looks like these expected edges could still be final or constants?

accumulo-store/src/test/java/gaffer/accumulostore/utils/AccumuloPropertyNames.java
```diff
@@ -21,4 +21,5 @@
 public class AccumuloPropertyNames extends TestPropertyNames {
     public static final String COLUMN_QUALIFIER = "columnQualifier";
     public static final String TIMESTAMP = "timestamp";
+    public static final String TEST_OPTION_KEY = "testOption";
```
p013570: This test option key isn't a property name, it looks like it is just used as an operation option key, so it shouldn't really be in the AccumuloPropertyNames constants class.

p013570: This test option key isn't a property name, it looks like it is just used as an operation option key, so it shouldn't really be in the AccumuloPropertyNames constants class.

PR #207, opened 2016-05-13
Gh 129 op auths

p013570:


gaffer-core/graph/src/main/java/gaffer/graph/Graph.java
```diff
@@ -107,7 +116,17 @@ private Graph(final Store store, final View view) {
             }
         }
 
-        return store.execute(operationChain, user);
+        for (GraphHook graphHook : graphHooks) {
```
a09631: gh-220 - Add final to enhanced loop variables to ensure that they aren't updated inside of the loop

Issue #215, opened 2016-05-18
Create Operation Chain Example


p013570:
This requires a new java/json example in the getting started module and a walkthrough of the code on the [wiki](https://github.com/GovernmentCommunicationsHeadquarters/Gaffer/wiki/Getting-Started#examples).

p013570:
Needs to be merged in after gh-214.

a09631:
Pull request approved, branch merged in to develop.

PR #218, opened 2016-05-23
gh-215 - Added op chain example

p013570:


example/src/main/java/gaffer/example/gettingstarted/generator/DataGenerator6.java
```diff
@@ -0,0 +1,55 @@
+/*
```
a09631: Out of scope for this pull request, but data generator code could be rationalised without reducing the readability of the actual example classes.

PR #222, opened 2016-05-23
Gh 217 operation examples

p013570:


example/pom.xml
```diff
@@ -63,6 +63,11 @@
             <version>1.6.2</version>
             <scope>compile</scope>
         </dependency>
+        <dependency>
```
a09631: This is now being used in several places.  Can this dependency be pulled in to the top level POM.

gaffer-core/operation/src/main/java/gaffer/operation/AbstractValidatable.java
```diff
@@ -126,6 +126,15 @@ protected Builder(final OP_TYPE op) {
          * @return this Builder
          * @see gaffer.operation.Validatable#setElements(Iterable)
          */
+        public Builder<OP_TYPE, OUTPUT> elements(final Element... elements) {
+            return elements(Arrays.asList(elements));
```
gchqdev03: This could throw a NullPointer as the method can be called with no args or a null object and this would throw a NullPointer after Arrays.asList

example/src/main/java/gaffer/example/operation/GetAllElementsExample.java
```diff
@@ -0,0 +1,80 @@
...
+    public static void main(final String[] args) throws OperationException {
+        new GetAllElementsExample().run();
+    }
+
+    public GetAllElementsExample() {
+        super(GetAllElements.class);
+    }
+
+    public void runExamples(final Graph graph) throws OperationException {
+        getAllEdges(graph);
```
gchqdev03: Methods should be called getAllElements, etc rather than getAllEdges for this example

Issue #225, opened 2016-05-25
Rest service graph factory should use a singleton graph by default


p013570:
Currently a new instance of Graph is created for every operation that is executed via the rest service. This may be required in some cases but it should reuse the same graph instance by default.

p013570:
Merged into release-0.3.6.

PR #226, opened 2016-05-25
gh-225 - Changed rest graph factory to use a singleton graph by default

p013570:


rest-api/src/test/java/gaffer/rest/GraphFactoryTest.java
```diff
@@ -60,4 +61,16 @@ public void shouldReturnNullWhenCreateOpAuthoriserWithNoSystemPropertyPath() {
         // Then
```
gchqdev03: Not directed related to this particular pull request, but a test to show that an IllegalArgumentException is thrown if the Graph Factory Class name is invalid would be good

Issue #230, opened 2016-05-26
Investigate accumuloStore Operation GetElementsBetweenSets operation returning some edges within the first seed set.


james010101101:
The accumuloStore BetweenSets operation is returning some edges within the first seed set.

james010101101:
Merged into develop.

PR #237, opened 2016-05-27
gh-230 Stop GetElementsBetweenSets returning edges within set A.

james010101101:
Also stop adding seed set A to bloom filter and check the destination end of the edge in the bloom filter iterator(s)


accumulo-store/src/main/java/gaffer/accumulostore/key/core/impl/byteEntity/ByteEntityBloomFilterIterator.java
```diff
@@ -0,0 +1,66 @@
...
+
+import java.util.Arrays;
+
+/**
+ * The BloomFilterIterator should filter out elements based on their membership
+ * of the provided bloomFilter. This implementation may not work as desired
+ * depending on your {@link gaffer.accumulostore.key.AccumuloKeyPackage}
+ * implementation.
+ */
+public class ByteEntityBloomFilterIterator extends AbstractCoreKeyBloomFilterIterator {
```
p013570: The Javadoc needs updating to make it specific to the 'ByteEntity' key package

accumulo-store/src/main/java/gaffer/accumulostore/key/core/impl/byteEntity/ByteEntityBloomFilterIterator.java
```diff
@@ -0,0 +1,66 @@
...
+
+    private boolean entity;
+    private int pos = -1;
```
p013570: Having these global fields methods the class is not thread safe. It is also a bit confusing that 'entity' is manipulated in the getSrcVertexFromKey method.

accumulo-store/src/main/java/gaffer/accumulostore/key/core/impl/byteEntity/ByteEntityBloomFilterIterator.java
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    protected byte[] getSrcVertexFromKey(final byte[] key) {
+        for (int i = 0; i < key.length; ++i) {
+            if (key[i] == ByteArrayEscapeUtils.DELIMITER) {
+                pos = i;
+                break;
+            }
+        }
+        if (key.length > pos + 2) {
```
p013570: If possible, could we use constants instead of magic numbers. It would help to clarify why we need to add 2 to the position.

james010101101: An Entity under this package will always be of length pos (the first delimiter) +2, if its larger its an edge.

accumulo-store/src/main/java/gaffer/accumulostore/key/core/impl/byteEntity/ByteEntityBloomFilterIterator.java
james010101101: Because we want to return the position but also do a check on wether what is returned is an entity or edge the check is identical for both cases but you would need to return a complex type if you don't manipulate the variable

accumulo-store/src/main/java/gaffer/accumulostore/key/core/AbstractCoreKeyBloomFilterIterator.java
```diff
@@ -14,57 +14,52 @@
...
-        if (pos != -1) {
-            return Arrays.copyOf(key, pos);
+        Pair<byte[]> vertices = getVertices(key.getRowData().getBackingArray());
+        if (vertices.getFirst() == null) {
+            return true;
         } else {
-            return key;
+            boolean inSrc = filter.membershipTest(new org.apache.hadoop.util.bloom.Key(vertices.getFirst()));
+            boolean inDst = filter.membershipTest(new org.apache.hadoop.util.bloom.Key(vertices.getSecond()));
+            return inSrc || inDst;
```
p013570: It would be slightly more efficient to return straight away if inSrc is true. That way we could skip calculating the inDst value.

accumulo-store/src/main/java/gaffer/accumulostore/key/core/impl/byteEntity/ByteEntityBloomFilterIterator.java
```diff
@@ -0,0 +1,50 @@
...
+        Pair<byte[]> vertices = new Pair<>();
+        for (int i = 0; i < key.length; ++i) {
+            if (key[i] == ByteArrayEscapeUtils.DELIMITER) {
+                pos = i;
+                break;
+            }
+        }
+        if (key.length <= pos + 2) {
+            return vertices;
+        } else {
```
p013570: We don't need the 'else' block here - it would be clearer without it. Just:
```java
vertices.setFirst(Arrays.copyOf(key, pos));
vertices.setSecond(Arrays.copyOfRange(key, pos + 3, key.length - 2));
return vertices;
```

accumulo-store/src/main/java/gaffer/accumulostore/retriever/impl/AccumuloIDBetweenSetsRetriever.java
```diff
@@ -131,15 +141,40 @@ protected boolean checkIfBothEndsInSet(final Object source, final Object destina
...
+                return false;
+            }
+            boolean destMatchesClientFilter;
+            try {
+                destMatchesClientFilter = clientSideFilter.membershipTest(
+                        new org.apache.hadoop.util.bloom.Key(elementConverter.serialiseVertexForBloomKey(destination)));
+            } catch (final AccumuloElementConversionException e) {
+                return false;
+            }
+            return (sourceIsInCurrent && destMatchesClientFilter) || (destIsInCurrent && sourceMatchesClientFilter);
```
p013570: This could be made slightly more efficient if we do (sourceIsInCurrent && destMatchesClientFilter) check and if true we could skip the calculation for destIsInCurrent and sourceMatchesClientFilter.

gaffer01:
It looks like both ends of the edge are checked to see if they are in the Bloom filter. Really it is only necessary to check the non-query end of the edge (i.e. the second one in the rowkey). Checking the query end as well will increase the number of false positives.

james010101101:
Good point
Ill make a change to iterate through the key backwards so we can grab the non query vertex alone which should also be quicker.

PR #246, opened 2016-06-03
Gh 243 tree set

p013570:


simple-serialisation-library/src/main/java/gaffer/serialisation/simple/TreeSetSerialiser.java
```diff
@@ -0,0 +1,79 @@
...
+            str = new String(bytes, CommonConstants.UTF_8);
+        } catch (UnsupportedEncodingException e) {
+            throw new SerialisationException(e.getMessage(), e);
+        }
+
+        final TreeSet<String> treeSet = new TreeSet<>();
+        final Iterable<String> items = Splitter.on(COMMA)
+                .omitEmptyStrings()
+                .split(str.substring(1, str.length() - 1));
+        for (String item : items) {
```
a09631: item should be marked final

simple-function-library/src/main/java/gaffer/function/simple/aggregate/TreeSetAggregator.java
```diff
@@ -0,0 +1,59 @@
...
+import java.util.TreeSet;
+
+/**
+ * An <code>TreeSetAggregator</code> is a {@link SimpleAggregateFunction} that
+ * combines {@link TreeSet}s.
+ */
+@Inputs(TreeSet.class)
+@Outputs(TreeSet.class)
+public class TreeSetAggregator extends SimpleAggregateFunction<TreeSet<String>> {
+    private TreeSet<String> treeSet;
```
james010101101: Can we remove the generics and make it just a set of objects, this will mean this aggregator can be re-used by different types

simple-serialisation-library/src/main/java/gaffer/serialisation/simple/TreeSetSerialiser.java
```diff
@@ -0,0 +1,79 @@
...
+import gaffer.serialisation.Serialisation;
+import java.io.UnsupportedEncodingException;
+import java.util.Iterator;
+import java.util.TreeSet;
+
+/**
+ * A <code>TreeSetSerialiser</code> is a serialiser for {@link TreeSet}s with
+ * {@link String} values.
+ */
+public class TreeSetSerialiser implements Serialisation {
```
james010101101: Can this be renamed to StringTreeSetSerialiser or some other variation that references that this serialiser will only serialise tree sets of type string (Or any object that has an accurate toString() method representation)

PR #252, opened 2016-06-08
Gh 139 python

p013570:


p013570:
Need to fix a few minor bugs with the operation classes

PR #262, opened 2016-06-13
Gh 193 docker

p013570:


p013570:
The centos folder inside docker is not required and could be removed.

PR #285, opened 2016-06-23
Gh 247 export operations

p013570:


gaffer-core/operation/src/main/java/gaffer/export/Exporter.java
```diff
@@ -0,0 +1,109 @@
...
+
+    /**
+     * Initialises the export. This base method just stores the current user.
+     * Override this method (and call super) to add further initialisation.
+     *
+     * @param key    export key
+     * @param config configuration for the export (This will be an instance of a gaffer Store)
+     * @param user   the user who initiated the export
+     */
+    public void initialise(final String key, final Object config, final User user) {
```
GCHQ-11: Could we template the class of the config object? Looking at the Gaffer exporter we cast it to a Store, if we template it, the child classes wouldn't require casting

PR #289, opened 2016-06-29
Gh 288 deduplicate results

p013570:


GCHQ-11:
Could de-duplication be done as a separate step in an Operation Chain? It could be implemented in the general operations rather than being Accumulo specific.

p013570:
I'll make the suggested chain and then reopen the pull request.

p013570:
I've now created a Deduplicate operation. There is now a deduplicate flag on GetOperation that will automatically add a deduplicate operation into your operation chain for you.

gaffer-core/graph/src/main/java/gaffer/graph/Graph.java
```diff
@@ -110,20 +110,25 @@ private Graph(final Store store, final View view, final List<GraphHook> graphHoo
...
+        final OperationChain<OUTPUT> operationChainClone = operationChain.clone();
+
+        // Update the view
+        for (Operation operation : operationChainClone.getOperations()) {
             if (null == operation.getView()) {
                 operation.setView(view);
             }
         }
 
         for (GraphHook graphHook : graphHooks) {
```
a09631: Mark graphHook final

PR #291, opened 2016-06-29
Gh 258 summarising over row id's in accumulo

james010101101:
Adding an iterator that will Aggregate multiple property sets across different RowID's (Vertex's) for a given range where the Column Family (Group) is the same.


accumulo-store/src/main/java/gaffer/accumulostore/operation/handler/SummariseGroupOverRangesHandler.java
```diff
@@ -0,0 +1,73 @@
...
+                                         final Context context, final Store store)
+            throws OperationException {
+        return doOperation(operation, context.getUser(), (AccumuloStore) store);
+    }
+
+    public Iterable<Element> doOperation(final SummariseGroupOverRanges<Pair<ElementSeed>, Element> operation,
+                                         final User user,
+                                         final AccumuloStore store) throws OperationException {
+
+        int edges = operation.getView().getEdgeGroups().size();
```
gaffer01: This could be final, and might be better named numEdgeGroups?

accumulo-store/src/main/java/gaffer/accumulostore/operation/handler/SummariseGroupOverRangesHandler.java
```diff
@@ -0,0 +1,73 @@
...
+        int entities = operation.getView().getEntityGroups().size();
```
gaffer01: This could be final, and might be better named numEntityGroups?

accumulo-store/src/main/java/gaffer/accumulostore/key/core/AbstractCoreKeyIteratorSettingsFactory.java
```diff
@@ -48,7 +49,7 @@ public IteratorSetting getElementFilterIteratorSetting(final View view, final Ac
     }
 
     @Override
-    public IteratorSetting getAggregatorIteratorSetting(final AccumuloStore store) throws IteratorSettingException {
+     public IteratorSetting getAggregatorIteratorSetting(final AccumuloStore store) throws IteratorSettingException {
```
gaffer01: The only difference here is that a space has been added at the start of the line - it should be deleted.

accumulo-store/src/main/java/gaffer/accumulostore/utils/AccumuloStoreConstants.java
```diff
@@ -25,7 +25,8 @@
     public static final String BLOOM_FILTER_ITERATOR_NAME = "Bloom_Filter";
     public static final String ELEMENT_FILTER_ITERATOR_NAME = "Element_Filter";
     public static final String EDGE_ENTITY_DIRECTED_UNDIRECTED_INCOMING_OUTGOING_FILTER_ITERATOR_NAME = "Edge_Entity_Directed_Undirected_Incoming_Outgoing_Filter";
-    public static final String QUERY_TIME_AGGREGATION_ITERATOR_NAME = "Query_Time_Aggregator";
+    public static final String COLUMN_QUALIFIER_AGGREGATOR_ITERATOR_NAME = "Column_Qualifier_Aggregator";
```
gaffer01: Does this actually aggregate over column visibility as well? If so, should the name reflect that?

accumulo-store/src/test/java/gaffer/accumulostore/key/impl/RowIdAggregatorTest.java
```diff
@@ -0,0 +1,274 @@
...
+            edge6.setSource("1");
+            edge6.setDestination("5");
+            edge6.setDirected(true);
+            edge6.putProperty(AccumuloPropertyNames.COLUMN_QUALIFIER, 2);
+            edge6.putProperty(AccumuloPropertyNames.PROP_1, 0);
+            edge6.putProperty(AccumuloPropertyNames.PROP_2, 0);
+            edge6.putProperty(AccumuloPropertyNames.PROP_3, 0);
+            edge6.putProperty(AccumuloPropertyNames.PROP_4, 0);
+
+
```
gaffer01: Unnecessary and inconsistent empty line.

accumulo-store/src/test/java/gaffer/accumulostore/key/impl/RowIdAggregatorTest.java
```diff
@@ -0,0 +1,274 @@
...
+            // Read data back and check we get one merged element
+            final Authorizations authorizations = new Authorizations(visibilityString);
+            final BatchScanner scanner = store.getMockConnector().createBatchScanner(store.getProperties().getTable(), authorizations, 1000);
+            try {
+                scanner.addScanIterator(store.getKeyPackage().getIteratorFactory().getRowIDAggregatorIteratorSetting(store, "BasicEdge2"));
+            } catch (IteratorSettingException e) {
+                fail(e.getMessage());
+            }
+            RangeFactory rangeF = store.getKeyPackage().getRangeFactory();
```
gaffer01: Could be final.

accumulo-store/src/test/java/gaffer/accumulostore/key/impl/RowIdAggregatorTest.java
```diff
@@ -0,0 +1,274 @@
...
+            Range r = rangeF.getRangeFromPair(new Pair<ElementSeed>((new EntitySeed("1")), new EntitySeed("4")), new SummariseGroupOverRanges());
```
gaffer01: Could be final.

accumulo-store/src/test/java/gaffer/accumulostore/key/impl/RowIdAggregatorTest.java
```diff
@@ -0,0 +1,274 @@
...
+            Range r2 = rangeF.getRangeFromPair(new Pair<ElementSeed>((new EntitySeed("5")), new EntitySeed("5")), new SummariseGroupOverRanges());
```
gaffer01: Could be final.

accumulo-store/src/test/java/gaffer/accumulostore/key/impl/RowIdAggregatorTest.java
```diff
@@ -0,0 +1,274 @@
...
+            expectedEdge.putProperty(AccumuloPropertyNames.PROP_1, 0);
+            expectedEdge.putProperty(AccumuloPropertyNames.PROP_2, 0);
+            expectedEdge.putProperty(AccumuloPropertyNames.PROP_3, 0);
+            expectedEdge.putProperty(AccumuloPropertyNames.PROP_4, 0);
+            expectedEdge.putProperty(AccumuloPropertyNames.COUNT, 3);
+
+            assertEquals(expectedEdge, readEdge);
+            assertEquals(5, readEdge.getProperty(AccumuloPropertyNames.COLUMN_QUALIFIER));
+            assertEquals(3, readEdge.getProperty(AccumuloPropertyNames.COUNT));
+            // Check no more entries
```
gaffer01: This comment doesn't make sense - the next line asserts that more entries do exist.

gaffer01:
Looks good. Can one of the classes (probably the main operation `SummariseGroupOverRanges`) have some Javadoc explaining what it does?

Iterators can't aggregate across multiple tablet servers. If a range spans multiple tablet servers, the current code will not return one aggregated result. Somewhere it should be documented that this is the case.

accumulo-store/src/main/java/gaffer/accumulostore/utils/AccumuloStoreConstants.java
james010101101: That will aggregate the column qualifier downwards so yes including visibility timestamp and value
I thought when compared with the RowID aggregator it'd start to make more contextual sense as the iterator naming pattern would be that key portion down.

gaffer01: That makes sense, leave it as it is.

PR #292, opened 2016-06-29
Operations can be configured via store.properties

GCHQ-11:
store.properties will refer to a new JSON file, generally called operationDeclarations.json


gaffer-core/store/src/test/java/gaffer/store/operations/OperationDeclarationJsonTest.java
```diff
@@ -0,0 +1,57 @@
+package gaffer.store.operations;
```
a09631: Missing file header.

PR #305, opened 2016-07-07
json deserialise option to use type references

GCHQ-11:
When receiving Iterable<Edge> from the REST API, I would like to be able to deserialise it in one go, the JSON library we are using supports this using TypeReferences.


gaffer-core/serialisation/src/main/java/gaffer/jsonserialisation/JSONSerialiser.java
```diff
@@ -124,31 +126,70 @@ public void serialise(final Object object, final JsonGenerator jsonGenerator, fi
...
+
+    /**
+     * @param stream the {@link java.io.InputStream} containing the bytes of the object to deserialise
+     * @param clazz   the class of the object to deserialise
+     * @param <T>    the type of the object
+     * @return the deserialised object
+     * @throws SerialisationException if the bytes fail to deserialise
+     */
+    public <T> T deserialise(final InputStream stream, final Class<T> clazz) throws SerialisationException {
+        return deserialise(stream, new ClassTypeReference<>(clazz));
```
p013570: These deserialise methods are used a lot, so we should minimise the number of objects we instantiate. Although it is more code, it might be better if 

```java
deserialise(final byte[] bytes, final Class<T> clazz)
```

just delegated directly to mapper.readValue(bytes, clazz);

PR #309, opened 2016-07-11
Gh 295 input format

gaffer01:


accumulo-store/src/main/java/gaffer/accumulostore/AccumuloStore.java
```diff
@@ -125,6 +137,53 @@ public Connector getConnection() throws StoreException {
...
+            if (!view.getEntityGroups().isEmpty() || !view.getEdgeGroups().isEmpty()) {
+                IteratorSetting elementFilter = getKeyPackage()
+                        .getIteratorFactory()
+                        .getElementFilterIteratorSetting(view, this);
+                InputConfigurator.addIterator(AccumuloInputFormat.class, conf, elementFilter);
+            }
+        } catch (final AccumuloSecurityException | UnsupportedEncodingException e) {
+            throw new StoreException(e);
+        } catch (final IteratorSettingException e) {
+            e.printStackTrace();
```
p013570: Would be better to use a logger or throw a StoreException

accumulo-store/src/main/java/gaffer/accumulostore/inputformat/ElementInputFormat.java
```diff
@@ -0,0 +1,116 @@
...
+import java.util.Map;
+
+/**
+ *
+ */
+public class ElementInputFormat extends InputFormatBase<Element, NullWritable> {
+
+    public static final String KEY_PACKAGE = "KEY_PACKAGE";
+    public static final String SCHEMA = "SCHEMA";
+    public static final String VIEW = "VIEW";
```
p013570: We have a VIEW and a SCHEMA key in AccumuloStoreConstants - perhaps we could reuse these?

PR #322, opened 2016-07-15
Gh 293 implement new accumulostore value storage

james010101101:


gaffer01:
Looks good, but can we wait before merging this? I'm about to submit a pull request for a serialisation of ints that uses a variable number of bytes (and only 1 byte if the value is less than 128). If we use this in this code where currently 4 bytes are written to give the length of the serialised property, then that should be a significant saving.

gaffer01:
Do the unit tests check that it works correctly with null properties? There is code to deal with that, and there are unit tests to check that missing properties are dealt with, but there don't seem to be any tests that check that it works if the property is null.

james010101101:
The unit tests don't specifically have a test for if a given property is set to null, 
But the case is covered by the tests for missing properties, because the behaviour will be the same (asking for a property that doesn't have an entry from the properties map will return null - the correct value and the code checks the value is null on the serialisation and just skips entering that property)
But I will add a test in so we can see this in action.

Issue #166, opened 2016-04-27
Accumulo store - should initialisation create table?


gaffer01:
In the LoadAndQuery1 example, if a query is made before elements have been added then an error is thrown as the table hasn't been created. It should be possible to query a graph which no data has been put into.

It should also be possible to create a graph, and then insert data from multiple independent processes. In this case it is important that the table exists before the data insertion starts.

Suggest that the initialise method on AccumuloStore creates the table.

gchqdev03:
I think it would be preferable to return a meaningful exception if someone tries to query a table that does not exist rather than creating a new table anytime someone attempts to query a table that doesn't exist

gaffer01:
The point is that initialising a graph with particular store properties should cause that store to be instantiated. Only initialising it when addElements is called leads to problems when multiple parallel processes simultaneously start writing to a new store.

You shouldn't be able to query a table that doesn't exist. You query via a graph and initialising the graph causes the store and hence table to be initialised. If you don't have permissions to create a table then a meaningful exception should be thrown.

james010101101:
Merged into develop.

PR #341, opened 2016-07-20
gh-166 Initialise method in AccumuloStore now creates table

gaffer01:


accumulo-store/src/main/java/gaffer/accumulostore/AccumuloStore.java
```diff
@@ -250,7 +251,7 @@ protected void addAdditionalOperationHandlers() {
      * @throws StoreException failure to insert the elements into a table
      */
     public void addElements(final Iterable<Element> elements) throws StoreException {
-        TableUtils.ensureTableExists(this);
+        TableUtils.ensureTableExists(this); // Needed for tests that access store directly
```
james010101101: I think we should remove this check and force all tests to call initialise.
Better than wasting a check every addElementsCall.
Similarly you can remove the check from the BulkImport Operation

gaffer01:
If I comment out the `TableUtils.ensureTableExists(this)` line in the `AddElements` method in `AccumuloStore` then lots of tests fail, e.g. `AccumuloIDBetweenSetsRetrieverTest` fails even though that test does initialise the store. This is because it uses `SingleUseMockAccumuloStore`. The `initialise` method in that calls `super.initialise` which ensures the table exists, but it is then deleted in the rest of the `initialise` method.

Changing the order of `super.initialise` and the deleting of the table calls does not solve the problem as then the `getConnection` method fails as the `MockInstance` has not been initialised.

This can be solved by calling `super.initialise` both before and after the delete table call in `initialise` in `SingleUseMockAccumuloStore` although that looks a bit odd.

Probably the best long-term solution is to avoid using `SingleUseMockAccumuloStore` and give `MockAccumuloStore` the option to delete the table. Or each test could create its own `MockAccumuloStore` with its own unique instance name.

PR #353, opened 2016-08-02
Gh 268 default serialisers

p013570:


gaffer-core/serialisation/src/main/java/gaffer/serialisation/implementation/SerialisationFactory.java
```diff
@@ -0,0 +1,51 @@
...
+ * limitations under the License.
+ */
+package gaffer.serialisation.implementation;
+
+import gaffer.serialisation.Serialisation;
+import gaffer.serialisation.implementation.raw.CompactRawIntegerSerialiser;
+import gaffer.serialisation.implementation.raw.CompactRawLongSerialiser;
+
+/**
+ * This class is used to serialise and deserialise a boolean value
```
gaffer01: This description of the class is wrong.

gaffer-core/serialisation/src/test/java/gaffer/serialisation/implementation/TreeSetStringSerialiserTest.java
```diff
@@ -0,0 +1,86 @@
...
+    public void shouldSerialiseAndDeserialiseATreeSet() throws SerialisationException {
+        // Given
+        final TreeSet<String> set = new TreeSet<>();
+        set.add("string1");
+        set.add("string2");
+        set.add("string3");
+        set.add("string4");
+
+        // When
+        final byte[] serialiedSet = SERIALISER.serialise(set);
```
gaffer01: Should be serialisedSet not serialiedSet.

gaffer-core/serialisation/src/test/java/gaffer/serialisation/implementation/TreeSetStringSerialiserTest.java
```diff
@@ -0,0 +1,86 @@
...
+        assertEquals(deserialisedSet, set);
+    }
+
+    @Test
+    public void shouldSerialiseAndDeserialiseAnEmptyTreeSet() throws SerialisationException {
+        // Given
+        final TreeSet<String> set = new TreeSet<>();
+
+        // When
+        final byte[] serialiedSet = SERIALISER.serialise(set);
```
gaffer01: Should be serialisedSet not serialiedSet.

gaffer01:
If no serialiser is provided for a vertex, is a default one used? If so, then it might be worth using a different default serialiser for an integer that is a vertex rather than an integer that is a property.  If a vertex is an integer, then you might want to use a serialiser that preserves the ordering (so not the CompactRaw one), but if the integer is just a property then you don't need to preserve ordering and so should use the CompactRaw one.

p013570:
Closing the pull request whilst the suggested changes are made.

PR #363, opened 2016-08-10
Added upsert operation and unit tests

m316257:


simple-types-library/src/main/java/gaffer/types/simple/FreqMap.java
```diff
@@ -38,4 +38,27 @@ public FreqMap(final int initialCapacity) {
...
+
+    /**
+     * Adds a new key and value to the map if the key is not already there.
+     * If the key is already there, the value supplied is added to the existing value for the key and the result is inserted into the map.
+     * @param key The key in the map to increment or insert.
+     * @param value The value to increment by or initialise to.
+     */
+    public void upsert(final String key, final Integer value) {
+        if (this.containsKey(key)) {
+            this.put(key, this.get(key) + value);
```
p013570: this.get(key) could return null. 

We should consider adding some protection around it. Also an extra unit test for the case when the current value is null would be good.

Perhaps the following:

```java
final Integer currentValue = get(key);
if(null == currentValue) {
    put(key, value)
} else {
    put(key, currentValue + value)
}
```

PR #369, opened 2016-08-16
Gh 276 - groupBy field in view to override schema groupBy

p013570:


accumulo-store/src/main/java/gaffer/accumulostore/key/AccumuloElementConverter.java
```diff
@@ -227,6 +227,42 @@ Properties getPropertiesFromColumnQualifier(final String group, final byte[] col
...
+    /**
+     * Truncates the provided bytes, returning the bytes for the specified
+     * number of properties.
+     *
+     * @param numProps the number of properties to extract
+     * @param bytes    the full list of property bytes
+     * @return the truncated property bytes.
+     * @throws AccumuloElementConversionException if truncation of the bytes fails
+     */
+    byte[] truncatePropertyBytes(final int numProps, final byte[] bytes) throws AccumuloElementConversionException;
```
james010101101: This method should not be on the interface it only makes sense for specific implementations of a key package.

p013570:
Needs some additional tests

PR #373, opened 2016-08-17
Gh 361 number of reducers for bulk import

gaffer01:


p013570:
Looks good but we could really do with some unit tests for the new functionality.

Issue #57, opened 2016-02-17
Rest Api - Investigate limiting results of an operation chain


james010101101:
Should an N limit apply to only the first step of an operation chain?
e.g generate first N elements. Then add all these N elements.

Or should it apply to every step in an operation chain?
e.g Return the first N elements matching the given seeds then pass these N elements as the seeds of a second get Operations and limit the results of that operation to N.

Or should it apply to only the last step in an operation chain?
e.g Return all elements matching the given seeds then pass these elements as the seeds of a second get operation and return only N results of that operation.

Also investigate what to do when the result on an operation chain is void?

p013570:
The limiting number could be added as a field to the operation class.

p013570:
Merged into develop.

p013570:
## Migration Steps
The query parameter 'n' has been removed from the rest api. Instead, each get operation now has a resultsLimit field - this allows each operation in a chain to be limited instead of just the final operation.

PR #379, opened 2016-08-22
gh-57 - added result limit to get operations

p013570:


GCHQ-11:
Could this result limit functionality be implemented as a separate operation?
OperationChain.first(GetElements).then(LimitElements)

Where Limit elements took INPUT -> Iterable<SOMETHING>, OUTPUT -> Iterable<SOMETHING>

p013570:
Closing pull request whilst we make the suggested changes

GCHQ-11:
Now that we have the Limit as a separate step, do we still need the limit parameter in the AbstractGetOperation (together with the need to add a function to all the child class builders)

p013570:
I think the limit parameter is actually quite useful. It makes it a lot simpler and avoids users having to provide 2 operations when they need to limit their results. It also follows the same pattern we did for deduplicating results.

PR #380, opened 2016-08-22
Gh 160 accumulo op examples

p013570:


docs/checkstyle.xml
```diff
@@ -75,7 +75,7 @@
         <module name="MethodParamPad"/>
         <module name="NoWhitespaceAfter"/>
         <module name="NoWhitespaceBefore"/>
-        <module name="OperatorWrap"/>
+        <!--<module name="OperatorWrap"/>-->
```
a09631: Can this be removed if it's no longer required?

example/src/main/java/gaffer/example/function/filter/OrExample.java
```diff
@@ -0,0 +1,94 @@
...
+    public void runExamples() {
+        isLessThan2OrIsMoreThan2();
+        property1IsLessThan2OrProperty2IsMoreThan2();
+    }
+
+    public void isLessThan2OrIsMoreThan2() {
+        runExample(new Or(Arrays.asList(
+                        new ConsumerFunctionContext.Builder<Integer, FilterFunction>()
+                                .select(0) // select first property
+                                .execute(new IsLessThan(2))
```
a09631: Shouldn't this be less than 20?

p013570: The method name is actually "is less than 2 or is more than 2".

Issue #235, opened 2016-05-26
Switch to using ClosableIterables instead of Iterables


p013570:
Like the Accumulo store implementation. most store implementations will probably return an iterable that needs to be closed. Currently if users fully iterate over the results then the Accumulo results iterable will be automatically closed. However, if only a few results are read then it is too easy to forget to manually close the iterable as users have to do something like:

```java
if(results instanceof CloseableIterable) {
    ((CloseableIterable)results).close();
}
```

This also couples client code directly to the Accumulo implementation and requires knowledge of how Accumulo handles operations. 

Note - implementing this change will require modifications to a large number of classes.

gaffer01:
Agreed. In general stores should return their results lazily, e.g. if the user queries for a node which has a lot of edges, those edges shouldn't be materialised into a set and then returned. Instead, edges should be read from the underlying data storage as the user consumes them by reading through the iterable. The ArrayList store materialises the set, which should be discouraged and so should be changed to return results lazily. I'll raise a separate issue about this.

p013570:
Merged into develop

PR #382, opened 2016-08-23
Gh 306 spark operations

gaffer01:


p013570:
The travis fix I put doesn't return the error code properly so the build is failing but it isn't getting reported correctly. I'll have a look at this now.

The errors are due to the new Closable Iterables being merged in (gh-235). Either we need to change your operation handlers to return closable iterables or implement gh-396 first.

Issue #294, opened 2016-06-30
Locality groups in Accumulo store


gaffer01:
Accumulo's locality groups could be used so that queries that only require certain groups do not have to scan other groups (see http://accumulo.apache.org/1.7/accumulo_user_manual#_locality_groups). This is very natural with the current approach of storing the group in the column family, as a locality group is just a set of column families.

A sensible default option might be for different groups to be stored in different locality groups. This should speed up scans which require only a subset of the groups. Once locality groups are enabled, Accumulo will automatically take advantage of them when the `fetchColumnFamilies` method is called (which is already used in Gaffer).

p013570:
Merged into develop.

PR #386, opened 2016-08-25
gh-294 Setting locality groups on table creation/schema update

james010101101:
The default will be each schema group as its own locality group, this can be changed by users in the accumulo shell if needed.


a09631:
Is it possible to put some testing around this change?

PR #394, opened 2016-09-07
Gh 16 add post filter

james010101101:


accumulo-store/src/main/java/gaffer/accumulostore/key/IteratorSettingFactory.java
```diff
@@ -55,7 +55,20 @@
      * @return A new {@link IteratorSetting} for an Iterator capable of filtering {@link gaffer.data.element.Element}s based on a {@link View}
      * @throws IteratorSettingException if an iterator setting could not be created
      */
-    IteratorSetting getElementFilterIteratorSetting(final View view, final AccumuloStore store)
+    IteratorSetting getElementPreAggregationFilterIteratorSetting(final View view, final AccumuloStore store)
+            throws IteratorSettingException;
+
+    /**
```
p013570: This javadoc and the one above should differentiate between pre and post aggregation.

p013570:
Please can you add some migration steps to the issue.

PR #395, opened 2016-09-09
Gh 393 graphql api

GCHQ-11:
This implements a conversion from Gaffer to GraphQL, not REST accessible yet, but proves the concept


graffer-ql/README.md
```diff
@@ -0,0 +1,130 @@
...
+
+    {Film(vertex:"filmA"):{name{value}}}
+  
+To get Edges you must provide either
+1. The source vertex as 'source'
+2. The destination vertex as 'destination'
+3. Both 'source' and 'destination'
+4. Either source or destination as 'vertex'
+
+E.g (assuming FriendsWith is directional)
```
p013570: FriendsWith isn't actually an Edge in the example schema. Perhaps it is worth just pointing this out.

graffer-ql/pom.xml
```diff
@@ -0,0 +1,28 @@
...
+        <dependency>
+            <groupId>com.graphql-java</groupId>
+            <artifactId>graphql-java</artifactId>
+            <version>2.1.0</version>
+        </dependency>
+
+        <dependency>
+            <groupId>gaffer</groupId>
+            <artifactId>example</artifactId>
+            <version>0.4.2-SNAPSHOT</version>
```
p013570: Can we change this to ${project.parent.version} please.

graffer-ql/src/test/java/graphql/gaffer/ContextBuilderTest.java
```diff
@@ -0,0 +1,99 @@
...
+import org.junit.Before;
+import org.junit.Test;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.fail;
+
+/**
+ * Created by joe on 9/8/16.
+ */
+public class ContextBuilderTest {
```
p013570: Unnecessary javadoc

graffer-ql/src/test/java/graphql/gaffer/ContextBuilderTest.java
```diff
@@ -0,0 +1,99 @@
...
+    @Test
+    public void testNoGraph() {
+        try {
+            final GrafferQLContext context = new GrafferQLContext.Builder()
+                    .user(user)
+                    .build();
+
+            fail("Should have thrown exception with missing graph");
+        } catch (GrafferQLException e) {
+            // goodie gumpdrops
```
p013570: Add throws GrafferQLException to the method so that if it is thrown the test will fail.

graffer-ql/README.md
```diff
@@ -0,0 +1,130 @@
...
+    User user = new User.Builder()
+        .userId("me")
+        .dataAuth("see_everything")
+                
+2. Build a context for GrafferQL to use, this requires
+  a. The gaffer.graph.Gaffer graph
+  b. The gaffer.user.User
+
+    GrafferQLContext context = new GrafferQLContext.Builder()
+        .graph(graph) // will be same value for all requests to this graph
```
GCHQ-11: yeah agreed, I struggled to get it to play ball. Sometimes it puts text in the little grey boxes, but not everywhere. There seems to be no way to explicitly mark it up as 'code'

graffer-ql/pom.xml
GCHQ-11: good catch :)

p013570:
We are planning to create a separate Gaffer Tools repository, where all the tools like this, the python shell and the tinkerpop implementation will live.

Issue #409, opened 2016-09-15
Example rest api sometimes fails to deploy


p013570:
It looks like there are conflicts between the automatic resource scanning and the use of the ApplicationConfig class. 

The rest easy docs state: 'If you have [Application] set, you should probably turn off automatic scanning as this will probably result in duplicate classes being registered".

p013570:
Merged into develop.

PR #410, opened 2016-09-15
gh-409 - switched swagger dependency to io.swagger

p013570:


p013570:
We have tested it on tomcat, jetty and firefly - that is probably enough different types.

Issue #398, opened 2016-09-12
Create simple UI that connects to a Gaffer REST API


p013570:

p013570:
Merged into develop

PR #416, opened 2016-09-19
gh-398 prototype ui

p013570:


ui/README.md
```diff
@@ -0,0 +1,45 @@
...
+
+
+Option 2 - Build using the standalone-ui profile
+=============================================
+
+The application can be built and then run as a basic executable standalone war file from Maven. When run in this format, the example rest api is also deployed.
+
+To build it and its dependencies, use the following command from the parent directory:
+
+'mvn clean install -P standalone-ui'
```
gaffer01: Is it worth adding `-DskipTests` to this line to speed it up?

ui/README.md
```diff
@@ -0,0 +1,45 @@
...
+This uses the 'standalone-ui' profile to start a tomcat server with the ui and example-rest wars deployed.
+
+The ui can then be accessed via http://localhost:8080/ui and the rest api is at The ui can then be accessed via http://localhost:8080/rest
```
gaffer01: Sentence starts twice: "and the rest api is at The ui can then be accessed".

p013570: Good spot - thanks.

ui/src/main/webapp/app/controller.js
```diff
@@ -0,0 +1,461 @@
+/*
```
gchqdev03: Would probably better to split this monolithic controller down a bit. Perhaps use angular services to encapsulate some of the logic and inject the service into the controller. 
Maybe even split the controller into multiple controllers that perform discrete functions

Issue #417, opened 2016-09-20
Create quick-compile maven profile


p013570:
As we don't yet have artefacts pushed up to maven central, users need to compile Gaffer before they can use it. To speed up this process we could create a maven profile that skips tests, checkstyle and findbugs.

p013570:
Merged into develop.

PR #426, opened 2016-09-22
gh-417 - added quick compile maven profile

p013570:


example/src/main/resources/example/gettingstarted/intro.md
```diff
@@ -13,5 +13,19 @@ To do this you need to do a few things:
...
 
+But before that you will need to download the code and compile it - sorry we are not yet publishing our artifacts to maven central.
+
+Start by cloning the gaffer GitHub project.
+
+```bash
+git clone https://github.com/GovernmentCommunicationsHeadquarters/Gaffer.git
+```
+
+Then from inside the Gaffer folder run the maven 'quick' build. There are quite a few dependencies to download so it is not actually that quick.
```
GCHQ-11: lol

PR #430, opened 2016-09-23
Gh 414 create operation chain limiter graph hook

james010101101:


gaffer-core/graph/src/main/java/gaffer/graph/hook/OperationChainLimiter.java
```diff
@@ -0,0 +1,262 @@
...
+ * user is authorised to execute an operation chain based on that user's maximum chain score and the configured score value for each operation in the chain.
+ * This class requires a map of operation scores, these can be added using setOpScores(Map<Class,Integer>) or
+ * addOpScore(Class, Integer...). Alternatively a properties file can be provided.
+ * When using a properties file the last entry in the file that an operation can be assigned to will be the score that is used for that operation.
+ * E.g if you put gaffer.operation.impl.add.AddElements = 8
+ * And then gaffer.operation.impl.add = 1
+ * The add elements will have a score of 1 not 8.
+ * So make sure to write your properties file in class hierarchical order.
+ *
+ * containing the operations and the score.
```
p013570: Looks like an incomplete sentence.

gaffer-core/graph/src/main/java/gaffer/graph/hook/OperationChainLimiter.java
```diff
@@ -0,0 +1,262 @@
...
+    @Override
+    public void postExecute(final Object result, final OperationChain<?> opChain, final User user) {
+        // This method can be overridden to add additional authorisation checks on the results.
+    }
+
+    protected Integer authorise(final Operation operation, final User user) {
+        Integer operationScore = null;
+        if (null != operation) {
+            final Class<? extends Operation> opClass = operation.getClass();
+            for (Map.Entry<Class<? extends Operation>, Integer> entry : operationScores.entrySet()) {
```
p013570: It would be more efficient to add a break when the first entry with assignable operation class is found.
If you need the last operation class then can you iterate through the operation scores in reverse order?

gaffer-core/graph/src/main/java/gaffer/graph/hook/OperationChainLimiter.java
```diff
@@ -0,0 +1,262 @@
...
+                       operationScore = entry.getValue();
+                   }
+                }
+            }
+        if (operationScore != null) {
+            return operationScore;
+        } else {
+            LOGGER.warn("The operation '" + operation.getClass().getName() + "' was not found in the config file provided the configured default value of " + DEFAULT_OPERATION_SCORE + " will be used");
+            return DEFAULT_OPERATION_SCORE;
+        }
```
p013570: It would be simpler to have a single return statement. Perhaps we could change this to:

```java
if (null == operationScore) {
    LOGGER.warn("The operation '" + operation.getClass().getName() + "' was not found in the config file provided the configured default value of " + DEFAULT_OPERATION_SCORE + " will be used");
    operationScore = DEFAULT_OPERATION_SCORE;
}

return operationScore;
```

gaffer-core/graph/src/main/java/gaffer/graph/hook/OperationChainLimiter.java
```diff
@@ -0,0 +1,262 @@
...
+    }
+
+    private void loadMapsFromProperties(final Properties operationScorePropertiesFile, final Properties operationAuthorisationScoreLimitPropertiesFile) {
+        Map<Class<? extends Operation>, Integer> opScores = new LinkedHashMap<>();
+        for (String opClassName : operationScorePropertiesFile.stringPropertyNames()) {
+            final Class<? extends Operation> opClass;
+            try {
+                opClass = Class.forName(opClassName).asSubclass(Operation.class);
+            } catch (ClassNotFoundException e) {
+                throw new IllegalArgumentException(e);
```
p013570: A error message would be helpful, e.g. "An operation class could not be found for operation score property " + opClassName.

gaffer-core/common-util/src/main/java/gaffer/commonutil/StreamUtil.java
```diff
@@ -37,6 +37,9 @@
     public static final String STORE_TYPES = SCHEMA_FOLDER + "storeTypes.json";
     public static final String STORE_PROPERTIES = "/store.properties";
     public static final String OP_AUTHS = "/opAuths.properties";
+    public static final String OP_SCORES = "/opScores.properties";
+    public static final String ROLE_SCORES = "/roleScores.properties";
```
p013570: It might be clearer to call these properties: "role max scores" or "max role scores". Just to ensure they are understood to be max scores and not scores that will be incremented.

james010101101: I think that makes it more confusing, its a file of authorisation scores. A max of these is taken but for the actual properties file thats irrelevant its just an auth has a score.

PR #431, opened 2016-09-23
Gh 420 unnecessary call to system.current time millis()

james010101101:


accumulo-store/src/main/java/gaffer/accumulostore/key/impl/RowIDAggregator.java
```diff
@@ -176,7 +176,7 @@ protected void findTop() throws IOException {
             topKey = new Key(workKey.getRowData().getBackingArray(), group.getBytes(CommonConstants.UTF_8),
                     elementConverter.buildColumnQualifier(group, topProperties),
                     elementConverter.buildColumnVisibility(group, topProperties),
-                    elementConverter.buildTimestamp(topProperties, workKey.getTimestamp()));
+                    elementConverter.buildTimestamp(topProperties));
```
p013570: It's not really a problem but something to be aware of ... 

When elements are merged together if there is no timestamp property defined in the schema, accumulo will now override the accumulo timestamp column to the current time. 

If there is no timestamp property, this column won't really be used by anything in gaffer, so it shouldn't be a problem  - but users may have custom code to use it.

james010101101: If you haven't specified a property for timestamp then you wont see the result anyway, if your using custom code its best as a general rule not to rely on anything not in the schema.

PR #445, opened 2016-10-06
Gh 442 jackson type info

p013570:


p013570:
Pull request gh-441 should be merged in first.

Issue #454, opened 2016-10-12
Shaded jar names are inconsistent


p013570:
e.g.
- accumulo-store-iterators-${GAFFER_VERSION}.jar
- simple-serialisation-library-${GAFFER_VERSION}-serialisers.jar
- simple-function-library-${GAFFER_VERSION}-functions.jar

but the installed artefacts in your maven repository are called:
- accumulo-store-${GAFFER_VERSION}-iterators.jar
- simple-function-library-${GAFFER_VERSION}-shaded.jar
- simple-serialisation-library-${GAFFER_VERSION}-shaded.jar 

These are also referenced incorrectly in the accumulo store guide on the wiki.

The results should be the following:
- accumulo-store-${GAFFER_VERSION}-iterators.jar
- simple-serialisation-library-${GAFFER_VERSION}-serialisers.jar
- simple-function-library-${GAFFER_VERSION}-functions.jar

This will involve adding something similar to the following in the poms:

```xml
 <shadedArtifactAttached>true</shadedArtifactAttached>
<shadedClassifierName>[insert name]</shadedClassifierName>
```

p013570:
Merged into develop.

p013570:
# Migration Steps
Make sure the correct shaded jar name is used. See comment above.

PR #474, opened 2016-10-18
gh-454 - Updated accumulo-store POM to ensure the resulting shaded JAR has the correct name

t616178:


p013570:
I think we are still missing these bits:

```xml
<shadedArtifactAttached>true</shadedArtifactAttached>
<shadedClassifierName>[insert name]</shadedClassifierName>
```

Once you have run a mvn clean install, if you look inside your repository in your .m2 folder, the jar names will still have 'shaded' appended.

a09631:
Both of these elements are now included in the POM.  Does this ticket cover a specific shaded jar name?  Otherwise happy to approve.

t616178:
I've corrected these issues now - the correct artifacts are created in the target/ directories and in the Maven repository.

Issue #359, opened 2016-08-10
Failing tests when building on OpenJDK 1.7.0.111-2.6.7.2.el7_2


a09631:
Running in a CentOS 7 VM, Gaffer fails to build when using OpenJDK 1.7.0.111-2.6.7.2.el7_2, however the problem does not present when building against a 1.8.0 version of OpenJDK.

The tests failing when building on a 1.7.0 release should be updated to ensure they pass using both JDK versions.

p013570:
We should add nightly builds to travis to check the different java versions.

p013570:
We now have Travis cron jobs enabled. Just need to configure it - https://docs.travis-ci.com/user/cron-jobs/

p013570:
It's not possible to run a travis cron build with a different .travis.yml. 

We could use the travis $TRAVIS_EVENT_TYPE variable to check whether the job is being run as a cron job (variable set to 'cron'). This would allow us to skip checks when it runs normally and run extra tests for different build environments when run overnight. This approach is a bit brittle. We probably need to investigate using our own CI server. See gh-477.

t616178:
The updates to Jenkins in issue gh-544 demonstrate that the current Gaffer build passes on Debian8 and Centos 7, under Java 7 and Java 8. Closing as this issue appears to be solved.

PR #476, opened 2016-10-18
gh-359 openjdkbuild

t616178:
Problems with tests resolved by making the AccumuloProperties objects in the test class non-static. I'm not sure why this behaviour would be different under JDK8, but this solves the problem on my machine.

Tested with JDK 1.7.0_111 on Arch Linux.


p013570:
Can you change the pull request to go into the develop branch please.

p013570:
We should add nightly builds to travis to check the different java versions. I'll close the pull request for now and reopen it when we have updated the travis configuration.

Issue #463, opened 2016-10-13
Visibility property not returned from queries in Accumulo Store


gaffer01:
If data with a visibility is inserted into an `AccumuloStore`, with the visibility property set, then it is not returned as part of the properties when a query is made.

Here's an example.

Schema:

```
{
  "entities" : {
    "entityGroup": {
      "vertex": "vertex.string",
      "properties": {
        "myVisibility": "string"
      }
    }
  },
  "types": {
    "vertex.string": {
      "class": "java.lang.String"
    },
    "true": {
      "class": "java.lang.Boolean",
      "validateFunctions": [
        {
          "function": {
            "class": "gaffer.function.simple.filter.IsTrue"
          }
        }
      ]
    },
    "string": {
      "class": "java.lang.String"
    }
  },
  "visibilityProperty": "myVisibility"
}
```

Add this data (note that a visibility of the empty string is a meaningful visibility):

```
Set<Element> elements = new HashSet<Element>();
Entity entity1 = new Entity("entityGroup", "A");
entity1.putProperty("myVisibility", "");
elements.add(entity1);
AddElements addElements = new AddElements.Builder()
                .elements(elements)
                .build();
graph.execute(addElements, new User());
```

Run a query:

```
GetRelatedElements<EntitySeed, Element> get = new GetRelatedElements.Builder<EntitySeed, Element>()
                .addSeed(new EntitySeed("A"))
                .build();
CloseableIterable<Element> results = graph.execute(get, new User());
for (final Element e : results) {
          System.out.println(e);
}
results.close();
```

Then the result is:

```
Entity{vertex=A, group='entityGroup', properties={}}
```

The result should be:

```
Entity{vertex=A, group='entityGroup', properties={myVisibility=<java.lang.String>}} 
```

It's not clear whether this only happens when the visibility is the empty string, or whether it happens with any visibility - further testing is needed.

A unit test should be added to avoid this problem occurring again.

gaffer01:
A quick look at the code suggests that it's because the visibility field is ignored if it's 0 length - it shouldn't be.

t616178:
This will be fixed by issue gh-485

p013570:
Merged into develop.

Issue #466, opened 2016-10-13
Provide visibility serialiser that's compatible with Accumulo's built-in visibility filtering


gaffer01:
We should provide a serialiser for string visibilities that works as you'd expect with Accumulo's built-in visibility filtering approach, e.g. if the visibility is "vis1&vis2" and the user has Accumulo authorisations "vis1" and "vis2", then they should be able to see it. This serialiser should automatically be used if the provided visibility property is a string. It should deal with the valid visibility of the empty string.

Note that it's possible that this already works fine with our standard string serialiser, but there don't seem to be any unit tests that check the combination of visibilities is dealt with correctly.

t616178:
This will be fixed by issue gh-485.

p013570:
Merged into develop.

PR #479, opened 2016-10-19
gh-463 - Modified the "getPropertiesFromColumnVisibility()" method

t616178:
gh-463 - Modified the "getPropertiesFromColumnVisibility()" method to create a default visibility group of "" if the visibility property is set to an empty String or a null value.


t616178:
This likely needs some extra work/thought before being merged in - what's here is a first pass of a solution to #463 and #466.

accumulo-store/src/main/java/gaffer/accumulostore/key/core/AbstractCoreKeyAccumuloElementConverter.java
```diff
@@ -272,8 +270,13 @@ public Properties getPropertiesFromColumnVisibility(final String group, final by
             if (null != propertyDef) {
                 final Serialisation serialiser = propertyDef.getSerialiser();
                 try {
-                    properties.put(schema.getVisibilityProperty(),
-                            serialiser.deserialise(columnVisibility));
+                    final Object columnVisibilityProperty;
+                    if (columnVisibility == null || columnVisibility.length == 0) {
+                        columnVisibilityProperty = "";
```
p013570: The visibility property could be any Object - it may not be a string, so unfortunately we can't default it to an empty string.

t616178:
Abandoning this pull request - this issue will be fixed by PR #488.

PR #488, opened 2016-10-20
Gh 485 serialisers null handling

t616178:


t616178:
This probably needs some more work/testing. I'm having issues building locally, so wanted to start the process off on Travis to see if the same issues are manifest in the CI environment.

t616178:
There are a couple of consequences to the changes that I've made here, so I'm happy to take suggestions if they cause more problems than they solve:

* Null handling behaviour is pushed down from the Accumulo Element converter to the individual serialiser classes. This is to provide enough flexibility to be able to discriminate between null values and empty (or default) values. An AbstractSerialiser class has been added which contains default handling for deserialising null values and for serialising empty values.
* Any Graphs which are created with a visibility property will always result in the visibility property being set on any elements retrieved from that Graph, even if the visibility contains an empty value.
* Element properties have been modified so that it is not permitted to create a property with a null key or a null value.

# Migration required

The format of the Serialiser classes has changed. The serialisers are now generic classes, with the generic parameter being the type of object serialised and the target type for deserialisation.

The Serialiser classes are now also responsible for handling nulls. Default behaviour is provided by the AbstractSerialiser class, which should be extended. The default behaviour is to serialise null values as empty byte arrays and to deserialise empty byte arrays to null values again. In combination with the configuration for the Jackson ObjectMapper used, these null values will not be serialised when written as JSON.

accumulo-store/src/main/java/gaffer/accumulostore/key/core/AbstractCoreKeyAccumuloElementConverter.java
```diff
@@ -272,8 +273,15 @@ public Properties getPropertiesFromColumnVisibility(final String group, final by
             if (null != propertyDef) {
                 final Serialisation serialiser = propertyDef.getSerialiser();
                 try {
-                    properties.put(schema.getVisibilityProperty(),
-                            serialiser.deserialise(columnVisibility));
+                    if (columnVisibility == null || columnVisibility.length == 0) {
+                        if (serialiser.deserialiseEmptyBytes() != null) {
```
p013570: Perhaps store the response from serialiser.deserialiseEmptyBytes() so we don't have to call it twice.

accumulo-store/src/test/java/gaffer/accumulostore/integration/AccumuloVisibilityIT.java
```diff
@@ -0,0 +1,430 @@
+package gaffer.accumulostore.integration;
```
p013570: These tests are really helpful and I think they would be useful for any Store implementation. Could we move them into the gaffer-integration-tests module and set the class to extend AbstractStoreIT. We could then reuse them for all Stores.

p013570: @t616178 did you manage to make this change?

t616178: Yes, this has been implemented now.

PR #492, opened 2016-10-20
Gh 90 remove element component key

p013570:


p013570:
I need to add some schema validation to prevent anyone using VERTEX, SOURCE, DESTINATION or DIRECTED as property names.

PR #498, opened 2016-10-25
Gh 435 tidy up maven dependencies

t616178:


pom.xml
```diff
@@ -108,17 +121,82 @@
...
+                <artifactId>scalap</artifactId>
+                <version>${scala.minor.version}</version>
+            </dependency>
+            <dependency>
+                <groupId>org.xerial.snappy</groupId>
+                <artifactId>snappy-java</artifactId>
+                <version>${snappy.version}</version>
             </dependency>
         </dependencies>
     </dependencyManagement>
```
p013570: As we are making use of the maven dependency management section, we can remove the dependency versions from the other poms that use any of these dependencies.

E.g in serialisation-library/pom.xml
```xml
<dependency>
    <groupId>com.clearspring.analytics</groupId>
    <artifactId>stream</artifactId>
</dependency>
```

gaffer01:
Is there any particular reason for using version 2.2.0 of Hadoop? That's relatively old - sticking with something like 2.6.3 or 2.7.0 might be better?

p013570:
Accumulo 1.7.2 depends on Hadoop 2.2.0 so it is probably best to stick with 2.2.0 until we update the version of Accumulo.

rest-api/pom.xml
```diff
@@ -101,7 +101,7 @@
         <dependency>
             <groupId>com.fasterxml.jackson.datatype</groupId>
             <artifactId>jackson-datatype-json-org</artifactId>
-            <version>2.3.3</version>
+            <version>${jackson.version}</version>
```
p013570: Can we not just remove these versions? As the version within the dependency management section in the parent pom should be automatically used.

t616178: This only works if the exact dependency is listed in the dependency management section - I was aiming for a balance between moving some things to the dependency management section of the parent POM and not having to list every single dependency there.

t616178:
The release notes for Accumulo 1.7.2 state that 2.2.0 or newer is required. In fact, they run the Accumulo integration tests on Hadoop 2.6.3 - it should be fine (and probably preferred) to stick with the latest Hadoop 2.x.x release.

p013570:
Let's leave the hadoop version as it is for now and raise a separate issue to update it.

Issue #455, opened 2016-10-12
Jackson serialisation using JsonTypeInfo needs improving


p013570:
The JsonTypeInfo annotation causes problems with generic types/type erasure. We have a temporary solution that adds a class property to Element and ElementSeed, but it isn't very nice.

p013570:
See comment on pull request - https://github.com/gchq/Gaffer/pull/505/files

p013570:
The changes have caused the "class" field to be missed off again - so we need a different fix.

PR #505, opened 2016-10-26
gh-455 - Simplified polymorphic deserialisation in Element class.

t616178:
These changes to the serialisation of Elements is simpler, more in line with the Jackson documentation and examples and has the same behaviour as the previous solution.

I'm happy to be challenged on this if there's some reason why the more complicated solution was required that I've not understood fully.


gaffer-core/operation/src/test/java/gaffer/operation/impl/add/AddElementsTest.java
```diff
@@ -57,8 +56,7 @@
             "    \"group\" : \"edge type 2\",%n" +
             "    \"source\" : \"source vertex 1\",%n" +
             "    \"destination\" : \"dest vertex 1\",%n" +
-            "    \"directed\" : true,%n" +
-            "    \"class\" : \"gaffer.data.element.Edge\"%n" +
```
p013570: When elements are serialised into JSON they should always have the "class" property, so consumers of the API can check if the element is an Edge or an Entity. Why has it been removed from the JSON here?

p013570: I think we may need to back out the changes.

Issue #439, opened 2016-09-30
AddElementsFromHDFS operation should provide a counter for elements that failed validation


james010101101:
At current there is no way to tell how many Elements failed validation when going an AddElementsFromHDFS operation this would be a useful for users to have.

p013570:
In addition to just logging this warning in AbstractAddElementsFromHdfsMapper.java
```java
LOGGER.warn("Element " + element + " did not validate.");
```

we should update a counter e.g.
```java
context.getCounter("Add Elements From HDFS", "invalid element count").increment(1L);
```

p013570:
Merged into develop.

PR #506, opened 2016-10-26
gh-439 - Added counter to track items which fail validation in the AddElementsFrom Hdfs operation.

t616178:
Naive solution to this issue - let me know if a more elegant way of implementing this is preferred.


p013570:
I think what the issue really wants is to update AbstractAddElementsFromHdfsMapper.java

In addition to just logging a warning 
```java
LOGGER.warn("Element " + element + " did not validate.");
```

we should update a counter e.g. 
```java
context.getCounter("Add Elements From HDFS", "invalid element count").increment(1L);
```

p013570:
I'll close the pull request for now. Just reopen it after the changes have been made.

PR #512, opened 2016-10-31
gh-439 - Add counter to AbstractAddElementsFromHdfsMapper

t616178:


t616178:
Currently, there are no unit tests which confirm that the HDFS counters are working as expected - is this tested elsewhere (e.g. in integration testing somewhere) or are these tests missing?

Would we like to add tests to check that the right counters are being incremented at the right time?

p013570:
Yeah we don't currently have tests for the counters. If possible, it would be good to write some as part of this issue.

PR #532, opened 2016-11-04
Gh 518 function-equals-hashCode-toString

t616178:


core/function/pom.xml
```diff
@@ -37,6 +37,12 @@
             <artifactId>jackson-databind</artifactId>
             <version>${jackson.version}</version>
         </dependency>
+
+        <dependency>
+            <groupId>commons-lang</groupId>
+            <artifactId>commons-lang</artifactId>
+            <version>2.6</version>
```
p013570: It would be better to use the commons-lang.version property.

p013570:
The issue wants toString, equals and hashcode methods on all functions, the view and the schema. It looks like there are a few missing?

t616178:
I've added the missing methods and made sure that properties from base classes are included in the methods (these were being excluded)

p013570:
Changes look good. Just wondering how we force new Function implementations to override the toString, hashcode and equals methods? Could we add a test into ConsumerFunctionTest to somehow check these methods have been implemented and all public fields are included in these methods?

t616178:
I think it would be better to just enforce that developers are expected to provide these methods if appropriate. I've updated the Wiki to reflect this.

gaffer01:
Looks good. Should there be unit tests on the `equals()` methods? Or is that unnecessary?

Issue #493, opened 2016-10-21
accumulo-store tests failing on some environments


t616178:
In my development environment, I am seeing test failures when building the accumulo-store module:

<pre><code>shouldAddElementsFromHdfs(gaffer.accumulostore.integration.AddElementsFromHdfsIT)  Time elapsed: 0.377 sec  <<< ERROR!
gaffer.operation.OperationException: Failed to fetch elements from HDFS
    at java.util.concurrent.FutureTask.report(FutureTask.java:122)
    at java.util.concurrent.FutureTask.get(FutureTask.java:192)
    at org.apache.hadoop.mapred.LocalDistributedCacheManager.setup(LocalDistributedCacheManager.java:139)
    at org.apache.hadoop.mapred.LocalJobRunner$Job.<init>(LocalJobRunner.java:163)
    at org.apache.hadoop.mapred.LocalJobRunner.submitJob(LocalJobRunner.java:731)
    at org.apache.hadoop.mapreduce.JobSubmitter.submitJobInternal(JobSubmitter.java:241)
    at org.apache.hadoop.mapreduce.Job$10.run(Job.java:1297)
    at org.apache.hadoop.mapreduce.Job$10.run(Job.java:1294)
    at java.security.AccessController.doPrivileged(Native Method)
    at javax.security.auth.Subject.doAs(Subject.java:422)
    at org.apache.hadoop.security.UserGroupInformation.doAs(UserGroupInformation.java:1656)
    at org.apache.hadoop.mapreduce.Job.submit(Job.java:1294)
    at org.apache.hadoop.mapreduce.Job.waitForCompletion(Job.java:1315)
    at gaffer.accumulostore.operation.hdfs.handler.job.tool.FetchElementsFromHdfsTool.run(FetchElementsFromHdfsTool.java:55)
    at org.apache.hadoop.util.ToolRunner.run(ToolRunner.java:70)
    at org.apache.hadoop.util.ToolRunner.run(ToolRunner.java:84)
    at gaffer.accumulostore.operation.hdfs.handler.AddElementsFromHdfsHandler.fetchElements(AddElementsFromHdfsHandler.java:61)
    at gaffer.accumulostore.operation.hdfs.handler.AddElementsFromHdfsHandler.doOperation(AddElementsFromHdfsHandler.java:45)
    at gaffer.accumulostore.operation.hdfs.handler.AddElementsFromHdfsHandler.doOperation(AddElementsFromHdfsHandler.java:40)
[...]

PR #537, opened 2016-11-09
gh-493 - Move tests to a temporary directory with correct access permissions

t616178:
Added configuration to select system default temporary folder.


store-implementation/accumulo-store/src/test/java/gaffer/accumulostore/utils/AccumuloTestConstants.java
```diff
@@ -0,0 +1,28 @@
...
+ * limitations under the License.
+ */
+
+package gaffer.accumulostore.utils;
+
+import java.io.File;
+
+public class AccumuloTestConstants {
+
+    public static final File tmpDirectory = new File(System.getProperty("java.io.tmpdir"));
```
p013570: Please could you change this constant to be in uppercase for consistency - e.g TMP_DIRECTORY
This will be useful for other modules too, so it might be better in a CommonTestConstants class in the CommonUtil module.

Issue #444, opened 2016-10-06
Add date/time utilities to common-util


m316257:
Gaffer is about summarisation, often over time and users often need to add time buckets with start and end times or dates on their edges and entities. 

Adding some optimised utility methods to help with this inside Gaffer would help a lot.

p013570:
What methods shall we add?

t616178:
I'm going to add some of the methods from Gaffer1, as well as some methods to carry out "bucketisation" of time fields.

p013570:
Merged into develop.

PR #541, opened 2016-11-10
gh-444 - Added CommonTimeUtil class

t616178:
Started the CommonTimeUtils class with a rough method for creating buckets of time from long timestamps.

The joda-time library has been added and replaces the functionality that was in the Gaffer1 TimeUtils class.

### DIFF
--- a/core/common-util/pom.xml
+++ b/core/common-util/pom.xml
@@ -34,6 +34,11 @@
             <artifactId>reflections</artifactId>
         </dependency>
         <dependency>
+            <groupId>joda-time</groupId>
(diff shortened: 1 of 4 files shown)


core/common-util/pom.xml
```diff
@@ -34,6 +34,11 @@
             <artifactId>reflections</artifactId>
         </dependency>
         <dependency>
+            <groupId>joda-time</groupId>
```
p013570: Can you add this new dependency to the NOTICES file please.

t616178: Done

Issue #360, opened 2016-08-10
Specifying split file in Accumulo bulk import


gaffer01:
The split file for bulk imports to Accumulo is currently specified in two places: in the store properties (option `accumulo.splits.file.path`) and as an option on the `AddElementsFromHdfs` operation (`accumulostore.operation.hdfs.user_provided_splits_file`). The first case is used when the operation will create the splits file from the split points that are currently on the table; the latter is used when the user provides the split file (as should happen when the user samples the data to find sensible split points prior to the first import to the table).
- It is confusing that the splits file is specified in two different places depending on a subtle difference in use.
- The splits file does not belong in the store properties - it is a property of a particular bulk import operation, not the store. There might be multiple bulk import operations happening at the same time and in this case if the same store properties file was used then the files would clash. When this happens an IOException with the message "Resource hdfs://server/splits_file changed on src filesystem (expected ____ was ____)".

Suggest:
- The splits file should be removed from the store properties, and should always be specified by the option on the operation.
- There should be an option on the operation to say whether the splits are to be found from the table and written to the splits file, or to be taken from the user provided splits file.

t616178:
Merged into develop.

p013570:
Merged into develop.

t616178:
# Migration steps

The method for specifying a splits file for Accumulo has changed.

The following options are required to set a splits file when carrying out a bulk import:

* accumulostore.operation.hdfs.use_provided_splits_file - a boolean flag, set to the String "true" or "false" 
* accumulostore.operation.hdfs.splits.file_path - String representing the file path to use

If "use_provided_splits_file" flag is true, the file path must be provided - this will be used to determine the splits.

If "use_provided_splits_file" flag is false, the file path must be provided - this will be used to determine where to write the generated splits file to.

t616178:
Reopened to solve issues with undefined states.

t616178:
Merged into develop.

PR #559, opened 2016-11-24
gh-360 - Updated the partitioner setup to handle all different combin…

t616178:
…ations of properties being set/unset


t616178:
This PR attempts to fix a bug (or at best some undefined behaviour) where there are some combinations of AccumuloStoreProperties which leave the system in an undefined state when setting up the Accumulo partitioner.

t616178:
Removing due to renamed branch.

Issue #562, opened 2016-11-24
Operation chain limiter can fail if operation scores are loaded in an unexpected order


t616178:
The operation chain limiter appears to require operation scores to be listed in ascending order, however it is possible for these to be loaded in a random order.

This causes tests to fail, as the operation authentication checks are not carried out in the correct order.

When loaded from disk, it must be ensured that the resulting Map is always ordered.

t616178:
Fixed and merged into gh-434

PR #563, opened 2016-11-28
Gh 434 update group

t616178:
There are a couple of issues which need to be fixed (one ignored test - will be fixed by gh-562), but it may be best to merge this change quickly to iron out any other problems that aren't identified by the unit tests (and to try to organise the process of releasing the artifacts to Maven Central).

I'll be working on these related issues straight away - let me know if there are any comments/concerns.


t616178:
gh-562 closed and issue fixed.

PR #566, opened 2016-11-29
Gh 514 accumulo without aggregators

t616178:


gaffer01:
I think there should be a unit test of what happens when the same element is added multiple times, if there is no aggregation. If the same element is added twice, it should be returned twice.

t616178:
@gaffer01 Agreed - I've added a test to AccumuloAggregationIT to confirm that this is the case.

PR #571, opened 2016-12-04
Gh 142 codecov v2 - testing

t616178:
Second attempt to add codecov integration.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,5 +11,8 @@ addons:
       - oracle-java8-installer
 install: true
 env:
-  - MAVEN_OPTS="-XX:+TieredCompilation -XX:TieredStopAtLevel=1"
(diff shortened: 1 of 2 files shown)


t616178:
The following changes have been applied in order to get codecov integration working:

* A new profile has been created (-Ptravis), specifically for building Gaffer on Travis. This is required since the previous profile used in the Travis build process required setting the Maven Surefire plugin's forkCount property to 0, which breaks JaCoCo and prevents the test reports from being generated.
* Since the forkCount property must be set to 1, the tests fail due to memory issues in the Travis containerised environment. To resolve this, fork reuse has been disabled and the amount of heap memory available to the Java process has been increased (-Xmx3072m).

If there are any issues with the Travis build process in future (entirely possible, as I'm not sure how marginal the build process is on memory at the moment), this approach may need to be revisited.

Issue #224, opened 2016-05-24
Create day based age off filter function


p013570:
Similar to [AgeOff](https://github.com/GovernmentCommunicationsHeadquarters/Gaffer/blob/master/simple-function-library/src/main/java/gaffer/function/simple/filter/AgeOff.java) filter function but this function should take 2 properties:
- timestamp - the time the element was added to the graph
- days - the number of days to keep the element

e.g:

``` java
@Inputs({Long.class, Integer.class})
public class AgeOffFromDays extends FilterFunction {

    @Override
    public boolean isValid(final Object[] input) {
        // Validate input
        ....

        final Long timestamp = (Long) input[0];
        final Integer days = (Long) input[1];

        // Check timestamp is within specified number of days
        ....
    }

    ...
}
```

p013570:
Merged into develop.

PR #572, opened 2016-12-04
gh-224 - Added AgeOffFromDays filter

t616178:


library/function-library/src/main/java/uk/gov/gchq/gaffer/function/simple/filter/AgeOffFromDays.java
```diff
@@ -0,0 +1,115 @@
...
+@Inputs({Long.class, Integer.class})
+public class AgeOffFromDays extends FilterFunction {
+    public static final long DAYS_TO_MILLISECONDS = 24L * 60L * 60L * 1000L;
+
+    /**
+     * The default age of time (1 year) in days.
+     */
+    public static final long AGE_OFF_TIME_DEFAULT = 365L;
+
+    private long ageOffTime = AGE_OFF_TIME_DEFAULT;
```
p013570: I don't think this field is needed.

t616178: Removed.

Issue #575, opened 2016-12-07
Unexpected behaviour in serialisation classes


t616178:
After the changes implemented in gh-485, some of the serialisation classes are not correctly implemented. This leads to some Element properties being skipped during deserialisation.

This affects 0.4.6 and later.

p013570:
Merged into hotfix-0.4.6.1

PR #576, opened 2016-12-07
gh-575 - Apply fix to serialisation classes

t616178:


serialisation-library/src/main/java/gaffer/serialisation/simple/TypeValueSerialiser.java
```diff
@@ -88,6 +90,11 @@ public TypeValue deserialise(final byte[] bytes) throws SerialisationException {
     }
 
     @Override
+    public TypeValue deserialiseEmptyBytes() throws SerialisationException {
+        return new TypeValue();
```
p013570: I think this should be null. In the serialise method, if the type value is blank an exception is thrown - so 'new TypeValue()' would not be able to be reserialised.

Issue #419, opened 2016-09-21
GetAllElements builder offers invalid method


gaffer01:
The `Builder` for `GetAllElements` offers an `addSeed` method but this is not valid for this operation. As the only reason for the existence of the builder is to make it easy to create a valid instance of the operation, it should not offer this method.

Also:
-  `GetAllElements` is a `GetElements<ElementSeed>` - shouldn't it be a `GetElements<Void>`?
- The `setIncludeIncomingOutGoing` method currently says that it is only possible to set this to outgoing edges only. This doesn't make sense - there is no concept of outgoing for `GetAllElements`. It would be better if this method threw an `IllegalArgumentException` (even if under the hood it uses that option to avoid returning the same edge twice).

t616178:
GetElements requires a generic type which extends ElementSeed, so it's not possible to create an object like GetElements<Void>.

If the main issue is that some operations specify invalid methods, it might be necessary to refactor the operation classes to move some of the methods into interfaces, e.g.

<pre><code>
GetElements extends AbstractGetIterableElementsOperation<ElementSeed, Element>
GetEntities extends GetElements implements WithSeed<ElementSeed> etc
</code></pre>

The alternative is to document which methods are supported for each operation and to throw exceptions if an unsupported one is used.

Is there any preference on this?

gaffer01:
Similarly, `GetAdjacentEntitySeeds` has some options that don't make sense, such as `populateProperties` and `includeEntities`.

p013570:
These options do really increase the complexity of implementing a new Store. Perhaps we could remove some options or make some of them optional.

p013570:
## Migration Steps
The following options have been removed:
- populateProperties (this never used to do anything)
- includeEntities (the view should be used to control if entities are required)
- includeEdges (the view should be used to control if edges are required)
- limit (users will now have to use the Limit operation)
- deduplicate (users will now have to use the Deduplicate operation)

A 'directedType' option has been added to replace includeEdges, this allows only directed or undirected edges to be filtered.

The hierarchy of some of the operations has also changed to try and avoid operations having illegal options.

t616178:
Merged into develop-0.7.0

PR #603, opened 2016-12-14
gh-419 - Overriden methods in GetAllElements and ensured they throw e…

t616178:
…xceptions if attempting to use the GetAllElements Builder in an unspecified way.

### DIFF
--- a/core/operation/src/main/java/uk/gov/gchq/gaffer/operation/impl/get/GetAllElements.java
+++ b/core/operation/src/main/java/uk/gov/gchq/gaffer/operation/impl/get/GetAllElements.java
@@ -55,6 +55,11 @@ public class GetAllElements<ELEMENT_TYPE extends Element>
     }
 
     @Override
+    public CloseableIterable<ElementSeed> getSeeds() {
+        return null;
+    }
+
+    @Override
     public void setSeeds(final CloseableIterable<ElementSeed> seeds) {
         if (null != seeds) {
             throw new IllegalArgumentException("This operation does not allow seeds to be set");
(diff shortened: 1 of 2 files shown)


core/operation/src/main/java/uk/gov/gchq/gaffer/operation/impl/get/GetAllElements.java
```diff
@@ -95,6 +95,38 @@ public void setIncludeIncomingOutGoing(final IncludeIncomingOutgoingType include
...
+        public CHILD_CLASS seeds(final CloseableIterable<ElementSeed> seeds) {
+            if (null != seeds) {
+                throw new IllegalArgumentException("This operation does not allow seeds to be set");
+            }
+            return self();
+        }
+
+        @Override
+        public CHILD_CLASS addSeed(final ElementSeed seed) {
+            if (null != seed) {
```
p013570: even a null seed would actually be invalid.

gaffer01:
Can the other problems from the issue be dealt with at the same time or should they be a separate issue?

- `GetAllElements` is a `GetElements` of `ElementSeed`  - shouldn't it be a `GetElements` of `Void`?
- The `setIncludeIncomingOutGoing` method currently says that it is only possible to set this to outgoing edges only. This doesn't make sense - there is no concept of outgoing for `GetAllElements`. It would be better if this method threw an `IllegalArgumentException` (even if under the hood it uses that option to avoid returning the same edge twice).

t616178:
Closing to continue working.

Issue #610, opened 2016-12-20
Add CollectionContains filter to filter library


t616178:

t616178:
Merged into develop.

PR #611, opened 2016-12-20
gh-610 - Added CollectionContains filter to filter library

t616178:


library/core-library/src/main/java/uk/gov/gchq/gaffer/function/filter/CollectionContains.java
```diff
@@ -0,0 +1,93 @@
...
+import org.apache.commons.lang.builder.EqualsBuilder;
+import org.apache.commons.lang.builder.HashCodeBuilder;
+import org.apache.commons.lang.builder.ToStringBuilder;
+import uk.gov.gchq.gaffer.function.SimpleFilterFunction;
+import uk.gov.gchq.gaffer.function.annotation.Inputs;
+import java.util.Collection;
+import java.util.Map;
+
+/**
+ * An <code>MapContains</code> is a {@link uk.gov.gchq.gaffer.function.SimpleFilterFunction} that checks
```
gaffer01: Should say `CollectionContains`

Issue #516, opened 2016-11-03
Add an example of adding vertex cardinalities to the Getting Started guide


p013570:
This basically involves adding a 'Cardinality' Entity onto every vertex. The 'Cardinality' entity needs to have a HyperLogLogPlus property. Then in your element generator, whenever an edge (A to B) is created, it should also create cardinality entities at A and B.

These cardinality entities will then be aggregated automatically and will provide an estimate of the connectivity of each vertex in the graph.

To create the cardinality entities for an edge, you can do something like:
```java
final HyperLogLogPlus hllpSource = new HyperLogLogPlus(5, 5);
hllpSource.offer(edge.getDestination());
final Entity sourceCardinality = new Entity.Builder()
        .vertex(edge.getSource())
        .group("Cardinality")
        .property("edgeGroup", edge.getGroup())
        .property("hllp", hllpSource)
        .build();

final HyperLogLogPlus hllpDest = new HyperLogLogPlus(5, 5);
hllpSource.offer(edge.getSource());
final Entity destCardinality = new Entity.Builder()
        .vertex(edge.getDestination())
        .group("Cardinality")
        .property("edgeGroup", edge.getGroup())
        .property("hllp", hllpDest)
        .build();
```

gaffer01:
Merged into develop.

PR #631, opened 2017-01-06
gh-516 - added cardinality example

p013570:


example/example-graph/src/main/resources/example/gettingstarted/9/walkthrough.md
```diff
@@ -0,0 +1,54 @@
+${HEADER}
+
+${CODE_LINK}
+
+This example demonstrates how we can use a HyperLogLogPlus object to estimate the cardinality of vertices in a graph.
```
gaffer01: Can we be more precise here? It should say "This example demonstrates how storing a HyperLogLogPlus object on each vertex in a graph allows us to quickly estimate its degree (i.e. the number of edges it is involved in). The estimate can be obtained without scanning through all the edges involving a node and so is very quick, even if the degree is very large."

p013570:
@gaffer01 - I've updated it with your suggested changes. Thanks.

Issue #634, opened 2017-01-08
Update Copyright date to 2017


p013570:
This will also require an update to the checkstyle header format

p013570:
They will be updated as we change files to 2016-2017.

PR #635, opened 2017-01-09
gh-634 - Updated copyright header from 2016 to 2017.

t616178:


t616178:
Changing approach - will reopen when finished.

PR #649, opened 2017-01-13
Gh 499 rest chunking

p013570:


example/example-rest/pom.xml
```diff
@@ -15,19 +15,25 @@
...
         <version>0.5.2-SNAPSHOT</version>
     </parent>
     <modelVersion>4.0.0</modelVersion>
     <artifactId>example-rest</artifactId>
-    <properties>
-        <jetty.version>9.2.11.v20150529</jetty.version>
-    </properties>
     <packaging>war</packaging>
     <dependencies>
+        <!-- Trick to prevent Maven from including this dependency in the WAR -->
```
t616178: I don't think this is required any more.

example/example-rest/pom.xml
```diff
@@ -73,46 +79,48 @@
...
             <id>standalone</id>
             <build>
                 <plugins>
                     <plugin>
-                        <groupId>org.eclipse.jetty</groupId>
-                        <artifactId>jetty-maven-plugin</artifactId>
-                        <version>${jetty.version}</version>
+                        <groupId>org.apache.tomcat.maven</groupId>
+                        <artifactId>tomcat7-maven-plugin</artifactId>
+                        <version>2.2</version>
```
t616178: Could this be moved to a property?

rest-api/src/main/java/uk/gov/gchq/gaffer/rest/service/IGraphConfigurationService.java
```diff
@@ -66,7 +65,7 @@
     @GET
     @Path("/operations")
     @ApiOperation(value = "Gets all operations supported by the store. See <a href='https://github.com/gchq/Gaffer/wiki/operation-examples' target='_blank' style='text-decoration: underline;'>Wiki</a>.", response = Class.class, responseContainer = "list")
-    Set<Class<? extends Operation>> getOperations();
+    Set<Class> getOperations();
```
t616178: Should we be using an unbounded type parameter here, rather than a raw type?

rest-api/src/main/java/uk/gov/gchq/gaffer/rest/service/SimpleGraphConfigurationService.java
```diff
@@ -147,12 +146,12 @@ public Schema getSchema() {
     }
 
     @Override
-    public Set<Class<? extends Operation>> getOperations() {
-        return graphFactory.getGraph().getSupportedOperations();
+    public Set<Class> getOperations() {
```
t616178: Same comment here regarding raw types?

rest-api/src/test/java/uk/gov/gchq/gaffer/rest/service/SimpleOperationServiceIT.java
```diff
@@ -0,0 +1,133 @@
+/*
+ * Copyright 2016 Crown Copyright
```
t616178: Do you need to update your copyright header template?

p013570:
Travis CI is killing the job. It is either using too much memory or using too many threads.

This is most likely due to the addition of the integration tests and requiring a server with the Gaffer REST API to to be started. However, the chunking rest api code also creates new threads.

rest-api/src/main/java/uk/gov/gchq/gaffer/rest/service/IGraphConfigurationService.java
p013570: I don't think REST API methods should return generic types - you get warnings when the REST API starts up if any are methods that return a generic type.

p013570:
I've removed the integration tests from Travis CI, they will still be run in Jenkins. This will have a big impact on our test coverage.

Issue #650, opened 2017-01-13
Align Jackson version with Hadoop


t616178:
Users have reported issues when executing operations using Hadoop due to conflicting versions of Jackson present on the classpath.

By changing the Jackson version used in Gaffer, we can simplify the process of integrating with Hadoop.

p013570:
Merged into develop.

PR #651, opened 2017-01-13
gh-650 - Update Jackson version to align with Hadoop

t616178:


core/data/src/main/java/uk/gov/gchq/gaffer/data/element/Element.java
```diff
@@ -42,7 +42,7 @@
  * <p>
  * Equals has been overridden to check groups are equal. NOTE - it does not compare property values.
  */
-@JsonTypeInfo(use = Id.CLASS, include = As.EXISTING_PROPERTY, property = "class")
+@JsonTypeInfo(use = Id.CLASS, include = As.PROPERTY, property = "class")
```
p013570: I think this needs to stay as EXISTING_PROPERTY - there is a getter and setter for "class" at the bottom of this class. If you set it to just PROPERTY you might end up with duplicate "class" fields in the json.

core/operation/src/main/java/uk/gov/gchq/gaffer/operation/data/ElementSeed.java
```diff
@@ -34,7 +34,7 @@
  * @see EntitySeed
  * @see EdgeSeed
  */
-@JsonTypeInfo(use = Id.CLASS, include = As.EXISTING_PROPERTY, property = "class")
+@JsonTypeInfo(use = Id.CLASS, include = As.PROPERTY, property = "class")
```
p013570: I think this needs to stay as EXISTING_PROPERTY - see other comment.

core/data/src/main/java/uk/gov/gchq/gaffer/data/element/Element.java
p013570: I realise the reason for changing it from EXTERNAL_PROPERTY, as EXTERNAL_PROPERTY doesn't work with databind in the older version of jackson.

However, changing to PROPERTY may cause duplicate "class" properties in the JSON.

Due to serialising and deserialising generic types we had an issue where the "class" field was being completely missed off. To ensure it is always present we added a getter and setter for "class". The EXISTING_PROPERTY annotation then just relies on the getter and setter being used.

pom.xml
```diff
@@ -54,7 +54,7 @@
         <accumulo.version>1.7.2</accumulo.version>
         <avro.version>1.7.7</avro.version>
         <hadoop.version>2.6.5</hadoop.version>
-        <jackson.version>2.6.5</jackson.version>
+        <jackson.version>2.3.1</jackson.version>
```
p013570: Using this version is now not compatible with the version of scala/spark - this exception is thrown:
com.fasterxml.jackson.databind.JsonMappingException: Jackson version is too old 2.3.1
	at com.fasterxml.jackson.module.scala.JacksonModule$class.setupModule(JacksonModule.scala:56)
	at com.fasterxml.jackson.module.scala.DefaultScalaModule.setupModule(DefaultScalaModule.scala:19)
	at com.fasterxml.jackson.databind.ObjectMapper.registerModule(ObjectMapper.java:537)
	at org.apache.spark.rdd.RDDOperationScope$.<init>(RDDOperationScope.scala:82)
	at org.apache.spark.rdd.RDDOperationScope$.<clinit>(RDDOperationScope.scala)
	... 37 more

p013570:
Closing this as I don't think moving to Jackson version 2.3.1 is a good idea as it is quite old and we lose several a few features. Also it would be incompatible with several libraries such as spark.

PR #676, opened 2017-02-01
Gh 671 upgrade spark version

gaffer01:


p013570:
We should probably put spark in the NOTICES file.

gaffer01:
I've updated the NOTICES file as suggested.

PR #681, opened 2017-02-02
Gh 673 abstract accumulo two set seeded operation seeds b serialises using a wrapper object

james010101101:


store-implementation/accumulo-store/src/main/java/uk/gov/gchq/gaffer/accumulostore/operation/AbstractAccumuloTwoSetSeededOperation.java
```diff
@@ -48,7 +47,6 @@ public AbstractAccumuloTwoSetSeededOperation(final Iterable<SEED_TYPE> seedsA, f
         this.setSeedsB(seedsB);
     }
 
-    @JsonTypeInfo(use = JsonTypeInfo.Id.CLASS, include = JsonTypeInfo.As.WRAPPER_OBJECT)
     @JsonGetter(value = "seedsB")
```
p013570: I think it should be the same annotation as seeds in AbstractGetOperation:
@JsonTypeInfo(use = JsonTypeInfo.Id.CLASS, include = JsonTypeInfo.As.PROPERTY, property = "class")

james010101101: The tests show it works without the annotation and the class property is already added in.

Issue #659, opened 2017-01-19
Consistent src and dest in undirected edges


gaffer01:
An undirected edge A-B is identical to the undirected edge B-A, and the `equals()` method on `Edge` will return true in this case. However when `toString()` is called the string representation of the edge will be different. The JSON version of the edge will also be different. In older versions of Gaffer, undirected edges were always created in a consistent manner, with the source being lexicographically before the destination. As the source and destination are now just objects, there may be no natural ordering, but `toString()` can be used instead.

When the edge is created the logic would be:

    if source and destination are both instances of comparable:
         if destination < source
              swap destination and source
    else
          if destination.toString() < source.toString()
              swap destination and source

This should make it easier for non-Java consumers of the data to dedupe data.

t616178:
This is difficult to achieve in practice due to the many ways in which is possible to construct an Edge object - particularly since it is possible to build an Edge incrementally using a no-arg constructor and setter methods.

When would it be necessary for the representation to be modified? Is this something we could implement only when the toString() method is called or when the Edge is serialised?

t616178:
# Migration steps

javadev001001:
@t616178 can you confirm that you will still be able to use the putProperty method, and that the only immutable parts of these objects are the core parts?

p013570:
We should do issue #906 first.

p013570:
@javadev001001 - yes you will still have access to putProperty. I've updated the migration steps above to explain that it is just the setSource, setDestination and setDirected methods that have been removed.

t616178:
Merged into develop.

PR #688, opened 2017-02-06
gh-659 - Provided implementation to ensure Edges are created in a con…

t616178:
…sistent way

### DIFF
--- a/core/data/src/main/java/uk/gov/gchq/gaffer/data/element/Edge.java
+++ b/core/data/src/main/java/uk/gov/gchq/gaffer/data/element/Edge.java
@@ -53,6 +53,7 @@ public class Edge extends Element {
         this.source = source;
         this.destination = destination;
         this.directed = directed;
+        standardise();
     }
 
     @JsonTypeInfo(use = JsonTypeInfo.Id.CLASS, include = JsonTypeInfo.As.WRAPPER_OBJECT, property = "class")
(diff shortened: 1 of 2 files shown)


core/data/src/main/java/uk/gov/gchq/gaffer/data/element/Edge.java
```diff
@@ -113,6 +116,16 @@ public void putIdentifier(final IdentifierType identifierType, final Object prop
         }
     }
 
+    private void standardise() {
+        if (null != destination && null != source) {
```
gaffer01: Would it be better to test to see if they're both comparable, and if so, cast them to comparable and use their natural ordering? That way if the vertices are ints or longs they will sort in a more natural order (with the source less than the destination preferably).

core/data/src/test/java/uk/gov/gchq/gaffer/data/element/EdgeTest.java
```diff
@@ -290,6 +290,25 @@ public void shouldReturnFalseForEqualsWhenDirectedIdentifiersFlipped() {
...
+
+        // Given
+        final Edge edge2 = new Edge("group");
+        edge2.setSource("dest vertex");
+        edge2.setDestination("source vertex");
+        edge2.setDirected(false);
+
+        // Then
+        assertEquals(edge1, edge2);
+        assertEquals(edge1.toString(), edge2.toString());
```
gaffer01: Is it worth adding a test that `edge1.getSource` equals `edge2.getSource`? The test of the `toString()` method partly tests that but it might be better to test it explicitly.

t616178:
Closing while changes are being made to Edge comparison and creation.

PR #701, opened 2017-02-10
Gh 677 move classes from spark accumulo library that are not accumulo specific

gaffer01:


library/spark/spark-library/src/main/java/uk/gov/gchq/gaffer/spark/operation/dataframe/FiltersToOperationConverter.java
```diff
@@ -13,7 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
```
a09631: Update copyright header

gaffer01:
Copyright headers updated as requested.

Issue #702, opened 2017-02-14
Add Aggregation Function to Bitmap Library


james010101101:
Bitmap library does not have a companion Aggregator for its serialiser currently.

p013570:
# Migration Steps
Package name for the bitmap library has changed from gaffer.bitmap to uk.gov.gchq.gaffer.bitmap

p013570:
Merged into develop.

PR #705, opened 2017-02-17
gh-702 RoaringBitmap aggregation function

james010101101:
Change package to have uk.gov.gchq prefix and add a RoaringBitmap aggregation function.

Adding shaded "deploy" jars build instructions to the pom of the bitmap and sketches library, so these can be used alongside core gaffer without other dependencies.

### DIFF
--- a/library/bitmap-library/src/main/java/uk/gov/gchq/gaffer/bitmap/function/aggregate/RoaringBitmapAggregator.java
+++ b/library/bitmap-library/src/main/java/uk/gov/gchq/gaffer/bitmap/function/aggregate/RoaringBitmapAggregator.java
@@ -1,0 +1,51 @@
+/*
+ * Copyright 2017 Crown Copyright
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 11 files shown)


library/bitmap-library/src/main/java/uk/gov/gchq/gaffer/bitmap/function/aggregate/RoaringBitmapAggregator.java
```diff
@@ -0,0 +1,51 @@
...
+public class RoaringBitmapAggregator extends SimpleAggregateFunction<RoaringBitmap> {
+    private RoaringBitmap result = null;
+
+    @Override
+    protected void _aggregate(final RoaringBitmap input) {
+        if (null == input) {
+            return;
+        }
+        if (null == result) {
+            result = input;
```
p013570: If possible we should clone the input instance like we do with other aggregators (e.g SetAggregator)

p013570:
I've changed the aggregator so it now clones the inputs.

Issue #449, opened 2016-10-10
Build fails with Accumulo 1.8.0


gaffer01:
Changing the version of Accumulo to 1.8.0 causes errors in `accumulo-store/src/test/java/gaffer/accumulostore/integration/performance/BloomFilterIT.java`:
- `method openWriter in class FileOperations cannot be applied to given types`
- `method openReader in class FileOperations cannot be applied to given types`

These should be minor issues, and it should be possible to support both Accumulo 1.7.\* and 1.8.0 with the same code.

james010101101:
It Looks like the API to read/write RFiles is a breaking change from Accumulo 1.7-1.8 and we need to remove this test to build against both.

p013570:
Merged into develop

p013570:
This should have been closed as it was released as part of version 0.6.3.

PR #715, opened 2017-02-21
gh-449 - Updated BloomFilterIT to work with Accumulo 1.8

t616178:
Applied required changes to upgrade Accumulo to v1.8.0


p013570:
I don't think we are ready to upgrade to Accumulo 1.8 yet. It would be good if we could compile against 1.7 and 1.8. I'll close this pull request for now.

Issue #685, opened 2017-02-03
Remove usages of internal sun.* APIs


t616178:
There are a number of places where we use propriety Sun APIs. These generate compiler warnings and are subject to deprecation at some point in the future. Usage of these APIs should be removed if possible.

p013570:
Merged into develop.

PR #716, opened 2017-02-21
gh-685 - Removed usages of sun.* APIs

t616178:


core/serialisation/src/main/java/uk/gov/gchq/gaffer/jsonserialisation/JSONSerialiser.java
```diff
@@ -185,7 +185,7 @@ public void serialise(final Object object, final JsonGenerator jsonGenerator, fi
      */
     public <T> T deserialise(final InputStream stream, final Class<T> clazz) throws SerialisationException {
         try (final InputStream stream2 = stream) {
-            final byte[] bytes = IOUtils.readFully(stream2, stream.available(), true);
+            final byte[] bytes = org.apache.commons.io.IOUtils.toByteArray(stream2);
```
p013570: We don't need to use the fully qualified class name here.

Issue #747, opened 2017-03-10
Use HK2 Binders to inject service and factory implementations for REST API


t616178:
Using HK2's dependency injection framework (via classes which extend AbstractBinder) will make it simpler for users to provide their own concrete implementations of classes used in the REST API (e.g. UserFactories, GraphFactories etc)

p013570:
Merged into develop.

PR #748, opened 2017-03-10
gh-747 - Refactored the REST API to use HK2 binders to inject factory instances

t616178:


rest-api/src/main/java/uk/gov/gchq/gaffer/rest/application/ApplicationConfig.java
```diff
@@ -41,6 +42,7 @@ public ApplicationConfig() {
         addServices();
         setupBeanConfig();
         registerClasses(resources);
+        register(new FactoriesBinder());
```
p013570: By having to register this in the ApplicationConfig anyone wanting to register a different User or Graph Factory, they will need to override this application config class, remove FactoriesBinder and add their own FactoriesBinder class. It seems like a lot more code than what they currently have to do. Currently users can just set the property class name for the factories.

Issue #719, opened 2017-02-22
Update wiki now that artifacts are hosted in Maven Central


t616178:
There are a number of places in the Wiki which have not been updated since we began hosting Gaffer JARs in Maven Central - these pages should be updated.

p013570:
Merged into develop.

PR #753, opened 2017-03-14
gh-719 - Updated wiki to include reference to JARs hosted in Maven central

t616178:


example/example-graph/src/main/resources/example/gettingstarted/intro.md
```diff
@@ -13,15 +13,29 @@ To do this you need to do a few things:
 
 When you've done these things you can write java applications to load and query the data.
 
-But before that you will need to download the code and compile it - sorry we are not yet publishing our artifacts to maven central.
+Gaffer is hosted on [Maven Central](https://mvnrepository.com/search?q=uk.gov.gchq.gaffer) and can easily be incorporated into your own projects.
+
+To use Gaffer on Accumulo the only required dependency is the accumulo-store dependency:
```
p013570: and Graph. See gh-755 - perhaps we could fix that bug at the same time?

p013570: and Graph. See gh-755 - perhaps we could fix that bug at the same time?

t616178: Fixed

Issue #757, opened 2017-03-14
Invalid schema not detected


gaffer01:
With the schema below, if the same edge is inserted twice, its properties silently disappear (if only one instance of the edge is inserted the properties are present). This schema should not be allowed because aggregation should (currently) be specified for all properties or none, but no error is given.

    Schema schema = new Schema.Builder()
               .type("intnoagg", new TypeDefinition.Builder()
                       .clazz(Integer.class)
                       .build())
               .type("int", new TypeDefinition.Builder()
                       .clazz(Integer.class)
                       .aggregateFunction(new Sum())
                       .build())
               .type("string", new TypeDefinition.Builder()
                       .clazz(String.class)
                       .aggregateFunction(new StringConcat())
                       .build())
               .type("boolean", Boolean.class)
               .edge("EDGE", new SchemaEdgeDefinition.Builder()
                       .source("string")
                       .destination("string")
                       .directed("boolean")
                       .property("p", "intnoagg")
                       .build())
               .entity("ENTITY", new SchemaEntityDefinition.Builder()
                       .vertex("string")
                       .property("p2", "int")
                       .build())
               .build();

t616178:
Merged into develop.

PR #759, opened 2017-03-14
gh-757 - fixed schema validation

p013570:


core/graph/src/test/java/uk/gov/gchq/gaffer/graph/GraphTest.java
```diff
@@ -489,11 +493,50 @@ public void shouldThrowExceptionIfStoreClassPropertyIsNotSet() throws OperationE
                     .addSchema(new Schema())
                     .storeProperties(new StoreProperties())
                     .build();
+            fail("exception expected");
         } catch (final IllegalArgumentException e) {
             assertEquals("The Store class name was not found in the store properties for key: " + StoreProperties.STORE_CLASS, e.getMessage());
         }
     }
 
+    @Test
```
t616178: Could use JUnit's exception checking here?

e.g. @Test(expected = SchemaException.class)

p013570: I tend to not use that annotation as it will only check that somewhere in the test method an exception is thrown. Using try/catch allows you to test a precise method throws an exception and to validate the exception has the correct message.

Issue #754, opened 2017-03-14
Unused fields in AccumuloStoreConstants


gaffer01:
`AccumuloStoreConstants` has several unused fields: TRANSFORM_PRIORITY, GAFFER_UTILS_TABLE, SCHEMA_KEY, KEY_PACKAGE_KEY. These should be deleted.

p013570:
Merged into develop.

PR #760, opened 2017-03-14
gh-754 - Removed unused constants

t616178:


store-implementation/accumulo-store/src/main/java/uk/gov/gchq/gaffer/accumulostore/utils/AccumuloStoreConstants.java
```diff
@@ -69,8 +67,6 @@
     public static final int COLUMN_QUALIFIER_AGGREGATOR_ITERATOR_PRIORITY = 36;
     // Applied only during scans.
     public static final int ELEMENT_POST_AGGREGATION_FILTER_ITERATOR_PRIORITY = 37;
-    // Applied only during scans.
-    public static final int TRANSFORM_PRIORITY = 50;
```
p013570: This constant isn't used but it is useful as you can see the order the iterators are applied.

Issue #756, opened 2017-03-14
NPE if table is not set in AccumuloStore


gaffer01:
This code throws a NullPointerException because the table name has not been set (if a line is added to set the table name on properties it's fine) - it should throw a more useful exception.

     Schema schema = new Schema.Builder()
                .type("int", Integer.class)
                .type("string", String.class)
                .type("boolean", Boolean.class)
                .edge("EDGE", new SchemaEdgeDefinition.Builder()
                        .source("string")
                        .destination("string")
                        .directed("boolean")
                    .build())
                .build();

     AccumuloProperties properties = new AccumuloProperties();
     properties.setStoreClass(SingleUseMockAccumuloStore.class.getName());

     Graph graph = new Graph.Builder()
                .addSchema(schema)
                .storeProperties(properties)
                .build();

p013570:
Merged into develop.

PR #761, opened 2017-03-14
gh-756 - Added exception in the event that the table name is not specified

t616178:


store-implementation/accumulo-store/src/main/java/uk/gov/gchq/gaffer/accumulostore/utils/TableUtils.java
```diff
@@ -91,6 +92,9 @@ public static synchronized void createTable(final AccumuloStore store)
             LOGGER.info("Table {} exists, not creating", tableName);
             return;
         }
+        if (null == tableName) {
```
p013570: This check should be before: connector.tableOperations().exists(tableName)

PR #774, opened 2017-03-20
Gh 419 op options

p013570:


p013570:
Closing while I tidy up some of the operation interfaces

PR #778, opened 2017-03-22
Gh 467 seed hierarchy

p013570:
Sorry another big pull request. I have basically replaced all uses of ElementSeed with ElementId.


core/operation/src/main/java/uk/gov/gchq/gaffer/operation/impl/get/GetAdjacentIds.java
```diff
@@ -31,20 +31,19 @@
 import java.util.Map;
 
 /**
- * An <code>GetAdjacentEntitySeeds</code> operation will return the
- * {@link uk.gov.gchq.gaffer.operation.data.EntitySeed}s at the opposite end of connected edges to a seed
- * {@link uk.gov.gchq.gaffer.operation.data.EntitySeed}.
+ * An <code>GetAdjacentVertices</code> operation will return the
```
gaffer01: This should say GetAdjacentIds

core/operation/src/main/java/uk/gov/gchq/gaffer/operation/impl/get/GetAdjacentIds.java
```diff
@@ -31,20 +31,19 @@
...
+ * vertices at the opposite end of connected edges to a provided seed vertices
```
gaffer01: "vertices" should be "vertex."

gaffer01:
I spotted a couple of typos so added a note about them. Otherwise, looks good, but worth @t616178 checking as well as it's a big change.

p013570:
Thanks @gaffer01 I'll get them fixed.

p013570:
I've made the requested changes.

PR #798, opened 2017-03-30
Gh 776 map store

gaffer01:


p013570:
There are a couple of merge conflicts - I'll fix them

PR #833, opened 2017-04-18
Gh 826 match hbase with accumulo

t616178:


store-implementation/accumulo-store/src/main/java/uk/gov/gchq/gaffer/accumulostore/utils/Pair.java
```diff
@@ -100,6 +100,6 @@ public boolean equals(final Object obj) {
 
     @Override
     public String toString() {
-        return "Pair{" + "first=" + first + ", second=" + second + '}';
+        return "AbstractPair{" + "first=" + first + ", second=" + second + '}';
```
p013570: Can this Pair class be removed now?

PR #835, opened 2017-04-21
Gh 659 consistent edge creation

t616178:
I have made the Edge class immutable (removed default constructors and setters). I've also done the same to the Entity class to have some consistency between the two Element implementations.

The classes are not *fully* immutable currently. In order to swap the source and destination in the Edge class I have kept the fields non-final. I have also left the "putIdentifier" methods on both classes. This is currently used only when creating ElementTuple objects, but it does allow a way of modifying the fields on an Element. If this is undesirable, then I'd suggest opening two new issues:

* Move vertex swapping in the Edge class to Builder method. This will have the side effect of forcing the source and destination to be specified when invoking the build() method. Currently it's possible to create an empty Edge (i.e new Edge.Builder().build()), which has null for both source and destination.
* Make ElementTuple immutable.

Apologies for the number of changes - this was always going to be a problem when changing the Edge and Entity classes.


core/data/src/main/java/uk/gov/gchq/gaffer/data/element/Edge.java
```diff
@@ -33,52 +40,43 @@
...
     }
 
-    public Object getSource() {
-        return source;
+    private Edge(final Builder builder) {
+        super(builder.group);
+        this.source = builder.source;
+        this.destination = builder.destination;
+        this.directed = builder.directed;
+        builder.properties.forEach((n, v) -> this.putProperty(n, v));
```
t616178: Iterating through properties here since the properties field in Element is private.

core/data/src/main/java/uk/gov/gchq/gaffer/data/element/Edge.java
```diff
@@ -97,19 +95,40 @@ public Object getIdentifier(final IdentifierType identifierType) {
     public void putIdentifier(final IdentifierType identifierType, final Object propertyToBeSet) {
```
t616178: Example of method which breaks true immutability.

core/data/src/main/java/uk/gov/gchq/gaffer/data/element/Edge.java
```diff
@@ -97,19 +95,40 @@ public Object getIdentifier(final IdentifierType identifierType) {
...
+                    }
+                } else if (source.toString()
+                                 .compareTo(destination.toString()) > 0) {
+                    swapVertices();
+                }
+            }
+        }
+    }
+
+    private void swapVertices() {
```
t616178: Could be moved to Builder?

core/data/src/main/java/uk/gov/gchq/gaffer/data/element/Edge.java
```diff
@@ -172,36 +191,55 @@ public String toString() {
                 + "} ";
     }
 
+    @JsonPOJOBuilder(withPrefix = "")
```
t616178: Various Jackson annotations are required here to ensure that the Builder can be used to deserialise objects.

core/operation/src/test/java/uk/gov/gchq/gaffer/data/generator/EntityIdExtractorTest.java
```diff
@@ -47,26 +47,26 @@ public void shouldGetIdentifierFromEntity() {
     public void shouldGetSourceFromEdge() {
         // Given
         final EntityIdExtractor extractor = new EntityIdExtractor(IdentifierType.SOURCE);
-        final Edge edge = new Edge(TestGroups.EDGE, "source", "destination", false);
+        final Edge edge = new Edge(TestGroups.EDGE, "a", "b", false);
```
t616178: Renamed the source and destination fields in a few places throughout the tests, since the new ordering logic would automatically rearrange these to make "destination" the source vertex and "source" the destination vertex. Thought it would be simpler to denote this as an Edge from A to B.

core/data/src/main/java/uk/gov/gchq/gaffer/data/element/Edge.java
gaffer01: Could this be improved if we made the properties field in Element protected? It seems a bit inefficient to have to put each property into the properties map twice (once in the builder and once here).

core/data/src/test/java/uk/gov/gchq/gaffer/data/element/EdgeTest.java
```diff
@@ -328,27 +350,151 @@ public void shouldSerialiseAndDeserialiseIdentifiers() throws SerialisationExcep
...
+                                                .source(sources.get(i))
+                                                .destination(destinations.get(i))
+                                                .build();
+
+            edges.add(edge);
+        }
+
+        // Then
+        assertThat(edges.stream().distinct().count(), greaterThan(1L));
+        assertThat(edges.stream().distinct().count(), greaterThan(1L));
```
gaffer01: These 2 lines are the same. Was the intention that the first line counted the number of unique values of the source and the second line counted the number of destinations?

core/data/src/test/java/uk/gov/gchq/gaffer/data/element/EdgeTest.java
```diff
@@ -328,27 +350,151 @@ public void shouldSerialiseAndDeserialiseIdentifiers() throws SerialisationExcep
...
+                                                .source(sources.get(i))
+                                                .destination(destinations.get(i))
+                                                .build();
+
+            edges.add(edge);
+        }
+
+        // Then
+        assertThat(edges.stream().distinct().count(), equalTo(1L));
+        assertThat(edges.stream().distinct().count(), equalTo(1L));
```
gaffer01: See above comment.

gaffer01:
Looks good - there are a couple of minor comments above.

I'm going to raise an issue about providing an example in the documentation about how this change affects queries which have a filter that specifies the destination, e.g. if undirected edge B-A is inserted into a graph, and B is queried for, with a filter saying that the destination should be A or match A* then no results will be returned, because B is the destination. When executing such queries on undirected edges, the user should specify that either the source or the destination is A.

core/data/src/main/java/uk/gov/gchq/gaffer/data/element/Edge.java
t616178: Yeah - this is the obvious fix. I'll add that in.

core/data/src/test/java/uk/gov/gchq/gaffer/data/element/EdgeTest.java
t616178: Good spot - yes, I'll fix that too.

p013570:
Closing the pull request while the issue with OPERATION_RETURN_MATCHED_SEEDS_AS_EDGE_SOURCE is investigated.

Issue #781, opened 2017-03-23
NullPointerException if the operation declarations file is missing


gaffer01:
If a graph is created when the store properties file contains say "gaffer.store.operation.declarations=/path/to/file" property, but the file does not exist, then a `NullPointerException` is thrown. A more useful exception should be thrown.

p013570:
Merged into develop.

PR #846, opened 2017-04-24
gh-781 - Add check for missing operations declaration path.

t616178:


core/store/src/main/java/uk/gov/gchq/gaffer/store/StoreProperties.java
```diff
@@ -23,6 +23,7 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import uk.gov.gchq.gaffer.commonutil.StreamUtil;
+import uk.gov.gchq.gaffer.core.exception.GafferRuntimeException;
```
p013570: I don't think this import is required.

core/store/src/main/java/uk/gov/gchq/gaffer/store/StoreProperties.java
```diff
@@ -193,6 +194,12 @@ public void setStorePropertiesClass(final Class<? extends StoreProperties> store
     }
 
     public String getOperationDeclarationPaths() {
```
p013570: If possible a simple test for this would be good.

Issue #86, opened 2016-03-02
Enable Graphs to be constructed from URLs


p013570:
It would be useful to be able to create a Graph by providing URLs for the schemas, properties and view.

p013570:
This should be done after gh-100.

p013570:
Merged into develop.

PR #853, opened 2017-04-26
gh-86 Enable Graphs to be constructed from URLs.

GCHQDeveloper404:


core/common-util/src/main/java/uk/gov/gchq/gaffer/commonutil/StreamUtil.java
```diff
@@ -165,6 +168,43 @@ public static InputStream opAuths(final Class clazz, final boolean logErrors) {
         return schemas;
     }
 
+
+    public static InputStream[] openStream(final String... urls) throws IOException {
+        final InputStream[] rtn = new InputStream[urls.length];
+        int count = 0;
+        for (final String url : urls) {
+            rtn[count++] = openStream(url);
```
p013570: If this throws an exception then you need to ensure the rtn InputStreams are closed.

core/common-util/src/main/java/uk/gov/gchq/gaffer/commonutil/StreamUtil.java
```diff
@@ -165,6 +168,43 @@ public static InputStream opAuths(final Class clazz, final boolean logErrors) {
         return schemas;
     }
 
+
+    public static InputStream[] openStream(final String... urls) throws IOException {
```
p013570: Perhaps remove these methods and just have the methods that take a URL. This method could be confused with the open stream from path.

core/graph/src/main/java/uk/gov/gchq/gaffer/graph/Graph.java
```diff
@@ -358,6 +380,26 @@ public Builder addSchema(final InputStream schemaStream) {
             }
         }
 
+        public Builder addSchema(final URL schemaURL) {
```
p013570: I don't think we need these builder methods. Users can just use the StreamUtil class to convert a URL into an InputStream and then use the addSchema(final InputStream schema) method.

GCHQDeveloper404: Doesn't this just save the user a step, make for easier cleaner code?

core/common-util/src/test/java/uk/gov/gchq/gaffer/commonutil/StreamUtilTest.java
```diff
@@ -0,0 +1,31 @@
...
+        //Given
+        final URL resource = getClass().getClassLoader().getResource(FILE_NAME);
+        if (resource == null) fail("Test json file not found:" + FILE_NAME);
+
+        //When
+        final InputStream[] inputStreams = StreamUtil.openStreams(resource);
+
+        //Then
+        assertNotNull(inputStreams);
+        assertFalse("InputStreams length is 0", inputStreams.length == 0);
```
p013570: Just need to close the input streams in a finally block.

rest-api/core-rest/src/main/java/uk/gov/gchq/gaffer/rest/service/IOperationService.java
```diff
@@ -45,10 +45,9 @@
 @Consumes(MediaType.APPLICATION_JSON)
 @Produces(MediaType.APPLICATION_JSON)
 public interface IOperationService {
-
     @POST
     @ApiOperation(value = "Performs the given operation chain on the graph", response = Object.class)
-    Object execute(final OperationChain opChain);
+    Object execute(final OperationChain opChain); //takes in operation TODO
```
p013570: Remove todo.

core/common-util/src/main/java/uk/gov/gchq/gaffer/commonutil/StreamUtil.java
```diff
@@ -128,7 +128,43 @@ public static InputStream authScores(final Class clazz) {
         return schemas;
     }
 
+    public static InputStream[] openStreams(final boolean isLogging, final URL... urls) throws IOException {
```
p013570: Don't think we need the isLogging parameter.

Issue #763, opened 2017-03-15
Create and integrate common cache service


d47853:
Currently there are a number of cache-dependent components all around Gaffer, such as the Named Operations and Job Trackers. There is a growing need for a common cache service that would remove repeated caching logic from these components and provide an interface for new cache-reliant features. 

Users may want to use a different cache implementation depending on their requirements (i.e. JCS, Hazelcast, Ehcache etc). Therefore the gaffer cache service hides the implementation behind an interface and loads the exact service at runtime using a system variable. This provides a clean interface for all components to use, whilst providing options in terms of specific implementation.

To start with the cache service will provide 3 implementations: JCS, Hazelcast and a basic HashMap. This can be expanded if required.

gaffer01:
Can you add a description of what this common cache service would do?

p013570:
The above comment has been updated with a description.

p013570:
# Migration steps
There is now a single shared cache for Gaffer stores. This replaces the separate caches that were previously used by Named Operations and the Job Tracker. See the Cache [README](https://github.com/gchq/Gaffer/blob/master/core/cache/README.md).

If you were previously using the Named Operations and the Job Tracker then there will be a few changes you will need to make:

Replace:

```xml
<dependency>
    <groupId>uk.gov.gchq.gaffer</groupId>
    <artifactId>job-tracker-jcs</artifactId>
    <version>gaffer.version</version>
</dependency>

<dependency>
  <groupId>uk.gov.gchq.gaffer</groupId>
  <artifactId>named-operations-library</artifactId>
  <version>${gaffer.version}</version>
</dependency>
```

with

```xml
<dependency>
    <groupId>uk.gov.gchq.gaffer</groupId>
    <artifactId>jcs-cache-service</artifactId>
    <version>gaffer.version</version>
</dependency>

 <dependency>
  <groupId>uk.gov.gchq.gaffer</groupId>
  <artifactId>named-operation-library</artifactId>
  <version>${gaffer.version}</version>
</dependency>
```


<br/>

And replace the following properties in you store.properties file:

```properties
gaffer.store.job.tracker.class=uk.gov.gchq.gaffer.jobtracker.JcsJobTracker
gaffer.store.job.tracker.config.path=/path/to/config/file
```

with the new properties:

```properties
gaffer.store.job.tracker.enabled=true

gaffer.cache.service.class=uk.gov.gchq.gaffer.cache.impl.JcsCacheService

# Optionally provide custom cache properties
[...]

PR #854, opened 2017-04-26
Gh 763 cache

d47853:


library/cache-library/cache-service/pom.xml
```diff
@@ -0,0 +1,37 @@
...
+    <parent>
+        <artifactId>cache-library</artifactId>
+        <groupId>uk.gov.gchq.gaffer</groupId>
+        <version>0.6.7-SNAPSHOT</version>
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
+
+    <artifactId>cache-service</artifactId>
+
+    <dependencies>
```
p013570: The cache service should not depend on java.servlet. It should work without the REST API. I think the CacheServiceLoader should be triggered by the initialise method in the Store class, not a web listener.

library/cache-library/cache-service/src/main/java/uk/gov/gchq/gaffer/cache/CacheServiceLoader.java
```diff
@@ -0,0 +1,95 @@
...
+import javax.servlet.ServletContextEvent;
+import javax.servlet.ServletContextListener;
+import javax.servlet.annotation.WebListener;
+
+/**
+ * Bootstrapped when the REST service starts. Looks at a system property to determine the cache service to load.
+ * Then initialises it, after which any component may use {@code CacheServiceLoader.getService()} to get the service
+ * that can retrieve the appropriate cache.
+ */
+@WebListener(value = "Gaffer cache service loader")
```
p013570: See other comment about removing this web listener and having it triggered by initialising a Gaffer Store.

library/cache-library/cache-service/src/main/java/uk/gov/gchq/gaffer/cache/CacheServiceLoader.java
```diff
@@ -0,0 +1,95 @@
...
+    private static ICacheService service;
+
+    /**
+     * Looks at a system property and initialises an appropriate cache service. If no cache service is specified in the
+     * system property, the loader falls back onto a default which is backed by HashMaps.
+     *
+     * @throws IllegalArgumentException if an invalid cache class is specified in the system property
+     */
+    public static void initialise() {
+        String cacheClass = System.getProperty(CacheSystemProperty.CACHE_SERVICE_CLASS);
```
p013570: This should come from a StoreProperty not system property so it works without a REST API.

p013570:
I don't think we would ever want any other JobTracker implementations apart from the GafferCacheJobTracker. If you wish to use the JobTracker with a different cache you would do that by writing a CacheService implementation instead. So we should make the JobTracker a concert class and put the GafferCacheJobTracker logic in there. Similarly with NamedOperations. You will need to do a bit of shuffling around of the code to sort of the dependency hierarchy of modules as currently JobTracker is in the Operation module.

StoreProperties will need to be updated:
gaffer.store.job.tracker.class can be removed
gaffer.store.job.tracker.config.path should be changed to be the general cache config path
We may need to add a boolean property to enable/disable the job tracker.

p013570:
Closing while the changes are implemented.

p013570:
Looks good. Please could you add some migration steps to the issue gh-763.

PR #857, opened 2017-04-27
Gh 781 null pointer missing operation declarations file

GCHQDeveloper404:


core/common-util/src/main/java/uk/gov/gchq/gaffer/commonutil/StreamUtil.java
```diff
@@ -158,37 +118,46 @@ public static InputStream opAuths(final Class clazz, final boolean logErrors) {
         int index = 0;
         final InputStream[] schemas = new InputStream[schemaFiles.size()];
         for (final String schemaFile : schemaFiles) {
-            schemas[index] = openStream(clazz, schemaFile, logErrors);
-            index++;
+            try {
+                schemas[index] = openStream(clazz, schemaFile);
+                index++;
+            } catch (IllegalArgumentException e) {
```
p013570: I think you would be safer catching all throwable objects here

core/common-util/src/main/java/uk/gov/gchq/gaffer/commonutil/StreamUtil.java
```diff
@@ -158,37 +118,46 @@ public static InputStream opAuths(final Class clazz, final boolean logErrors) {
...
     }
 
-    public static InputStream openStream(final Class clazz, final String path) {
-        return openStream(clazz, path, false);
+    private static int closeAllInputStreams(final InputStream[] inputStreams) {
+        int closedStreamsCount = 0;
+        for (final InputStream stream : inputStreams) {
+            try {
+                stream.close();
+            } catch (IOException e) {
```
p013570: I think you would be safer catching all throwable objects here.

PR #865, opened 2017-05-03
Gh 856 rest api execute single operation

GCHQDeveloper404:


rest-api/core-rest/src/main/java/uk/gov/gchq/gaffer/rest/service/IOperationService.java
```diff
@@ -51,6 +52,11 @@
     Object execute(final OperationChain opChain);
 
     @POST
+    @Path("/operation")
```
t616178: Might it be better to change around the paths for these operations now we're adding the "execute single operation" option?

Something like:

Execute operation chain: /graph/doOperation/chain
Execute single operation: /graph/doOperation

?

rest-api/core-rest/src/main/java/uk/gov/gchq/gaffer/rest/service/IOperationService.java
```diff
@@ -51,6 +52,11 @@
...
+    @ApiOperation(value = "Performs the given operation on the graph", response = Object.class)
+    Object execute(final Operation operation);
+
+    @POST
     @Path("/chunked")
```
t616178: For consistency, do we want to add a chunked implementation of the "execute single operation" endpoint?

t616178:
Could some integration tests be added? This would probably involve adding some utility methods to the RestApiTestUtil class and then adding some new tests to the OperationTestIT class.

rest-api/core-rest/src/main/java/uk/gov/gchq/gaffer/rest/service/IOperationService.java
p013570: I know there are a lot of changes in version 0.7.0 already, but that could be quite an annoying breaking change. It makes this tidier but doesn't really give any benefit. Perhaps something we think about doing at a later date.

Issue #864, opened 2017-05-03
Refactor existing toString(), equals() and hashCode() methods to consistently use the Apache Commons Lang 3 style


t616178:
All classes which implement toString(), equals() and hashCode() in the Gaffer codebase should do in a consistent way. We have chosen the Apache Commons Lang style for use throughout the project.

Classes which implement these methods should contain implementations similar to that described below:

```java
    @Override
    public boolean equals(final Object obj) {
        if (this == obj) {
            return true;
        }

        if (obj == null || getClass() != obj.getClass()) {
            return false;
        }

        final MyObject myObj = (MyObject) obj;

        return new EqualsBuilder()
                .append(field1, myObj.field1)
                .append(field2, myObj.field2)
                .isEquals();
    }

    @Override
    public int hashCode() {
        return new HashCodeBuilder(17, 37)
                .append(field1)
                .append(field2)
                .toHashCode();
    }

    @Override
    public String toString() {
        return new ToStringBuilder(this)
                .append("field1", field1)
                .append("field2", field2)
                .toString();
    }
```

GCHQDeveloper404:
toString portion of this ticket has been done.

goti921:
@t616178 
Is it up for grabs ?

p013570:
Hi @goti921,

Yeah go for it, thanks! 

@GCHQDeveloper404 has completed the toString part of the issue, I'm not sure how much is actually left to do. Please branch off the ```develop``` branch and refer to our [ways of working](https://github.com/gchq/Gaffer/wiki/Ways-of-Working#pull-requests) wiki page.

Let me know if you have any questions.

m55624:
I am going to start work on this.  @goti921 please let me know if you have done anything on this and where and I will stop.

goti921:
Hi @m55624 
I have almost completed refactoring of the equals(0 method. 
You can carry on with the hashcode() method.

m55624:
@goti921 
I will leave this for now.  Thanks for doing this!

p013570:
Merged into develop-1.0.0.

PR #884, opened 2017-05-11
gh-864 refactoring toString() to use ToStringBuilder.

GCHQDeveloper404:
This is only the toString portion of the ticket.


core/operation/src/main/java/uk/gov/gchq/gaffer/operation/OperationChain.java
```diff
@@ -97,23 +98,9 @@ void setOperationArray(final Operation[] operations) {
 
     @Override
     public String toString() {
-        final StringBuilder strBuilder = new StringBuilder("OperationChain[");
```
t616178: We might need to check to see if some of these special implementations should be left as they are - there's a debate to be had around making the implementation uniform throughout the project vs keeping some flexibility to handle special cases like this. @p013570 might have an opinion on this.

core/exception/src/main/java/uk/gov/gchq/gaffer/core/exception/Status.java
```diff
@@ -89,10 +90,10 @@
 
     public static Status fromStatusCode(final int statusCode) {
         return Arrays.asList(values())
-                     .stream()
```
t616178: I've noted on gh-890 that we need to make sure we're all using the same formatting settings to avoid changes like this creeping into pull requests too often. We need to agree and publish this somewhere and make sure we're all following it.

core/store/src/main/java/uk/gov/gchq/gaffer/store/schema/Schema.java
```diff
@@ -201,7 +202,9 @@ public String getTimestampProperty() {
     @Override
     public String toString() {
         try {
-            return "Schema" + new String(toJson(true), CommonConstants.UTF_8);
+            return new ToStringBuilder(this)
+                    .append(new String(toJson(true), CommonConstants.UTF_8))
```
t616178: Should this line read:

```
.append("schema", new String(toJson(true), CommonConstants.UTF_8))
```

?

GCHQDeveloper404: No, Why would we need "schema" append within an object call schema?

**current**
`uk.gov.gchq.gaffer.store.schema.Schema@6ce139a4[{
  "edges" : {
  ...
}]`

vs

**suggestion**
`uk.gov.gchq.gaffer.store.schema.Schema@6ce139a4[schema={
  "edges" : {
  ...
}]`

core/operation/src/main/java/uk/gov/gchq/gaffer/operation/OperationChain.java
GCHQDeveloper404: I've checked with @p013570 and we decided to keep them uniform. The output of this is expressive enough.

`uk.gov.gchq.gaffer.operation.OperationChain@598067a5[
operations=[
uk.gov.gchq.gaffer.operation.impl.add.AddElements@3c0ecd4b, uk.gov.gchq.gaffer.operation.impl.get.GetAllElements@14bf9759
]]`

Issue #889, opened 2017-05-11
Ensure any operations that have a closable input are closed if an exception is thrown


p013570:
In the Graph class we should wrap the execution of an OperationChain in a try-catch and if an exception is caught we should close all operation inputs.

t616178:
Is there any benefit to using try with resources blocks to automatically close the inputs if an exception is thrown?

This may negate the need for any additional try blocks inside the finally blocks?

p013570:
We can't use the try with resources for a lot of the cases as we only want to close the input if there is an exception thrown. I think the try with resources would always close the input at the end of the try statement. We need the input to stay open until the output of the operation chain has been consumed by the user, either via REST or java.

t616178:
Merged into develop.

PR #890, opened 2017-05-11
gh-889 - made operations closeable

p013570:


core/operation/src/main/java/uk/gov/gchq/gaffer/operation/Operation.java
```diff
@@ -28,7 +30,12 @@
  * Operations must be JSON serialisable in order to make REST API calls.
  */
 @JsonTypeInfo(use = JsonTypeInfo.Id.CLASS, include = JsonTypeInfo.As.PROPERTY, property = "class")
-public interface Operation {
+public interface Operation extends Closeable {
+
+    default void close() throws IOException {
```
t616178: If there's no implementation here, maybe this is the wrong level at which to introduce this interface?

core/graph/src/main/java/uk/gov/gchq/gaffer/graph/Graph.java
```diff
@@ -133,19 +134,25 @@ public void execute(final Operation operation, final User user) throws Operation
...
-            result = graphHook.postExecute(result, operationChain, user);
-        }
+            for (final GraphHook graphHook : graphHooks) {
+                result = graphHook.postExecute(result, operationChain, user);
+            }
 
-        return result;
+            return result;
+
+        } catch (final Exception e) {
```
t616178: Can this only be an OperationException, or do we want to catch RuntimeExceptions here too?

core/graph/src/main/java/uk/gov/gchq/gaffer/graph/Graph.java
```diff
@@ -160,16 +167,26 @@ public JobDetail executeJob(final OperationChain<?> operationChain, final User u
...
+                graphHook.preExecute(operationChain, user);
+            }
 
-        O result = store.execute(operationChain, user);
+            result = store.execute(operationChain, user);
+
+            for (final GraphHook graphHook : graphHooks) {
+                result = graphHook.postExecute(result, operationChain, user);
+            }
+        } catch (final Exception e) {
```
t616178: Same comment regarding OperationExceptions.

core/operation/src/test/java/uk/gov/gchq/gaffer/operation/OperationChainTest.java
```diff
@@ -66,11 +71,11 @@ public void shouldSerialiseAndDeserialiseOperationChain() throws SerialisationEx
         assertNotNull(deserialisedOp);
         assertEquals(2, deserialisedOp.getOperations().size());
         assertEquals(OperationImpl.class, deserialisedOp.getOperations()
-                                                        .get(0)
```
t616178: Do we need to agree a standard for formatting? Or at least make sure that everything is using the agreed specification? I've been seeing quite a few places in PRs recently where there are formatting differences.

core/graph/src/main/java/uk/gov/gchq/gaffer/graph/Graph.java
p013570: We need to ensure the operation inputs are always closed when any exception is thrown, otherwise we risk leaving connections to accumulo/hbase open.

core/operation/src/main/java/uk/gov/gchq/gaffer/operation/Operation.java
p013570: All operation implementations must be able to close all closable fields it has. But just to simplify things for simple operations I added the default method here to avoid having to duplicate the empty implementation.

p013570: I'll add a comment

javadev001001: Are all standard gaffer operations going to implement this method?  Or is this something that's going to be done by exception rather than default?

p013570: I'm not sure what the best solution is. As it is all operations are closeable and users can chose to override the close method.

It also does makes sense for an operation implementation to chose whether to implement closable or not and add "implements Closeable". 

I think making all operations implement closeable will encourage more users to think about adding a custom implementation of close.

It also means in OperationChain.close I can just iterate round all operations and call close. Alternatively I would need to do instanceof and then cast to Closeable for each operation.

core/operation/src/test/java/uk/gov/gchq/gaffer/operation/OperationChainTest.java
p013570: Yeah that would be useful. Not sure the best way to share that standard though. We can add notes to the ways of working wiki page. But better would be to share settings files for intellij/eclipse etc.

PR #900, opened 2017-05-16
Gh 821 849 new operations

t616178:


p013570:
Sorry @t616178 , I've finished tinkering with your branch now. Please could check you have happy with the changes I made.

core/data/src/main/java/uk/gov/gchq/gaffer/data/element/comparison/ElementPropertyComparator.java
```diff
@@ -0,0 +1,179 @@
...
+        }
+        if (val2 == null) {
+            return -1;
+        }
+
+        if (reversed) {
+            return ((Comparable) val2).compareTo(val1);
+        }
+
+        return ((Comparable) val1).compareTo(val2);
```
t616178: Should we delegate to the supplier comparator object here?

PR #909, opened 2017-05-18
Gh 872 documentation

p013570:
Added READMEs for the core modules and the proxy store. The rest of the javadoc can be done separately.


p013570:
Thanks @t616178 for correcting the typos - looks good.

PR #911, opened 2017-05-19
Gh 752 edge seed direction

p013570:


p013570:
To implement the change I basically just changed the directed boolean on EdgeId to be a nullable Boolean. So it is now optional to specify the directed field on an EdgeSeed when doing a query.

p013570:
Closing whilst we investigate switching to use an Enum to represent the directed flag - 'DIRECTED', 'UNDIRECTED' or 'EITHER'.

PR #926, opened 2017-05-23
Gh 752 edge seed direction enum

p013570:


p013570:
@t616178 happy with the changes?

Issue #190, opened 2016-05-10
Refactor Accumulo Store code and package structure


a09631:
Refactor some of the Accumulo Store code, and update the package structure to be more descriptive.

p013570:
@a09631 as this is nearly a year old it would be good to progress this issue. We don't really want to start refactoring too much of the accumulo store as we will end up breaking some of the complex key package logic. Perhaps we could break this issue into smaller issues that target specific bits of the code.

GCHQDeveloper404:
Removing myself from the assignee, this is an ongoing ticket which I have done some work with. 
To echo @p013570 " Perhaps we could break this issue into smaller issues that target specific bits of the code."

t616178:
Closing for now as the Accumulo store has been tidied up enough. If there are any further issues with the Accumulo store these should be raised under a new ticket.

PR #929, opened 2017-05-24
gh-190 Refactor AbstractCoreKeyAccumuloElementConverter.java

GCHQDeveloper404:
Please use your finest toothed comb please.


GCHQDeveloper404:
There will be more, I'm just doing this in byte sized pieces.

store-implementation/accumulo-store/src/main/java/uk/gov/gchq/gaffer/accumulostore/key/core/AbstractCoreKeyAccumuloElementConverter.java
```diff
@@ -291,41 +270,34 @@ public Properties getPropertiesFromColumnVisibility(final String group, final by
...
-                    } else {
-                        final byte[] bytes = serialiser.serialiseNull();
-                        writeBytes(bytes, out);
-                    }
-                } else {
-                    writeBytes(AccumuloStoreConstants.EMPTY_BYTES, out);
+            byte[] bytes = AccumuloStoreConstants.EMPTY_BYTES;
+            if (serialiser != null) {
+                Object value = properties.get(propertyName);
+                if (value != null) {
```
p013570: If value is null then we should use serialiser.serialiseNull()

GCHQDeveloper404: I shall implement a test for a serialiser that overwrites the default.

GCHQDeveloper404:
Testing added for serialiser that overwrites the default serialiseNull()

PR #932, opened 2017-05-25
Gh 880 split points

p013570:


gaffer01:
Is there any reason to use `Base64` from `net.iharder.base64` rather than `org.apache.commons.codec.binary.Base64` which is what is used in `IngestUtils`?

p013570:
@gaffer01 don't think so, it would make more sense to use the same. I'll make that change now, thanks.

p013570:
@gaffer01 happy with the changes?

gaffer01:
@p013570 Yes, looks good, but CI build is failing.

PR #934, opened 2017-05-31
Gh 659 consistent edge creation

t616178:
Enforce consistent creation of undirected Edge objects, determined by comparing the source and destination objects.


core/data/src/main/java/uk/gov/gchq/gaffer/data/element/Edge.java
```diff
@@ -183,36 +220,61 @@ public String toString() {
...
         public Builder source(final Object source) {
-            edge.setSource(source);
+            this.source = source;
             return this;
         }
 
-        public Builder dest(final Object dest) {
-            edge.setDestination(dest);
+        @JsonTypeInfo(use = JsonTypeInfo.Id.CLASS, include = JsonTypeInfo.As.PROPERTY, property = "class")
+        public Builder destination(final Object destination) {
```
p013570: To reduce the migration for users, please can we keep this as 'dest'. You should be able to add a JsonProperty annotation on to tell jackson it is called 'destination'

example/example-graph/src/main/java/uk/gov/gchq/gaffer/example/gettingstarted/generator/DataGenerator2.java
```diff
@@ -0,0 +1,34 @@
+/*
```
p013570: Everything in example/gettingstarted has been moved to the doc module.

core/data/src/main/java/uk/gov/gchq/gaffer/data/element/Edge.java
t616178: This field is referred to as "destination" everywhere else in the code base. There will some migration required, so is it worth changing this so that this is consistent everywhere?

p013570: OK, we can have that discussion in a separate issue and do that migration separately if needed.

p013570:
Closing until the conflicts have been addressed.

PR #947, opened 2017-06-06
Gh 190 refactor accumulo store

GCHQDeveloper404:


GCHQDeveloper404:
The hotfix for for a bug causes test failures so I will apply that separately.

GCHQDeveloper404:
Merging with other work.

PR #950, opened 2017-06-07
Gh 839 create timestamp set property

gaffer01:


library/time-library/src/test/java/uk/gov/gchq/gaffer/time/serialisation/RBMBackedTimestampSetSerialiserTest.java
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    /**
+     * Used to calculate the sizes of the serialised {@link RBMBackedTimestampSet} given in the Javadoc for
+     * {@link RBMBackedTimestampSetSerialiser}.
+     *
+     * @param args No arguments are needed.
+     * @throws SerialisationException Won't happen
+     */
+    public static void main(final String[] args) throws SerialisationException {
```
p013570: Could this be converted into a test with assertions?

gaffer01: Yes, that would be another way of doing it - I didn't make it at a test because it's not core functionality, but it could be done as a test.

gaffer01:
I'll close this for now while I make a couple of changes:

1. Make the `main` method that generates the serialised sizes a unit test.
2. Add an example for the wiki.
3. Rename `RBMBackedTimestampSetAggregator` to `RBMBackedTimestampSetUnionAggregator`.
4. Add an aggregator that merges two `RBMBackedTimestampSet`s together, but only keeps the most recent N timestamps.

PR #953, opened 2017-06-09
Gh 874 accumulo and hbase serialisers

m55624:


core/store/src/main/java/uk/gov/gchq/gaffer/store/Store.java
```diff
@@ -158,6 +159,18 @@ public void initialise(final Schema schema, final StoreProperties properties) th
         addExecutorService();
     }
 
+    public void initialise(final Schema schema, final StoreProperties properties, final SchemaOptimiser schemaOptimiser) throws StoreException {
```
p013570: Instead of having to add this extra method and the extra optimiseSchema method, it would be neater to just override the createSchemaOptimiser() method in the AccumuloStore class:

```java
protected SchemaOptimiser createSchemaOptimiser() {
          return new SchemaOptimiser(new AccumuloSerialisationFactory());
}
```

core/serialisation/src/main/java/uk/gov/gchq/gaffer/serialisation/implementation/SerialisationFactory.java
```diff
@@ -23,13 +24,17 @@
...
 /**
  * A <code>SerialisationFactory</code> holds a list of core serialisers and
  * is design to provide compatible serialisers for given object classes.
  */
 public class SerialisationFactory {
-    private static final Serialiser[] SERIALISERS = new Serialiser[]{
+
+    private final List<Serialiser> serialisers;
+    private static final Serialiser LAST_RESORT_FINALISER = new JavaSerialiser();
+    private static final Serialiser[] CORE_SERIALISERS = new Serialiser[]{
```
p013570: If you made this a List you wouldn't need your serialiserAlreadyinList method. You could just do CORE_SERIALISERS.contains(serialiser)

p013570: Argh - ignore that comment as the serialiser implementations don't have hashcode/equals methods so it wouldn't work.

PR #963, opened 2017-06-12
Gh 945 improve cache service loader

d47853:


d47853:
undergoing further work

PR #968, opened 2017-06-13
Gh 515 aggregator schema validation

m55624:


core/store/src/main/java/uk/gov/gchq/gaffer/store/schema/SchemaElementDefinitionValidator.java
```diff
@@ -137,6 +137,9 @@ private ValidationResult validateAggregator(final ElementAggregator aggregator,
         }
 
         if (!elementDef.isAggregate()) {
+            if (!elementDef.getGroupBy().isEmpty()) {
```
p013570: Please can you add a unit test to SchemaElementDefinitionValidatorTest to test this logic.

PR #969, opened 2017-06-13
Gh 822 add group by function

d47853:


core/data/src/main/java/uk/gov/gchq/gaffer/data/elementdefinition/view/View.java
```diff
@@ -160,6 +162,22 @@ public void expandGlobalDefinitions() {
...
     }
 
+    @JsonIgnore
+    public Function<Element, Set<Object>> getGroupByFunction() {
+        return element -> {
+            Set<Object> key = new HashSet<>();
+            String group = element.getGroup();
+            Set<String> groupBy = getElementGroupBy(group);
+            if (groupBy == null) {
+                return key;
```
p013570: You could optimise this by not creating the HashSet unless you actually need it.

In the case the groupBy is null you could return Collections.emptySet();

When you create the HashSet you can create it using the size of the groupBy set.

core/data/src/main/java/uk/gov/gchq/gaffer/data/elementdefinition/view/View.java
```diff
@@ -160,6 +162,22 @@ public void expandGlobalDefinitions() {
         }
     }
 
+    @JsonIgnore
+    public Function<Element, Set<Object>> getGroupByFunction() {
```
p013570: Perhaps change the name of the method to createGroupByFunction so user's are more aware that it is doing some work and creating a new instance of the function each time.

p013570: The key should also include the Element identifiers (SOURCE, DESTINATION, DIRECTED, VERTEX) and the element group. These are all the things that make up the element key.

p013570:
In the case then aggregation is disabled the elements should not be grouped together and aggregated.

So, what should we do in the case that aggregation is disabled for the element?
e.g
!schema.getElement(element.getGroup()).isAggregate()

In the view we don't know anything about the schema, so can we rely on the method calling this getGroupByFunction() to check?

d47853:
Only adding this function to the Schema for the time being.

core/data/src/test/java/uk/gov/gchq/gaffer/data/elementdefinition/view/ViewTest.java
```diff
@@ -211,154 +212,118 @@ public void shouldJsonSerialiseAndDeserialise() {
...
-                "      }, {%n" +
-                "        \"predicate\" : {%n" +
-                "          \"class\" : \"uk.gov.gchq.gaffer.function.ExampleFilterFunction\"%n" +
-                "        },%n" +
-                "        \"selection\" : [ \"dateProperty\" ]%n" +
-                "      } ]%n" +
-                "    }%n" +
-                "  }%n" +
-                "}"), new String(view.toJson(true)));
+            "  \"edges\" : {%n" +
```
p013570: Please can you tell your IDE to not change this indenting? Also occurred in the SchemaTest file.

core/store/src/main/java/uk/gov/gchq/gaffer/store/schema/Schema.java
```diff
@@ -123,6 +128,49 @@ public boolean isAggregationEnabled() {
...
+    /**
+     * Creates a Function that takes and element as input and outputs a set of objects including the
+     * Group-by values, the Identifiers and the group. These act as a key and can be used in a
+     * Collector. This function is dependent on the hashcode of the key components so make sure the
+     * Identifiers and Group-By objects all have a valid hashcode method.
+     *
+     * @return a Set which makes up a unique Identifier. The set is composed of the input Element's
+     * Group-By, Identifiers and Group.
+     */
+    public Function<Element, Set<Object>> createGroupByFunction() {
```
p013570: Oh, I've just been thinking about this a bit more and we can't just use a Set as the key as we lose the context as to what things are in the key. For example:
if Entity 1 = {
group: "Red"
vertex: "1"
properties: {}
}

Entity2 = {
group: "1"
vertex: "Red"
properties: {}
}

These entities are different but when Red and 1 are put in the Set they would have the same key.

Instead of using a Set perhaps we should just use the Element class.

For example, something like:

```java
public Function<Element, Element> toElementKey(final Schema schema) {
        return element -> {
            final Element key = element.emptyClone();
            final SchemaElementDefinition elDef = schema.getElement(key.getGroup());
            if (elDef == null) {
                throw new RuntimeException("received element " + element + " which belongs to a " +
                        "group not found in the schema");
            }

            final Set<String> groupBy = elDef.getGroupBy();
            for (final String property : groupBy) {
                key.putProperty(property, element.getProperty(property));
            }
            return key;
        };
    }
```

p013570:
@d47853 thanks for making the changes. 
Before merging it in, I'm going to add some extra methods to the branch.

p013570:
I've added the extra methods/classes. We just need to test them.

PR #986, opened 2017-06-16
Gh 839 create timestamp set property

gaffer01:


library/time-library/src/main/java/uk/gov/gchq/gaffer/time/BoundedTimestampSet.java
```diff
@@ -0,0 +1,213 @@
...
+/**
+ * A <code>BoundedTimestampSet</code> is an implementation of {@link TimestampSet} that can contain a maximum number
+ * N of timestamps. If more than N timestamps are added then a uniform random sample of size approximately N of the
+ * timestamps is retained.
+ *
+ * <p>This is useful in avoiding the in-memory or serialised size of the set of timestamps becoming too
+ * large. If less than N timestamps are added then the timestamps are stored in a {@link RBMBackedTimestampSet}. If
+ * more than N timestamps are added then a uniform random sample of size N of the timestamps is retained.
+ */
+public class BoundedTimestampSet implements TimestampSet {
```
p013570: This class is not json serialisable. To enable it to be used over the rest api, it needs to be json serialisable, i.e add a default constructor, getters and setters.

Perhaps add a simple test like this:

```java
    @Test
    public void shouldSerialiseAndDeserialiseOperation() throws SerialisationException {
        // Given
        final JSONSerialiser serialiser = new JSONSerialiser();
        final BoundedTimestampSet boundedTimestampSet = new BoundedTimestampSet(CommonTimeUtil.TimeBucket.SECOND, 10);
        IntStream.range(0, 20)
                .forEach(i -> {
                    boundedTimestampSet.add(Instant.ofEpochMilli(i * 1000L));
                    if (i <= 9) {
                        assertEquals(BoundedTimestampSet.State.NOT_FULL, boundedTimestampSet.getState());
                    } else {
                        assertEquals(BoundedTimestampSet.State.SAMPLE, boundedTimestampSet.getState());
                    }
                });


        // When
        byte[] json = serialiser.serialise(boundedTimestampSet, true);
        final BoundedTimestampSet deserialisedObj = serialiser.deserialise(json, BoundedTimestampSet.class);

        // Then
        assertEquals(boundedTimestampSet, deserialisedObj);
    }
```

library/time-library/src/main/java/uk/gov/gchq/gaffer/time/RBMBackedTimestampSet.java
```diff
@@ -0,0 +1,211 @@
...
+ *
+ * <p>Internally this class stores the timestamps in a {@link RoaringBitmap}.
+ *
+ * <p>NB: This class does not accept {@link Instant}s that are before the Unix epoch or after the {@link Instant}
+ * which is <code>Integer.MAX_VALUE * 1000L</code> milliseconds after the epoch (approximately 3:14 on January 19th
+ * 2038). This is due to {@link RoaringBitmap} only accepting integers. As the smallest {@link TimeBucket} is
+ * a second then the maximum supported {@link Instant} is the maximum integer multiplied by 1000L milliseconds after
+ * the epoch.
+ */
+public class RBMBackedTimestampSet implements TimestampSet {
```
p013570: This class is also not json serialisable. See comment on BoundedTimestampSet.

p013570:
I've updated your branch to make the sets json serialisable. Please could you check you are happy with my changes. It looks like I have changed a lot, I've just done:
renamed get() to getTimestamps()
added builders for the 2 timestamp sets to allow the json deserialisation to work
added a new dependency on jackson-datatype-jsr310 so jackson can serialisation/deseriable Instant objects
added/updated the toString, equals, hashcode methods

gaffer01:
Looks fine.

The only question is whether `BoundedTimestampSet` should have an `equals()` method at all - if it's in the state where it's sampling then it's using a `ReservoirLongsUnion` which maintains a random sample. The `equals()` method uses the random sample. So two `BoundedTimestampSet`s which have had exactly the same data added might not be equal (as the random sample might be different). We could either accept this, and decide `equals()` means that they happened to retain exactly the same sample, or not implement `equals()` at all.

p013570:
I think it is better than using the default equals on Object. I've added a javadoc to the equals and hash code methods.

Issue #991, opened 2017-06-19
Create an operation to calculate the Operation Chain Score for a provided operation chain


p013570:
Prior to executing an operation chain, users could use this operation to calculate the score of their operation chain.

p013570:
Merged into develop.

PR #993, opened 2017-06-19
gh-991 - added ScoreOperationChain operation

p013570:
- Created a new operation - ScoreOperationChain
- Moved the score logic from OperationChainLimiter to ScoreOperationChainHandler to avoid duplication.

### DIFF
--- a/core/store/src/main/java/uk/gov/gchq/gaffer/store/operation/handler/ScoreOperationChainHandler.java
+++ b/core/store/src/main/java/uk/gov/gchq/gaffer/store/operation/handler/ScoreOperationChainHandler.java
@@ -171,13 +171,12 @@ public class ScoreOperationChainHandler implements OutputOperationHandler<ScoreO
         try {
             loadMapsFromProperties(readProperties(operationScores), readProperties(authScores));
         } finally {
-            CloseableUtil.close(operationScores);
-            CloseableUtil.close(authScores);
+            CloseableUtil.close(operationScores, authScores);
         }
     }
 
     private void loadMapsFromProperties(final Properties operationScoreProperties, final Properties operationAuthScoreLimitProperties) {
-        Map<Class<? extends Operation>, Integer> opScores = new LinkedHashMap<>();
+        final Map<Class<? extends Operation>, Integer> opScores = new LinkedHashMap<>();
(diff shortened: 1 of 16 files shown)


core/store/src/main/java/uk/gov/gchq/gaffer/store/operation/handler/ScoreOperationChainHandler.java
```diff
@@ -0,0 +1,265 @@
...
+
+    private void loadMapsFromProperties(final Path operationScores, final Path authScores) {
+        loadMapsFromProperties(readProperties(operationScores), readProperties(authScores));
+    }
+
+    private void loadMapsFromProperties(final InputStream operationScores, final InputStream authScores) {
+        try {
+            loadMapsFromProperties(readProperties(operationScores), readProperties(authScores));
+        } finally {
+            CloseableUtil.close(operationScores);
```
t616178: Would a method with varargs be useful or more readable in some cases? e.g.

CloseableUtil.close(operationScores, authScores)

core/store/src/main/java/uk/gov/gchq/gaffer/store/operation/handler/ScoreOperationChainHandler.java
```diff
@@ -0,0 +1,265 @@
...
+                throw new IllegalArgumentException("Failed to load store properties file : " + e
+                        .getMessage(), e);
+            } finally {
+                CloseableUtil.close(stream);
+            }
+        }
+        return props;
+    }
+
+    private static <K, V extends Comparable<? super V>> Map<K, V> sortByValue(final Map<K, V> map) {
```
t616178: Alternatively:

```java
return map.entrySet()
          .stream()
          .sorted(Map.Entry.comparingByValue())
          .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue, (v1, v2) -> v1, LinkedHashMap::new));
```

Not sure which is more readable?

p013570: Good question. Both are a complex, so I don't think it is worth making the change.

Issue #998, opened 2017-06-19
Create serialisers for common Gaffer graph objects


p013570:
When using products like hazelcast or mapDB, to optimise the performance you need to provide serialisers for the key and value classes used in maps.

Also, it is common for store implements to need to implement serialisation for classes like Elements, Properties, etc.

So, we should create some serialisers for the following classes:

- Element
- Edge
- Entity
- Properties
- ElementId
- EdgeId
- EntityId

p013570:
Merged into develop.

PR #999, opened 2017-06-19
gh-998 - added some useful serialisers

p013570:


p013570:
Closing while I add some more tests.

core/data/src/main/java/uk/gov/gchq/gaffer/data/element/GroupedProperties.java
```diff
@@ -0,0 +1,63 @@
...
+
+    public String getGroup() {
+        return group;
+    }
+
+    public void setGroup(final String group) {
+        this.group = group;
+    }
+
+    @Override
```
m55624: This equals method should match Apache Commons Lang 3 style.  See https://github.com/gchq/Gaffer/issues/864

PR #1008, opened 2017-06-23
Gh 936 get elements documentation

d47853:
Not sure if this is all you wanted doing


core/operation/src/main/java/uk/gov/gchq/gaffer/operation/impl/get/GetElements.java
```diff
@@ -36,7 +36,6 @@
  * Gets elements from Gaffer based on {@link ElementId}s as
  * seeds and returns {@link uk.gov.gchq.gaffer.data.element.Element}s
  * There are various flags to filter out the elements returned.
- * See extensions of {@link GetElements} for further details.
  */
 public class GetElements implements
```
p013570: Please could you update the class javadoc too, to include some information on how the different flags work together.

p013570: For example, explaining what SeedMatchingType EQUALS and RELATED.

Some quick notes:
 * An Entity is RELATED to an EntityId if the Entity's vertex is equal to the EntityId vertex
 * An Entity is RELATED to an EdgeId if either the Entity's vertex is equal to either the EdgeId's source or destination
 * An Edge is RELATED to an EntityId if either the Edge's source or destination matches the EntityId's vertex
 * An Edge is RELATED to an EdgeId if the Edge's source, destination and directed values are equal to the EdgeId. Note that the EdgeId directed value could be EITHER, this will mean Edges that are DIRECTED OR UNDIRECTED will be matched.

PR #1009, opened 2017-06-26
Gh 892 map store updates

p013570:
The main changes in this pull request:

- added a MapFactory so users can provide a more complex map implementation like Hazelcast or MapDB.
- elements are now split up into different maps based on their groups
- non aggregated and aggregated elements are now in separate maps
- added a store property to configure batching elements up and applying aggregation before adding them to the existing map store


store-implementation/map-store/README.md
```diff
@@ -32,3 +37,18 @@ Some examples of how this can be used are:
 It allows very quick calculation of the total number of elements in the graph subject to the default view.
 
 Note that this store requires that the classes used for the vertices, and for all the group-by properties, have an implementation of the hashCode() method.
+
+To configure your choice of Map implementation you can either:
+
+- Set this "gaffer.store.mapstore.map.class" store property to be you chosen Map implementation class. This will require that java class to be on the classpath.
```
gaffer01: 'you' should be 'your'

store-implementation/map-store/README.md
```diff
@@ -32,3 +37,18 @@ Some examples of how this can be used are:
...
+
+or
+
+- For more complex Map implementation that require additional configuration and tuning you can use the "gaffer.store.mapstore.map.factory" map factory store property.
+This allows you to implement your own map factory to use different Map implementations like Hazelcast and MapDB. 
+However due to the nature of having to query-update-put in order to add a new element other implementations may be slow. 
+In addition you can provide you map factory with configuration using the "gaffer.store.mapstore.map.factory.config" store property.
+
+This map store does not attempt to handle concurrent adding of elements. Elements should
```
gaffer01: No new line after 'should'

store-implementation/map-store/src/main/java/uk/gov/gchq/gaffer/mapstore/impl/AddElementsHandler.java
```diff
@@ -31,147 +32,123 @@
...
+
+        // Copy properties that exist in the schema
+        final SchemaElementDefinition elementDef = schema.getElement(element.getGroup());
+        if (elementDef.getProperties().equals(element.getProperties().keySet())) {
+            elementClone.copyProperties(element.getProperties());
+        } else {
+            for (final String property : elementDef.getProperties()) {
+                elementClone.putProperty(property, element.getProperty(property));
+            }
+        }
```
gaffer01: Is the test in the if necessary? The `elementClone.copyProperties` method just iterates through the provided properties and adds them. The for loop in the else code does the same thing. Either way the result is the same - only properties from `elementDef` get added (which is the right thing to do). Comparing two sets to see if they're the same is possibly a bit expensive (and this method will be called a lot).

store-implementation/map-store/src/main/java/uk/gov/gchq/gaffer/mapstore/impl/MapImpl.java
```diff
@@ -15,61 +15,248 @@
...
-    final Set<String> groupsWithNoAggregation = new HashSet<>();
-    final Schema schema;
-    final List<String> aggregatedGroups;
-
-    public MapImpl(final Schema schema, final MapStoreProperties mapStoreProperties) throws StoreException {
+    /**
+     * aggElements maps from an Element containing the group-by properties
+     * to a Properties object without the group-by properties
+     */
+    private final Map<String, Map<Element, GroupedProperties>> aggElements = new HashMap<>();
```
gaffer01: The original design tried to minimise the number of map look-ups that needed to be made before an element could be aggregated into the existing data. The element in the `Map<Element, Properties> elementToProperties` contained the group so only 1 look-up was needed. Now 2 are needed - 1 to get the right map for the group, and another to get the properties for the element. As the first look-up is always a local one (in a Java HashMap) the performance probably isn't worth worrying about, but is there any advantage in having the group as a key?

The Javadoc above this line needs updating as well.

store-implementation/map-store/src/main/java/uk/gov/gchq/gaffer/mapstore/impl/MapImpl.java
```diff
@@ -15,61 +15,248 @@
...
+
+    /**
+     * nonAggElements maps from a non aggregated Element to the count of the
+     * number of times that element has been seen.
+     */
+    private final Map<String, Map<Element, Integer>> nonAggElements = new HashMap<>();
```
gaffer01: Think it's best to replace Integer by Long.

store-implementation/map-store/src/main/java/uk/gov/gchq/gaffer/mapstore/impl/MapImpl.java
```diff
@@ -15,61 +15,248 @@
...
  */
 public class MapImpl {
+    public static final String AGG_ELEMENTS = "aggElements";
+    public static final String NON_AGG_ELEMENTS = "nonAggElements";
+    public static final String ENTITY_ID_TO_ELEMENTS = "entityIdToElements";
+    public static final String EDGE_ID_TO_ELEMENTS = "edgeIdToElements";
+    public static final Set<String> MAP_NAMES = Collections.unmodifiableSet(
+            Sets.newHashSet(AGG_ELEMENTS, NON_AGG_ELEMENTS));
+    public static final Set<String> MULTI_MAP_NAMES = Collections.unmodifiableSet(
+            Sets.newHashSet(ENTITY_ID_TO_ELEMENTS, EDGE_ID_TO_ELEMENTS));
```
gaffer01: `MAP_NAMES` and `MULTI_MAP_NAMES` aren't used anywhere - are they necessary?

p013570: I added these in when I was investigating MapDB and Hazelcast implementations, probably don't really need them.

store-implementation/map-store/src/main/java/uk/gov/gchq/gaffer/mapstore/impl/MapImpl.java
p013570: OK I'll make that change.

p013570:
@gaffer01 thanks for reviewing this. I've made your suggested changes.

store-implementation/map-store/src/main/java/uk/gov/gchq/gaffer/mapstore/impl/MapImpl.java
p013570: I made the change to make the maps a bit smaller. This will reduce the number of hashcode collisions and should make it a bit faster. The Outer map is always a HashMap containing probably only a small number of groups so that initial look up will be very fast. In the future we could then optimise some of the queries to only look in the maps for the groups requested.

Issue #328, opened 2016-07-18
Bulk import to Accumulo store


gaffer01:
To bulk import data to a new table in Accumulo, the following steps are necessary:
- Estimate split points from a sample of the data you want to ingest.
- Create table.
- Set the split points on the table.
- Convert the data to rfiles of Accumulo keys and values.
- Hand those files to Accumulo.

To help bulk import data to a new graph backed by an Accumulo store, the following operations are provided:
- `SampleDataForSplitPoints` - this creates a file of split points.
- `SplitTable` - this splits a table but does not create the table.
- `AddElementsFromHdfs` - this creates the table if it does not exist, but does not add split points. It hands the data over to Accumulo.

To use these operations to bulk import to a new graph, it is necessary to run `SampleDataForSplitPoints`, then run `SplitTable`, then run `AddElementsFromHdfs`. This is overly complex, and not documented properly.

It should be possible to run one operation, which bulk imports data to a new graph - this operation should run all the above steps. Chaining the `SampleDataForSplitPoints`, `SplitTable` and `AddElementsFromHdfs` operations together should work. It would be useful to create one operation which runs all of these.

p013570:
Although it makes it simpler for a user having just 1 operation to do all the above steps, if we did this for lots of different cases then we could end up with a huge number of operations which would be confusing. Perhaps we just need to improve the documentation and create an example operation chain in the examples section of the wiki.

p013570:
Perhaps this is something that could be achieved via a named operation - gh-668.

p013570:
OK, perhaps we could just update the AddElementsFromHdfs operation to delegate to SampleDataForSplitPoints and SplitTable operations first if the table has no split points already defined.

p013570:
# Migration Steps

The following operations have been renamed:
- uk.gov.gchq.gaffer.accumulostore.operation.hdfs.operation.SampleDataForSplitPoints -> uk.gov.gchq.gaffer.hdfs.operation.SampleDataForSplitPoints
- uk.gov.gchq.gaffer.accumulostore.operation.hdfs.operation.SplitTable -> uk.gov.gchq.gaffer.operation.impl.SplitStore


These magic AddElementsFromHdfs operation options have been replaced with fields on the operation:
- accumulostore.operation.hdfs.use_accumulo_partitioner -> partitioner (if you wish to disable using a partitioner, you can set the partitioner class to uk.gov.gchq.gaffer.hdfs.operation.partitioner.NoPartitioner)
[...]

PR #1010, opened 2017-06-26
Gh 328 add from hdfs

p013570:


p013570:
This is the main fail to focus on: [AddElementsFromHdfsHandler](https://github.com/gchq/Gaffer/blob/gh-328-add-from-hdfs/store-implementation/accumulo-store/src/main/java/uk/gov/gchq/gaffer/accumulostore/operation/hdfs/handler/AddElementsFromHdfsHandler.java)

p013570:
See the migration steps for the issue https://github.com/gchq/Gaffer/issues/328 for a list of the significant changes.

library/hdfs-library/src/main/java/uk/gov/gchq/gaffer/hdfs/operation/mapper/GafferMapper.java
```diff
@@ -20,16 +20,18 @@
...
 import java.io.IOException;
 import java.io.UnsupportedEncodingException;
 
+import static uk.gov.gchq.gaffer.hdfs.operation.handler.job.factory.AddElementsFromHdfsJobFactory.MAPPER_GENERATOR;
+import static uk.gov.gchq.gaffer.hdfs.operation.handler.job.factory.AddElementsFromHdfsJobFactory.SCHEMA;
+import static uk.gov.gchq.gaffer.hdfs.operation.handler.job.factory.AddElementsFromHdfsJobFactory.VALIDATE;
 
 /**
- * An <code>AbstractAddElementsFromHdfsMapper</code> is a {@link Mapper} that uses a
+ * An <code>GafferMapper</code> is a {@link Mapper} that uses a
```
gaffer01: Minor - Should be "A" not "An"

store-implementation/accumulo-store/src/main/java/uk/gov/gchq/gaffer/accumulostore/operation/hdfs/handler/AddElementsFromHdfsHandler.java
```diff
@@ -52,9 +79,98 @@ public void doOperation(final AddElementsFromHdfs operation, final AccumuloStore
...
+
+        // Sample data for split points and split the table
+        final Class<? extends MapperGenerator> mapperGeneratorClass;
+        try {
+            mapperGeneratorClass = Class.forName(operation.getMapperGeneratorClassName()).asSubclass(MapperGenerator.class);
+        } catch (final ClassNotFoundException e) {
+            throw new IllegalArgumentException("Mapper generator class name was invalid: " + operation.getMapperGeneratorClassName(), e);
+        }
+
+        final String outputPath = "/tmp/" + context.getJobId() + "/output";
```
gaffer01: Should this be taken from an option on the operation? Putting split points from a table of data which has visibilities to protect the data into /tmp could be unsecure.

p013570:
@gaffer01 - thanks, good spot. I've created a workingPath field on AddElementsFromHdfs and we will use that instead of the /tmp.

p013570:
I think we should add a "numSplits" field to SplitStore as it might be useful for other store implementations.

Also, there are some edge cases when you might not want Accumulo to Sample and Split before running the AddElementsFromHdfs, so we will need to provide an option to disable that.

Closing while we implement the above.

PR #1015, opened 2017-06-27
gh-855 serialisation migration mitigation

GCHQDeveloper404:
Testing added serialisation against static primitive byte[].


GCHQDeveloper404:
Extra Serialisers have been added and need changing.

PR #1021, opened 2017-06-28
Gh 987 update to string builders

m55624:


core/common-util/src/main/java/uk/gov/gchq/gaffer/commonutil/DebugUtil.java
```diff
@@ -0,0 +1,51 @@
...
+    public static final String DEBUG = "gaffer.error-mode.debug";
+    public static final String DEBUG_DEFAULT = String.valueOf(false);
+    public static boolean isDebug;
+    private static final Logger LOGGER = LoggerFactory.getLogger(DebugUtil.class);
+
+    private DebugUtil() {
+        // this class should not be instantiated - it contains only util methods and constants.
+    }
+
+    public static boolean checkDebugMode() {
```
p013570: We shouldn't be looking up the system property every time this method is called. This logic should go in the updateDebugMode.

Perhaps rename this method to isInDebugMode(). If isDebug is null then call updateDebugMode.

Then create a isInDebugMode getter that just returns

core/common-util/src/main/java/uk/gov/gchq/gaffer/commonutil/collection/LimitedSortedSet.java
```diff
@@ -164,7 +164,7 @@ public String toString() {
         return new ToStringBuilder(this)
                 .append("limit", limit)
                 .append("backingSet", backingSet)
-                .toString();
+                .build();
```
p013570: Just leave this as toString().

core/common-util/src/main/java/uk/gov/gchq/gaffer/commonutil/pair/Pair.java
```diff
@@ -76,22 +76,22 @@ public boolean equals(final Object obj) {
...
+                .toHashCode();
     }
 
     @Override
     public String toString() {
         return new ToStringBuilder(this)
                 .append("first", first)
                 .append("second", second)
-                .toString();
+                .build();
```
p013570: Just leave this as toString().

core/exception/src/main/java/uk/gov/gchq/gaffer/core/exception/Error.java
```diff
@@ -98,36 +95,16 @@ public String toString() {
         return new ToStringBuilder(this)
                 .append("statusCode", statusCode)
                 .append("simpleMessage", simpleMessage)
-                .toString();
+                .build();
```
p013570: Just leave this as toString().

core/common-util/src/main/java/uk/gov/gchq/gaffer/commonutil/DebugUtil.java
m55624: isDebug cannot be null as it is a boolean.  The idea of it setting on checkDebugMode() stops a user having to run updateDebugMode() then isInDebugMode() every time in case it changes or is set somewhere after the last time it has been updated.

p013570: I imagined:

```java
public static Boolean isDebug;

public static void isInDebugMode() {
    if(null == isDebug) {
        updateDebugMode();
    }

    return isDebug;
}
```

Every time toString is called on an object this method will be called - so that could be a lot. The value of the system property should never really change and if a user decides to change the system property value then it is easy enough for them to call updateDebugMode at the same time. You could have another method - setDebugMode() that sets the system property and updates the isDebug value.

Issue #113, opened 2016-03-22
Enable AddElementsFromHdfs to be executed via REST


p013570:
Currently AddElementsFromHdfs operations have to be run as a hadoop job directly on hdfs. It would be useful to be able to trigger the operation via the REST API.

p013570:
This should done after gh-129.

p013570:
After some thought, there are some operations (let's call them advanced operations) that we should just disable via the REST API. So we will create a new store property 'gaffer.store.operation.advanced.mode' to toggle advanced mode. When we are not in advanced mode (i.e. when running over a rest api) we should disable these advanced operations such as AddElementsFromHdfs.

By default we will set the advanced mode to true. The rest api DefaultGraphFactory will disable advanced mode.

p013570:
Getting this working via the rest api would be quite difficult due to conflicting dependencies. For now the rest api will now just automatically disable this operation.

PR #1022, opened 2017-06-29
gh-113 - added advanced mode

p013570:
By default Stores will register all operations. When advanced mode is disabled, the advanced operations (e.g. AddElementsFromHdfs) are not registered. This means when a Graph is instantiated for use via the rest api, advanced mode can be disabled and the advanced operations are then not available via the rest api.

I'm not completely convinced with the "advanced mode" idea. Perhaps it needs a different name. What do you think?


gaffer01:
Not sure. Maybe the general pattern is that a store can have an optional list of enabled operations in its properties? So admins can instantiate a store with all options enabled, but the store that is used in a REST API will only have a small number of operations enabled?

p013570:
@gaffer01 perhaps yeah. I'm not quite sure how that would look though.

Currently admins can enable/disable any operations by providing an operations declarations file when the store is initialised. This feature is hidden away a bit but it is possible.

gaffer01:
If we can already provide similar functionality with the declarations file, then I don't think "advanced" makes sense. The only option I can think of is a "don't enable additional operations" option. If this is on, then you only get to use the default operations that all stores provide.

gaffer01:
Another thought: is there any value in having a "read-only" mode? If this is on then you can't run any operation that changes the graph.

p013570:
I think there are some operations like AddElementsFromHdfs, AddElementsFromRdd, GetRddOfElements etc. that should not be executed over the rest api. It would be nice if there was a way the core rest api could disable these operations rather than every admin having to write their own operation declaration file to disable them.

p013570:
OK I'll try using the operation declarations to disable them. I'll close this pull request for now.

Issue #954, opened 2017-06-12
Create a Graph Library to store and retrieve Schemas and StoreProperties


p013570:
Currently every time you construct a Graph instance you need to provide the original Schema and store properties. If the Graph instance is constructed with a different Schema then it will conflict with the original Schema and cause problems.

The Schema and Store Properties should be persisted by the Graph using a GraphLibrary so it doesn't have to be provided every time a user constructs a new Graph instance.

To implement this change we will need to introduce a few new IDs. A Schema ID, a Store Properties ID and a Graph ID. The Graph ID should be used to uniquely define a graph - it will be associated with a Schema ID and a Store Properties ID. 

We should write a GraphLibrary interface that has methods for persisting and retrieving schemas and store properties.

Then we can update the Graph.Builder class to allow Graphs to be built using just a Graph ID (and GraphLibrary), in which the Schema and Store Properties can be retrieved from the GraphLibrary.

Alternatively a user could create a new Graph with a new Graph ID, new Schema, but refer to existing Store Properties using a Store Properties ID.

In the case of accumulo and hbase, we should remove the 'accumulo.table' and 'hbase.table' properties and use the graphId instead. 

Note the Schema persisted should be the full (non-compact) schema including all the description fields.

p013570:
# Migration Steps
When creating an instance of Graph (or initialising a Store) you should provide a graphId. This is a unique identifier String for your Graph. You can now configure your store to persist the graphId and schema you initialise it with, so following calls to instantiate a Graph do not need to provide the schema again.

To configure your store to persist your schema add the following to your store.properties:
```properties
gaffer.store.schema.library.class=uk.gov.gchq.gaffer.store.schema.library.FileSchemaLibrary
gaffer.store.schema.library.path=/path-to-a-folder-to-store-the-schemas/schemaLibrary
```

The table name store property for Accumulo and HBase has been deprecated. This has been replaced with the graphId. For existing systems, you can continue using the table property instead of the graphId for now (the graphId will just be set to your table name), but it will be removed in the near future.

p013570:
Merged into develop.

PR #1023, opened 2017-06-29
gh-954 - renamed JsonUtil to JsonAssert

p013570:
Created a JsonUtil class to compare json objects.

This is required for gh-954. The rest of the changes for gh-954 will come later.


p013570:
@m55624 thanks for adding the null test and fixing the bug. Looks good.

PR #1024, opened 2017-06-29
gh-954 - created getTableName method

p013570:
Added this for the Accumulo and HBase stores. This is required for issue gh-954. The rest of the changes are coming later


m55624:
Could this getTableName() method not just be put within Store.java?  Avoids duplicating it in HBase and Accumulo and any other stores in the future.

p013570:
@m55624 it is not really applicable to other stores, so I think it would be better to just keep it in the HBase and Accumulo store classes.

PR #1025, opened 2017-06-30
Add a graphId to uniquely identify an instance of a Gaffer graph

p013570:
A large number of files have changed, but the change in the majority of files is just to simply added a graphId whenever a Graph is instantiated or a Store is initialised.

I have also deprecated the 'table' property in HBase and Accumulo, so these stores will now use the graphId as a table name.

This is needed for issue gh-954. The rest of the changes for that issue will come later. The final part of that issue is to add the Graph Library. The Graph Library needs this graphId to uniquely identify a schema and store properties.

Is everyone happy with this idea of requiring a graphId when you create a Gaffer graph?


gaffer01:
How do we deal with uniqueness of IDs? If I'm creating multiple graphs in the same Accumulo instance, then it is easy to check there isn't already a graph with that id (if we're using the id as the table name).

But what if multiple Accumulo instances are being used? Will the graph need to check with the schema registry to ensure it's not re-using an ID?

p013570:
When a graph is initialised and you are using a Schema Library then it will check to see if the graphId already exists, if it does and the schema is different then it will error.

store-implementation/accumulo-store/src/main/java/uk/gov/gchq/gaffer/accumulostore/AccumuloStore.java
```diff
@@ -131,20 +131,31 @@
...
      * @param properties the accumulo store properties
      * @throws StoreException the store could not be initialised.
      */
-    public void preInitialise(final Schema schema, final StoreProperties properties) throws StoreException {
-        super.initialise(schema, properties);
+    public void preInitialise(final String graphId, final Schema schema, final StoreProperties properties) throws StoreException {
+        final String deprecatedTableName = ((AccumuloProperties) properties).getTable();
+        if (null == graphId && null != deprecatedTableName) {
+            // Deprecated
+            super.initialise(deprecatedTableName, schema, properties);
```
m55624: Should this not be completely removed to stop users initialising an AccumuloStore with a graphId of the table name but possibly not realising they have done it?  Same for HBaseStore.

store-implementation/hbase-store/src/main/java/uk/gov/gchq/gaffer/hbasestore/utils/TableUtils.java
```diff
@@ -68,22 +68,48 @@ private TableUtils() {
...
             System.err.println("Wrong number of arguments. \nUsage: "
-                    + "<schema directory path> <store properties path>");
+                    + "<graphId> <schema directory path> <store properties path>");
             System.exit(1);
         }
 
-        final HBaseStore store = new HBaseStore();
-        store.preInitialise(Schema.fromJson(Paths.get(args[0])),
-                HBaseProperties.loadStoreProperties(args[1]));
+        final StoreProperties storeProps = StoreProperties.loadStoreProperties(args[2]);
```
m55624: Possibly neater to create private static methods, as in AddUpdateTableIterator to retrieve the args values.  Also makes it a lot easier to read.

store-implementation/accumulo-store/src/main/java/uk/gov/gchq/gaffer/accumulostore/AccumuloStore.java
p013570: This will avoid making a breaking change and give users some time to make changes to their code whilst the table name property is deprecated. When we remove the table name altogether in a later version we will remove this too.

PR #1030, opened 2017-07-03
Gh 1011 serialiser and aggregator for hllsketch

gaffer01:


library/sketches-library/src/main/java/uk/gov/gchq/gaffer/sketches/datasketches/cardinality/binaryoperator/HllUnionAggregator.java
```diff
@@ -0,0 +1,33 @@
+/*
```
p013570: Is it worth adding another walkthrough for the doc module for HllpUnion, similar to UnionSketch.java? Otherwise these serialisers and aggregators may just get lost and forgotten about.

gaffer01:
Closing while suggested changes are made.

gaffer01:
Created ReservoirItemsSketchSerialiser and made ReservoirStringsSketchSerialiser and ReservoirNumbersSketchSerialiser extend it to reduce code duplication as requested.

Suggest creating a separate issue to update the examples/documentation for the sketches library. The examples are currently contained in the properties guide - this could do with a better introduction to what sketches are, which libraries we use and the differences between the unions and the sketches.

p013570:
OK - I'm happy for the examples to be updated in a separate issue - can you create that? Your changes look nice.

Issue #538, opened 2016-11-09
Make REST API more "RESTful"


t616178:
The current REST API does not look like a typical RESTful API (based around resource collections and the application of HTTP methods against those collections).

Could the REST API be refactored slightly to align more closely with REST best practice?

E.g.

<pre><code>POST http://<host>:<port>/api/graph/get/elements/related
{
  "seeds" : [ "elementSeed1", "elementSeed2" ]
}</pre></code>

Could become:

<pre><code>GET http://<host>:<port>/api/graph/elements/<elementSeed1>,<elementSeed2>/related</pre></code>

Use of path & matrix parameters could eliminate some of the POST bodies currently required for some of the simple element retrieval operations.

p013570:
It would be nice to make the api more like a typical RESTful api, but things like putting the seeds in the url is not possible as there is a limit on the number of characters in a url, therefore users would then be limited to the number of seeds they could supply.

Perhaps we should rename the REST API to WEB API?

p013570:
Merged into develop.

t616178:
# Migration steps

Version 2 of the Gaffer REST API contains a number of changes which require user intervention.

* The OperationChain class is now a subclass of Operation. In order for deserialisation to work correctly, any operation chains written in JSON and submitted to the REST API must contain the type ID in the JSON body. For example:

```javascript
{
	"class" : "uk.gov.gchq.gaffer.operation.OperationChain",
	"operations" : [
		...
	]
}
```

* The API now has four distinct sections:

/graph				The graph root resource
/graph/config		Resources relating to the graph configuration
/graph/jobs			Resources relating to submitting Jobs to the graph
/graph/operations	Resources relating to executing operations on the graph

A full list of endpoints (with descriptions) is provided below:

GET 	/graph/ 									status page of some description? or redirect to home page? a summary of all graph information?

Operations

GET		/graph/operations 							list of all operations supported by the current store
POST	/graph/operations/execute					perform the submitted operation
POST	/graph/operations/execute/chunked			perform the submitted operation
POST 	/graph/operations/generate/elements			generate elements, as specified in the request body
POST 	/graph/operations/generate/objects			generate objects, as specified in the request body
GET 	/graph/operations/{className}				get information about the specified operation class
GET 	/graph/operations/{className}/example		get an example JSON body for the specified operation class
GET 	/graph/operations/{className}/next			list of all acceptable operations which can be performed after the specified operation

Jobs

[...]

PR #1036, opened 2017-07-04
gh-538 - Make REST API more RESTful and update to version 2.

t616178:
Some questions regarding the change to the REST API in version 2:

* I have added a custom media type to indicate to the REST client which version of the API their request has been executed against. I can remove this if we think it's not useful, or we could implement this differently (for example, by adding a custom header to the response)
* I've added a additional endpoint (/latest) which redirects to the latest version of the API - is this useful?
* At the moment, only one version of the API can be served up through Swagger - this needs to be configured by the application. Are there use cases where we would want to allow access to all versions of the API via Swagger?
* Some methods have been changed to return Jersey ```Response``` objects - this allows for greater control over the object which is returned to the calling service (e.g. headers, status codes etc). Is it worth standardising this across the services so that all methods are treated the same?


p013570:
/latest sounds useful

Yeah I think we will need to expose both versions in swagger.

Standardising the responses would be sensible, but only for the new version of the rest api.

PR #1037, opened 2017-07-04
Gh 824 flink library

p013570:


library/flink-library/src/main/java/uk/gov/gchq/gaffer/flink/operation/handler/AddElementsFromFileHandler.java
```diff
@@ -0,0 +1,47 @@
...
+import uk.gov.gchq.gaffer.operation.OperationException;
+import uk.gov.gchq.gaffer.operation.impl.add.AddElementsFromFile;
+import uk.gov.gchq.gaffer.store.Context;
+import uk.gov.gchq.gaffer.store.Store;
+import uk.gov.gchq.gaffer.store.operation.handler.OperationHandler;
+
+public class AddElementsFromFileHandler implements OperationHandler<AddElementsFromFile> {
+    @Override
+    public Object doOperation(final AddElementsFromFile op, final Context context, final Store store) throws OperationException {
+        final StreamExecutionEnvironment env = StreamExecutionEnvironment.getExecutionEnvironment();
```
javadev001001: Just a thought - it might be better to use Flinks batch API to load from a file rather than the Streaming API.

PR #1047, opened 2017-07-06
Gh 954 graph library

m55624:


core/graph/src/main/java/uk/gov/gchq/gaffer/graph/library/FileGraphLibrary.java
```diff
@@ -0,0 +1,120 @@
...
+        try {
+            return path.toFile().exists() ? Files.readAllBytes(path) : null;
+        } catch (IOException e) {
+            throw new SchemaException("Unable to read schema bytes from file: " + getSchemaPath(graphId));
+        }
+    }
+
+    @Override
+    protected StoreProperties _getProperties(final String propertiesId) {
+        return StoreProperties.loadStoreProperties(getPropertiesPath(propertiesId));
```
p013570: You will need to put a null check in to check the file getPropertiesPath(propertiesId) actually exists and has properties in. A user could provide any propertiesId they want and it might not exist.

core/graph/src/main/java/uk/gov/gchq/gaffer/graph/library/FileGraphLibrary.java
```diff
@@ -0,0 +1,120 @@
...
+    }
+
+    private Path getSchemaPath(final String schemaId) {
+        return Paths.get(path + "/" + schemaId + ".json");
```
p013570: schemaId and propertiesId could be the same. So you will need to add a suffix/prefix onto the path like with getGraphsPath - it adds "Graphs" onto the end of the file name.

core/graph/src/main/java/uk/gov/gchq/gaffer/graph/library/FileGraphLibrary.java
```diff
@@ -0,0 +1,120 @@
...
+    }
+
+    private Path getPropertiesPath(final String propertiesId) {
+        return Paths.get(path + "/" + propertiesId + ".json");
```
p013570: I think the properties file created will actually be a .properties file not .json.

core/graph/src/test/java/uk/gov/gchq/gaffer/graph/library/FileGraphLibraryTest.java
```diff
@@ -0,0 +1,169 @@
...
+import uk.gov.gchq.gaffer.store.StoreProperties;
+import uk.gov.gchq.gaffer.store.schema.Schema;
+import java.io.File;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
+
+public class FileGraphLibraryTest {
```
p013570: Please can you add a test for when the graphId, schemaId and propertiesId are all the same value.

p013570: Please can you add a test for when the schema is null or the store properties are null for a given id.

store-implementation/accumulo-store/src/main/java/uk/gov/gchq/gaffer/accumulostore/utils/AddUpdateTableIterator.java
```diff
@@ -54,7 +57,7 @@
     public static final String UPDATE_KEY = "update";
     public static final String REMOVE_KEY = "remove";
     public static final String ADD_KEY = "add";
-    private static final int NUM_REQUIRED_ARGS = 4;
+    private static final int NUM_REQUIRED_ARGS = 5;
```
p013570: The pathToFileGraphLibrary is not a required argument. As you mention in the javadoc if it is not provided then NoGraphLibrary will be used instead. Therefore you should set the NUM_REQUIRED_ARGS to 4. Same for HBase.

store-implementation/accumulo-store/src/main/java/uk/gov/gchq/gaffer/accumulostore/utils/AddUpdateTableIterator.java
```diff
@@ -192,6 +198,16 @@ public static void main(final String[] args) throws Exception {
...
 
+        GraphLibrary library;
+
+        if (getFileGraphLibraryPathString(args) == null) {
+            library = new NoGraphLibrary();
+        } else {
+            library = new FileGraphLibrary(getFileGraphLibraryPathString(args));
+        }
+
+        library.add(getGraphId(args), schema, storeProps);
```
p013570: We should call addOrUpdate as we want to update the schema and store properties. Similarly for HBase.

core/graph/src/main/java/uk/gov/gchq/gaffer/graph/library/FileGraphLibrary.java
m55624: Within loadStoreProperties when the input stream is opened it will throw the error if there is no such file.  Also if it is empty does that matter? it may intentionally not have any properties in, and just return an empty Properties?

store-implementation/hbase-store/src/main/java/uk/gov/gchq/gaffer/hbasestore/utils/TableUtils.java
```diff
@@ -156,6 +156,10 @@ private static String getStorePropertiesPathString(final String[] args) {
         return args[2];
     }
 
+    private static String getFileGraphLibraryPathString(final String[] args) {
+        return args[3];
```
p013570: You should check the length of args here. If args.length < 3 then return null;

Similarly for Accumulo but < 4.

PR #1066, opened 2017-07-12
Gh 659 consistent edge dir

p013570:
The main change was to reorder the source and destination of undirected edges to ensure the source is 'less than' the destination.

I have removed the magic operation option - "return_matched_id_as_edge_source" and instead added a matchedVertex field to Edge and EdgeSeed. This will be set when a GetElements query is done, so users will now be able to see which end of the edge was matched by a seed.


core/operation/src/main/java/uk/gov/gchq/gaffer/operation/data/EdgeSeed.java
```diff
@@ -56,58 +77,76 @@ public Object getSource() {
...
+        return matchedVertex;
     }
 
-    @Override
-    public void setDirectedType(final DirectedType directed) {
-        if (null == directed) {
-            this.directed = DirectedType.EITHER;
-        } else {
-            this.directed = directed;
+    private void orderVertices() {
```
t616178: Could this method be moved to EdgeId to reduce duplication? It's shared between the EdgeSeed and Edge classes.

core/operation/src/main/java/uk/gov/gchq/gaffer/operation/data/EdgeSeed.java
```diff
@@ -56,58 +77,76 @@ public Object getSource() {
...
+        if (!DirectedType.isDirected(directed)) {
+            if (VERTEX_COMPARATOR.compare(source, destination) > 0) {
+                swapVertices();
+            }
         }
     }
 
+    private void swapVertices() {
```
t616178: Could this method be moved to EdgeId to reduce duplication? It's shared between the EdgeSeed and Edge classes.

store-implementation/accumulo-store/src/test/java/uk/gov/gchq/gaffer/accumulostore/key/core/AbstractCoreKeyAccumuloElementConverterTest.java
```diff
@@ -60,53 +58,76 @@ public void shouldReturnOverriddenSerialiseNull() throws Exception {
...
+        assertEquals(AbstractCoreKeyAccumuloElementConverter.EdgeDirection.DIRECTED, direction);
     }
 
-    private class TestImplementation extends AbstractCoreKeyAccumuloElementConverter {
-        public TestImplementation(Schema schema) {
-            super(schema);
-        }
+    @Test
+    public void shouldDeserialiseSourceDestinationValuesIncorrectWayRound() {
+        // Given
```
t616178: Is this an extra character that has crept in? I've seen a few of these dotted around the test classes.

store-implementation/hbase-store/src/main/java/uk/gov/gchq/gaffer/hbasestore/serialisation/ElementSerialisation.java
```diff
@@ -42,13 +43,20 @@
...
 import java.util.Arrays;
 import java.util.Iterator;
-import java.util.Map;
 
 public class ElementSerialisation {
     private static final Logger LOGGER = LoggerFactory.getLogger(ElementSerialisation.class);
 
     private final Schema schema;
 
+    protected enum EdgeDirection {
```
t616178: It feels like we have declarations for variations on this "EdgeDirection" enum in a few different places - would it be better to have one implementation in the core module?

core/operation/src/main/java/uk/gov/gchq/gaffer/operation/data/EdgeSeed.java
p013570: Same as above - the method needs to be private and the swapVertices needs access to set the values of source/destination/matchedVertex.

core/operation/src/main/java/uk/gov/gchq/gaffer/operation/data/EdgeSeed.java
p013570: The method needs to be private and the swapVertices needs access to set the values of source/destination/matchedVertex.

store-implementation/hbase-store/src/main/java/uk/gov/gchq/gaffer/hbasestore/serialisation/ElementSerialisation.java
p013570: OK good idea, I've moved it into the data module.

p013570:
@t616178 - thanks. I've made your suggested changes - please can you approve it if you're happy with  it.

PR #1086, opened 2017-07-21
Gh 287 refactor export to other graph handler

GCHQDeveloper404:
ExportToOtherGraphHandler#createGraph

potentially the method fields could be removed and the inner three methods be simply passed export & store. this would be in keeping with the clean practice of small method parameter volume.

### DIFF
--- a/core/graph/src/main/java/uk/gov/gchq/gaffer/operation/export/graph/handler/ExportToOtherGraphHandler.java
+++ b/core/graph/src/main/java/uk/gov/gchq/gaffer/operation/export/graph/handler/ExportToOtherGraphHandler.java
@@ -63,19 +63,16 @@ public class ExportToOtherGraphHandler extends ExportToHandler<ExportToOtherGrap
     }
 
     private Graph createGraphAfterResolvingSchemaAndProperties(final ExportToOtherGraph<?> export, final Store store) {
-        final Graph rtn;
-
         StoreProperties storeProperties = resolveStoreProperties(export, store);
 
         Schema schema = resolveSchema(export, store);
 
-        rtn = new Builder()
+        return new Builder()
                 .graphId(export.getGraphId())
                 .library(store.getGraphLibrary())
                 .addSchema(schema)
                 .storeProperties(storeProperties)
                 .build();
-        return rtn;
     }
 
     private StoreProperties resolveStoreProperties(final ExportToOtherGraph<?> export, final Store store) {
--- a/core/graph/src/main/java/uk/gov/gchq/gaffer/operation/export/graph/handler/ExportToOtherGraphHandler.java
+++ b/core/graph/src/main/java/uk/gov/gchq/gaffer/operation/export/graph/handler/ExportToOtherGraphHandler.java
@@ -40,160 +40,129 @@ public class ExportToOtherGraphHandler extends ExportToHandler<ExportToOtherGrap
     @Override
     protected OtherGraphExporter createExporter(final ExportToOtherGraph export, final Context context, final Store store) {
 
-        return new OtherGraphExporter(
-                context.getUser(),
-                context.getJobId(),
-                createGraph(export, store));
+        return new OtherGraphExporter(context.getUser(),
+                                      context.getJobId(),
+                                      createGraph(export, store));
     }
 
     protected Graph createGraph(final ExportToOtherGraph<?> export, final Store store) {
-        validationHandling(store, export);
+        validate(store, export);
 
         final String exportGraphId = export.getGraphId();
-        final Schema exportSchema = export.getSchema();
-        final StoreProperties exportStoreProperties = export.getStoreProperties();
-        final List<String> exportParentSchemaIds = export.getParentSchemaIds();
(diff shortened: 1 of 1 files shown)


core/graph/src/main/java/uk/gov/gchq/gaffer/operation/export/graph/handler/ExportToOtherGraphHandler.java
```diff
@@ -46,97 +47,146 @@ protected OtherGraphExporter createExporter(final ExportToOtherGraph export, fin
     }
 
     protected Graph createGraph(final ExportToOtherGraph<?> export, final Store store) {
+        validationHandling(store, export);
 
         final String exportGraphId = export.getGraphId();
```
p013570: I don't think we really need some of these variables. Then you can just pass the export operation to the methods instead of passing so many variables around.

core/graph/src/main/java/uk/gov/gchq/gaffer/operation/export/graph/handler/ExportToOtherGraphHandler.java
```diff
@@ -46,97 +47,146 @@ protected OtherGraphExporter createExporter(final ExportToOtherGraph export, fin
...
+                .graphId(exportGraphId)
+                .library(graphLibrary)
+                .addSchema(schema)
+                .storeProperties(storeProperties)
+                .build();
+        return rtn;
+    }
+
+    private StoreProperties resolveStoreProperties(final StoreProperties exportStoreProperties, final String exportParentStorePropertiesId, final GraphLibrary graphLibrary, final StoreProperties storeStoreProperties) {
+        StoreProperties storeProperties = null;
```
m55624: If we are using rtn as the returned object above can we stick with it here? Also applies to resoveSchema below.

core/graph/src/main/java/uk/gov/gchq/gaffer/operation/export/graph/handler/ExportToOtherGraphHandler.java
```diff
@@ -46,97 +47,146 @@ protected OtherGraphExporter createExporter(final ExportToOtherGraph export, fin
...
+        final StoreProperties properties = (null == exportStoreProperties) ? storeStoreProperties : exportStoreProperties;
+        return new Builder()
+                .graphId(exportGraphId)
                 .addSchema(schema)
-                .storeProperties(storeProperties)
+                .storeProperties(properties)
                 .build();
     }
 
+    private void validationHandling(final Store store, final ExportToOtherGraph<?> export) {
```
p013570: I think we can probably just merge this validationHandling into logic into the validate method and make the validate method private.

core/graph/src/main/java/uk/gov/gchq/gaffer/operation/export/graph/handler/ExportToOtherGraphHandler.java
```diff
@@ -46,97 +47,146 @@ protected OtherGraphExporter createExporter(final ExportToOtherGraph export, fin
...
-        if (!validationResult.isValid()) {
-            throw new IllegalArgumentException(validationResult.getErrorString());
+        if (null == graphLibrary) {
+            rtn = getGraphWithoutLibrary(exportGraphId,
+                                         exportStoreProperties,
+                                         storeStoreProperties,
+                                         exportSchema,
+                                         storeSchema);
+        } else if (graphLibrary.exists(exportGraphId)) {
+            rtn = getGraphWithLibraryAndID(exportGraphId, graphLibrary);
```
p013570: perhaps call the methods createGraphX instead of getGraphX

core/graph/src/main/java/uk/gov/gchq/gaffer/operation/export/graph/handler/ExportToOtherGraphHandler.java
m55624: Could we add all the validate to the validationHandling to just make one validate method?
This would also avoid the use of calling export.getX() twice as this is already done in the createGraph method once.

core/graph/src/main/java/uk/gov/gchq/gaffer/operation/export/graph/handler/ExportToOtherGraphHandler.java
GCHQDeveloper404: I fully agree as I described in the comments of this pull request.

m55624: This was originally like that, as described above, but to limit the usage of calls we decided to pass individual variables.  I don't disagree with this though.

core/graph/src/main/java/uk/gov/gchq/gaffer/operation/export/graph/handler/ExportToOtherGraphHandler.java
GCHQDeveloper404: I didn't see that it was validation within the same class.

Issue #1060, opened 2017-07-10
Allow multiple properties to be aggregated using a single aggregate function


p013570:
We need to update the SchemaElementDefinition class to allow a ElementAggregator to be provided. Then getFullAggregator, getIngestAggregator and getQueryAggregator methods will need to be updated to use combine the Type aggregators and the schema element definition aggregator.

We need to restrict which properties can be aggregated at the same time. For example, you should not be allow to aggregate a groupBy property and a non-groupBy property in the same Binary Operator. When ingest aggregation is applied, only the non-groupBy properties are aggregated so we need Binary Operators that only apply to non-groupBy properties. The different types of properties are: groupBy, non-groupBy, visibility, timestamp. 

As schemas can be split up into different files we probably should have another think about the structure of these files and in which file the new ElementAggregator functions go. The aggregation logic is Store specific so it shouldn't be mixed up with the other data schema parts. 

We could change the schema files (dataSchema, dataTypes, storeTypes) to be something like:

- schema
- types
- serialisation   (store specific)
- validation    (overrides type validation and allows you to validate multiple properties at once)
- aggregation   (overrides type aggregation and allows you to aggregate multiple properties at once)

p013570:
Requirement for this came from #1058.

gaffer01:
> We need to update the SchemaElementDefinition class to allow a ElementAggregator to be provided. Then getFullAggregator, getIngestAggregator and getQueryAggregator methods will need to be updated to use combine the Type aggregators and the schema element definition aggregator.

Not sure I understand what this means. What does an ElementAggregator do? What if the type aggregators and element aggregators disagree on what to do?

If multiple properties need to be aggregated by one function then they could be put in one property. If you were maintaining a mean and need to update the sum and the number of items then those 2 fields could be made into one property which could then easily be aggregated with a normal aggregation function. Is there a good use-case which can't be satisfied by this approach?

Not sure the extra complexity that this will introduce (which feels like a lot) is justified. Users already find writing a schema the hardest thing to do. Too much generality will result in a framework that is too hard to use.

ddseapy:
I've been doing as you said, bundling them all in a property together.  Essentially I have a couple properties that affect how the rest of the properties get aggregated.  Instead of duplicating each property in all of the rest and writing a custom type for all them it's been easier making them all one property.

An example might be timestamp and prop1 affect the others (int, double, date...), so the options are:

[...]

PR #1087, opened 2017-07-21
Gh 1060 multi property aggregation

p013570:
A lot of the files that have changed are simply the schema json files.

dataSchema.json -> elements.json
dataTypes.json + storeTypes.json -> types.json

In some cases we used a storeSchema.json file that contained a vertex serialiser. This has been renamed to serialisation.json.

Advanced aggregation and validation has been separated out into aggregation.json and validation.json. 

Renaming these schema files is not a breaking change and does not require users to do the same. I just think that for new users it will be clearer and easier to understand if we split the schema into just the elements and the types.

I have written quite a few unit tests to check the logic for the multi property schema aggregation, but it is a bit tricky so please review that bit carefully.

The change has added an extra section to the Schema but this is optional so hopefully it won't complicate things for novice users of Gaffer. There was a bit of complexity around the change due to not be able to aggregation multiple properties together if they are stored in different places. For example - we will not allow groupBy and non-groupBy properties to be aggregated using the same BinaryOperator. 

See #1060 for a discussion around this feature and for the update Schema wiki page content.

Before we merge this in we should check that we are all happy with the extra "aggregator" section in the schema and that it doesn't complicate things for users.


p013570:
Closing whilst I make a few changes.

core/common-util/src/main/java/uk/gov/gchq/gaffer/commonutil/CollectionUtil.java
```diff
@@ -46,4 +47,36 @@ private CollectionUtil() {
 
         return treeSet;
     }
+
+    public static boolean containsAny(final Collection collection, final Object[] objects) {
```
t616178: Is it a good idea to write these ourselves? There are several other solutions from the libraries:

```
Collections.disjoint(collection, objects); // pure Java libraries
CollectionUtils.containsAny(collection, object); // Apache Commons
```

core/data/src/main/java/uk/gov/gchq/gaffer/data/element/function/ElementAggregator.java
```diff
@@ -57,6 +60,19 @@ public Properties apply(final Properties state, final Properties properties) {
...
     @Override
+    public List<TupleAdaptedBinaryOperator<String, ?>> getComponents() {
+        if (readOnly) {
+            return Collections.unmodifiableList(super.getComponents());
+        }
+
+        return super.getComponents();
+    }
+
+    public void lock() {
```
t616178: Do we need to be able to unlock as well?

core/store/src/main/java/uk/gov/gchq/gaffer/store/schema/SchemaElementDefinition.java
```diff
@@ -474,6 +583,20 @@ public CHILD_CLASS identifiers(final Map<IdentifierType, String> identifiers) {
...
         }
 
+        public CHILD_CLASS aggregator(final ElementAggregator aggregator) {
+            elDef.aggregator = aggregator;
+            return self();
+        }
+
+        @JsonSetter("aggregateFunctions")
+        public CHILD_CLASS aggregateFunctions(final List<TupleAdaptedBinaryOperator<String, Tuple<String>>> aggregateFunctions) {
+            if (null == getElementDef().aggregator) {
```
t616178: Is this direct access to the property definitely safe?

core/data/src/main/java/uk/gov/gchq/gaffer/data/element/function/ElementAggregator.java
p013570: No, I think once the ElementAggregator is locked - no changes should be allowed. If you need to make changes then you should create a new instance of ElementAggregator.

core/common-util/src/main/java/uk/gov/gchq/gaffer/commonutil/CollectionUtil.java
p013570: Those 2 methods takes collection, collection. We need to ones that works with an array. I could do Arrays.asList(array) and then use those methods but it seemed wasteful to create an extra object. What do you think?

t616178: That sounds reasonable - I just wanted to make sure we'd considered use of the libraries.

core/data/src/main/java/uk/gov/gchq/gaffer/data/element/function/ElementAggregator.java
t616178: Fine.

PR #1092, opened 2017-07-25
Gh 1089 graph hooks

p013570:


core/graph/src/main/java/uk/gov/gchq/gaffer/graph/hook/OperationAuthoriser.java
```diff
@@ -110,40 +65,53 @@ public void preExecute(final OperationChain<?> opChain, final User user) {
...
 
-    public Set<String> getAllOpAuths() {
-        return Collections.unmodifiableSet(allOpAuths);
+    public Map<Class<?>, Set<String>> getAuths() {
+        return Collections.unmodifiableMap(auths);
+    }
+
+    public void setAuths(final Map<Class<?>, Set<String>> auths) {
+        this.auths.clear();
+        this.auths.putAll(auths);
```
t616178: Elsewhere we are carrying out null checks before calling Collection.putAll() - should we be doing that everywhere?

core/graph/src/test/java/uk/gov/gchq/gaffer/graph/hook/GraphHookTest.java
```diff
@@ -0,0 +1,59 @@
...
+
+package uk.gov.gchq.gaffer.graph.hook;
+
+import org.junit.Test;
+import uk.gov.gchq.gaffer.commonutil.StreamUtil;
+import uk.gov.gchq.gaffer.exception.SerialisationException;
+import uk.gov.gchq.gaffer.jsonserialisation.JSONSerialiser;
+
+
+public abstract class GraphHookTest<T> {
```
t616178: Should we take a look at the hierarchy of test classes we're using throughout Gaffer and come up with something which is reusable in different modules? For example, the ```GraphHookTest.shouldJsonSerialiseAndDeserialise()``` test method should appear everywhere that we test a class (or type of class) which should be serialisable. Is there any value in a hierarchy which features a ```SerialiseableTest<T>``` interface (possibly with a default implementation)?

core/store/src/main/java/uk/gov/gchq/gaffer/store/operation/handler/ScoreOperationChainHandler.java
```diff
@@ -145,124 +112,55 @@ protected int authorise(final Operation operation) {
...
-        loadMaps(Paths.get(operationScoresFileName), Paths.get(authScoresFileName));
+    public Map<Class<? extends Operation>, Integer> getOpScores() {
+        return Collections.unmodifiableMap(opScores);
     }
 
-    private void loadMaps(final Path operationScores, final Path authScores) {
-        loadMaps(readEntries(operationScores), readEntries(authScores));
+    public void setOpScores(final LinkedHashMap<Class<? extends Operation>, Integer> opScores) {
+        this.opScores.clear();
+        this.opScores.putAll(opScores);
```
t616178: Same comment as earlier regarding null checks on method parameters.

core/graph/src/test/java/uk/gov/gchq/gaffer/graph/hook/GraphHookTest.java
p013570: Good idea. I've created a separate issue to do it: #1098.

Issue #923, opened 2017-05-23
Create an Apache Parquet backed store


ac74475:
It would be useful to have a store that scales well and is suitable for carrying out full table scans.
Apache Parquet seems to be a good candidate as Spark has been optimised for high performance reads of Parquet data. By sorting the data before writing, Parquet can be optimised for random access as well as full scans.

The initial version should include:
- Support for ingest aggregation
- Support for seeded queries with pre-aggregation filters that naturally fit to Parquet filters
- Store the data in an optimised way to make best use of the Parquet file format
- Expose the data as a Spark Dataframe

The Initial version will not be suitable for:
- Streaming data into
- Query time transformations
- Query time aggregation
- Fine grain security

Typical use cases could include:
- For applying a Page Rank algorithm against yesterday's data
- As part of a larger system where you stream data into a different store to act as your days cache, while still query able and then overnight load that days data into a Parquet store where it is available for full table scan queries as well as random access queries.

p013570:
Merged into develop.

PR #1094, opened 2017-07-26
gh-923 - tidying up parquet store

p013570:
#- Reordered methods in ParquetStore
- Changed some logging from info to debug
- Changed logging to use parameters
- Removed isValidationRequired method from Store. It was never used.
- Added implementation of Store.doUnhandledOperation to Store to avoid duplication as all stores implemented it in the same way.


ac74475:
Renaming the ParquetStoreProperties to ParquetProperties (for consistency with other stores) may cause confusion between the Gaffer ParquetProperties and the Parquet ParquetProperties. Therefore I think we should leave this as it is.

Issue #1098, opened 2017-07-26
Create an abstract class 'JsonSerialiseableTest' to be used by all classes that should be json serialisable


p013570:
This abstract class should have an abstract method:

```java
@Test
public abstract void shouldJsonSerialiseAndDeserialise();
```

We have several existing abstract classes that contain this method - these should be updated to extend the JsonSerialiseableTest class.

It would be useful to have some sort of default implementation if possible.

m607123:
# Migration
For any class that should JSON serialise/deserialise, their respective test class/superclass will now need to extend "JSONSerialisationTest", implementing the "getTestObject()" method to simply return an instance of the class under test.
Instances of empty/simple classes under test are now handled by the JSONSerialisationTest, and as such only custom/complex implementations are required in your test class.

p013570:
Merged into develop-1.0.0.

PR #1120, opened 2017-07-27
gh-1098 - Added JSON Serialisation Abstract Test

m607123:
Created default implementation for JSON serialisation.
Due to type erasure, couldn't manage to keep the Deserialisation from a path, but a default to and from method is provided as a minimum.
Abstract method just needs to return a new instance of the class under test.


core/serialisation/src/test/java/uk/gov/gchq/gaffer/JSONSerialisationTest.java
```diff
@@ -0,0 +1,58 @@
...
+
+import org.junit.Test;
+import uk.gov.gchq.gaffer.exception.SerialisationException;
+import uk.gov.gchq.gaffer.jsonserialisation.JSONSerialiser;
+
+import static org.junit.Assert.assertNotNull;
+
+public abstract class JSONSerialisationTest<T> {
+
+    JSONSerialiser SERIALISER = new JSONSerialiser();
```
p013570: this could be 'protected static final'

PR #1145, opened 2017-08-01
Gh 1085 lock down graph export

m55624:


core/graph/src/main/java/uk/gov/gchq/gaffer/operation/export/graph/handler/ExportToOtherAuthorisedGraphHandler.java
```diff
@@ -0,0 +1,183 @@
...
+        final ValidationResult result = new ValidationResult();
+
+        if (store.getGraphId().equals(graphId)) {
+            result.addError("Cannot export to the same Graph: " + graphId);
+        }
+        if (null == graphLibrary) {
+            // GraphLibrary is required as only a graphId, a parentStorePropertiesId or a parentSchemaId can be given
+            result.addError("Store GraphLibrary is null");
+        }
+        if (graphLibrary.exists(graphId)) {
```
p013570: This should be "else if" as if graphLibrary is null then currently you would get a null pointer exception here.

core/graph/src/main/java/uk/gov/gchq/gaffer/operation/export/graph/handler/ExportToOtherAuthorisedGraphHandler.java
```diff
@@ -0,0 +1,183 @@
...
+    protected Graph createGraph(final ExportToOtherAuthorisedGraph export, final Context context, final Store store) {
+        final String exportGraphId = export.getGraphId();
+        final String exportParentSchemaId = export.getParentSchemaId();
+        final String exportParentStorePropertiesId = export.getParentStorePropertiesId();
+        final GraphLibrary graphLibrary = store.getGraphLibrary();
+        final Graph graph;
+
+        validate(exportGraphId, exportParentSchemaId, exportParentStorePropertiesId, graphLibrary, store);
+
+        if (!isAuthorised(context.getUser(), idAuths.get(exportGraphId))) {
```
p013570: We need to do all the authorisation first. If a user is not allow to export to graph1 then we should not attempt to load graph1 from the graphLibrary. This also applied to checking the user is authorised to use the schema ids and properties id.

core/graph/src/main/java/uk/gov/gchq/gaffer/operation/export/graph/ExportToOtherAuthorisedGraph.java
```diff
@@ -0,0 +1,110 @@
...
+
+public class ExportToOtherAuthorisedGraph implements
+        Operation,
+        MultiInput<Element>,
+        ExportTo<Iterable<? extends Element>> {
+
+    @Required
+    private String graphId;
+    private Iterable<? extends Element> input;
+    private String parentSchemaId;
```
p013570: Please can we make this a list of parent schema ids like we have in ExportToOtherGraph.

core/graph/src/main/resources/ExportToOtherAuthorisedGraphOperationDeclarations.json
```diff
@@ -0,0 +1,24 @@
+{
```
p013570: This should not be in src/main

It is only example configuration, users will need to create their own version of this file with their own authorisations.

doc/src/main/java/uk/gov/gchq/gaffer/doc/operation/ExportToOtherAuthorisedGraphExample.java
```diff
@@ -0,0 +1,89 @@
...
+
+public class ExportToOtherAuthorisedGraphExample extends OperationExample {
+    public static void main(final String[] args) throws OperationException {
+        new ExportToOtherAuthorisedGraphExample().run();
+    }
+
+    public ExportToOtherAuthorisedGraphExample() {
+        super(ExportToOtherAuthorisedGraph.class, "These export examples export all edges in the example graph to another Gaffer instance using Operation Auths against the user. \n\n" +
+                "To add this operation to your Gaffer graph you will need to include the ExportToOtherAuthorisedGraphOperationDeclarations.json in your store properties, i.e. set this property: " +
+                "gaffer.store.operation.declarations=ExportToOtherAuthorisedGraphOperationDeclarations.json\n");
```
p013570: Users will need to write their own json file first.

rest-api/core-rest/src/main/java/uk/gov/gchq/gaffer/rest/SystemProperty.java
```diff
@@ -26,6 +26,7 @@
  */
 public abstract class SystemProperty {
     // KEYS
+    public static final String GRAPH_LIBRARY = "gaffer.graph.library.path";
```
p013570: This should be:

public static final String GRAPH_LIBRARY_CLASS = "gaffer.graph.library.class";
public static final String GRAPH_LIBRARY_CONFIG = "gaffer.graph.library.config";

Then users can choose what graph library they want to use.
We could then load it using something like:

```java
String graphLibraryClassName = System.getProperty(SystemProperty.GRAPH_LIBRARY_CLASS);
if(null != graphLibraryClassName) {
    GraphLibrary library = null;
    try {
        library = Class.forName(graphLibraryClassName).asSubClass(GraphLibrary.class).newInstance();
    } catch (ClassNotFoundException e) {
        throw new RuntimeException("Unknown GraphLibrary class: + " + graphLibraryClassName);
    }
    library.initialise(System.getProperty(SystemProperty.GRAPH_LIBRARY_CONFIG));
}
```

rest-api/core-rest/src/main/java/uk/gov/gchq/gaffer/rest/factory/DefaultGraphFactory.java
```diff
@@ -134,6 +135,10 @@ public void setSingletonGraph(final boolean singletonGraph) {
         builder.storeProperties(storeProperties);
         builder.graphId(getGraphId());
 
+        if (System.getProperty(SystemProperty.GRAPH_LIBRARY) != null) {
```
p013570: See comment above.

example/road-traffic/road-traffic-demo/src/main/resources/graphLibrary/roadTrafficGraphs.json
```diff
@@ -0,0 +1 @@
+roadTraffic,roadTraffic
```
p013570: I don't think the graphLibrary directory should be committed.

example/road-traffic/road-traffic-demo/pom.xml
```diff
@@ -68,8 +68,12 @@
                                 </webapp>
                             </webapps>
                             <systemProperties>
-                                <gaffer.rest-api.basePath>${standalone-path}/v1
+                                <gaffer.rest-api.basePath>
+                                    ${standalone-path}/v1
                                 </gaffer.rest-api.basePath>
+                                <gaffer.graph.library.path>
+                                    ${project.parent.basedir}/road-traffic-demo/src/main/resources/graphLibrary
```
p013570: the path should be ${project.build.outputDirectory}/graphLibrary

core/store/src/main/java/uk/gov/gchq/gaffer/store/library/HashMapGraphLibrary.java
```diff
@@ -27,6 +27,14 @@
     private static final Map<String, byte[]> SCHEMAS = new HashMap<>();
     private static final Map<String, StoreProperties> PROPERTIES = new HashMap<>();
 
+    public HashMapGraphLibrary() {
+
+    }
+
+    public void initialise(final String file) {
+        // Do nothing, only used with FileGraphLibrary
```
p013570: Don't reference the FileGraphLibrary here as there could be other libraries in the future. Change the comment to:
// Do nothing.

core/store/src/main/java/uk/gov/gchq/gaffer/store/library/NoGraphLibrary.java
```diff
@@ -29,6 +29,10 @@ public NoGraphLibrary() {
         LOGGER.debug("Your schema will not be stored in a graph library. So you will need to provide it each time you create an instance of Graph.");
     }
 
+    public void initialise(final String path) {
+        // Do nothing, only used with FileGraphLibrary
```
p013570: Change to just:
// Do nothing.

example/road-traffic/road-traffic-demo/pom.xml
```diff
@@ -72,7 +72,7 @@
                                     ${standalone-path}/v1
                                 </gaffer.rest-api.basePath>
                                 <gaffer.graph.library.path>
```
p013570: This system property needs updating to use the new gaffer.graph.library.class and gaffer.graph.library.config properties.

Issue #1144, opened 2017-08-01
Sort operation will skip elements when comparator returns 0


p013570:
The Sort operation uses a TreeSet to sort the results. The TreeSet implementation performs all element comparisons using the comparator, so two elements that are deemed equal by the comparator will be deduplicated.

For example if you use the Sort operation to sort the the following 2 entities based on the value of the count property:

```java
final Entity entity1 = new Entity.Builder()
        .group(TestGroups.ENTITY)
        .vertex("1")
        .property("count", 1)
        .build();
final Entity entity2 = new Entity.Builder()
        .group(TestGroups.ENTITY)
        .vertex("2")
        .property("count", 1)
        .build();
```
then you would just get back 1 of the entity as they are deemed to be equal and are deduplicated.

We recommend not using the Sort operation until it has been fixed. In the meantime there are Max and a Min Operations that you could use.

You can easily disable the Sort operation by creating a file (DisableSortOperation.json) containing the following:

```json
{
  "operations": [
    {
      "operation": "uk.gov.gchq.gaffer.operation.impl.compare.Sort",
      "handler": null
    }
  ]
}
```

then added it to your store properties:
```properties
gaffer.store.operation.declarations=DisableSortOperation.json
```

p013570:
Merged into develop.

PR #1146, opened 2017-08-01
gh-1144 - changed LimitedSortedSet to LimitedSortedInMemoryIterable

p013570:


core/common-util/src/main/java/uk/gov/gchq/gaffer/commonutil/collection/LimitedSortedList.java
```diff
@@ -0,0 +1,300 @@
...
+    // Dummy value to associate with an Object in the backing Map
+    private static final Object PRESENT = new Object();
+
+    private final Comparator<E> comparator;
+    private final boolean deduplicate;
+    private final Integer limit;
+    private LinkedList<E> backingList;
+
+    // Used to deduplicate elements if required
+    private HashMap<E, Object> backingHashSet;
```
t616178: Is it inefficient to use both backing collections?

core/common-util/src/main/java/uk/gov/gchq/gaffer/commonutil/collection/LimitedSortedList.java
```diff
@@ -0,0 +1,300 @@
...
+    }
+
+    public LimitedSortedList(final Comparator<E> comparator, final Integer limit, final boolean deduplicate, final LinkedList<E> backingList) {
+        if (null == comparator) {
+            throw new IllegalArgumentException("Comparator is required");
+        }
+        if (null != limit && limit < 1) {
+            throw new IllegalArgumentException("Limit should be more than 0");
+        }
+        if (null == backingList) {
```
t616178: Would we ever want to expose this constructor to allow users to specify their own backing list? Could we make this private?

core/common-util/src/main/java/uk/gov/gchq/gaffer/commonutil/collection/LimitedSortedList.java
```diff
@@ -0,0 +1,300 @@
...
+        }
+        return backingList.containsAll(c);
+    }
+
+    @Override
+    public boolean addAll(final Collection<? extends E> c) {
+        boolean result = false;
+        for (final E e : c) {
+            if (add(e)) {
+                result = true;
```
t616178: Do we want to keep a track of the result like this in the removeAll method too? (and any other method which is expected to report whether an operation on the collection was successful)

Issue #1103, opened 2017-07-26
Improve how the Parquet Index is generated


ac74475:
Currently the index is generated by taking the min and max statistics for each column related to the Gaffer identifiers. This causes a problem when the identifier is split into more than 1 column as the second column's last value may not be the max value. Really this needs to get the first and last row in each file and record the value for each column related to the Gaffer identifier.
This would then allow for the accurate targeting of filters to files when the identifier is split over more than one column.

p013570:
Merged into develop.

PR #1159, opened 2017-08-03
gh-1103 Fixed issue with index

ac74475:
This commit will dramatically improve seeded queries on graphs where
the vertex type is serialised into multiple columns


ac74475:
Updated the indexing to only store the Parquet objects of the first row which correspond to the indexed column. This removes the need to either run a Spark reduce job over the whole file or having to iterate over all the index for each seed and potentially still send a filter to a file and get no results.

store-implementation/parquet-store/src/main/java/uk/gov/gchq/gaffer/parquetstore/index/MinValuesWithPath.java
```diff
@@ -25,26 +25,15 @@
  * for a specific gaffer column, i.e. if the vertex was a {@link uk.gov.gchq.gaffer.types.TypeValue} object then the Min would consist of two Strings,
  * the first representing the type and the second representing the value.
  */
-public class MinMaxPath {
+public class MinValuesWithPath {
```
p013570: Javadoc needs updating. It still refers to 'Max'.

store-implementation/parquet-store/src/main/java/uk/gov/gchq/gaffer/parquetstore/operation/addelements/impl/ReduceRowsToGetLastRow.java
```diff
@@ -0,0 +1,45 @@
+/*
```
gaffer01: This class isn't used anywhere.

Issue #1151, opened 2017-08-02
Shaded jar for time-library


gaffer01:
It would be useful if the pom of time-library specified a shaded jar. This jar could then be deployed to the tablet servers of an Accumulo store to allow properties from this library to be used. The pom should include:

    org.roaringbitmap:RoaringBitmap
    uk.gov.gchq.gaffer:bitmap-library
    uk.gov.gchq.gaffer:sketches-library

p013570:
Merged into develop.

PR #1163, opened 2017-08-08
gh-1151 - Specified shaded jar for time-library

m607123:
Added the roaringbitmap dependency to the time-library pom - followed the same style as in another pom so may need tweaking.


m607123:
Think the build may fail, but I believe the reasons as to why have been addressed in another branch (Parquet test unordering, and a checkstyle issue in SchemaElementDefinition)

library/time-library/pom.xml
```diff
@@ -26,6 +26,11 @@
 
     <dependencies>
         <dependency>
+            <groupId>org.roaringbitmap</groupId>
```
p013570: Do  you need to add this dependency, it should be included in the bitmap-library dependency.

library/time-library/pom.xml
```diff
@@ -51,4 +56,33 @@
...
+                        <goals>
+                            <goal>shade</goal>
+                        </goals>
+                        <configuration>
+                            <shadedArtifactAttached>true
+                            </shadedArtifactAttached>
+                            <shadedClassifierName>deploy
+                            </shadedClassifierName>
+                            <artifactSet>
+                                <includes>
```
p013570: I think there are a few more things you need to include here (as specified in the issue):

org.roaringbitmap:RoaringBitmap
uk.gov.gchq.gaffer:bitmap-library
uk.gov.gchq.gaffer:sketches-library

Once the shaded jar is compiled, please could you look at what is included and check it contains all the required dependencies for this time library.

Issue #1123, opened 2017-07-28
Move clearspring serialisers, aggregators to their own subpackage


gaffer01:
The serialisers and aggregators for classes from the data sketches library are in a `datasketches` subpackage. We should create a `clearspring` subpackage for serialisers and aggregators for sketches from the clearspring library and move the existing HyperLogLog serialiser and aggregator into there.

m607123:
# Migration

HyperLogLogPlus aggregator and serialiser are now in the subpackage:
uk.gov.gchq.gaffer.sketches.clearspring.cardinality

Any references should in future be updated accordingly.

t616178:
Merged into develop-1.0.0.

PR #1169, opened 2017-08-09
gh-1123 - Moved HLLP to new subpackage

m607123:
Moved the hyperloglogplus serialiser and aggregator to a new subpackage, likewise for their respective test classes.


m607123:
@gaffer01 Just to confirm - was this the correct subpackage layout you intended?

gaffer01:
Almost - I think it would be better for the subpackage to be called `cardinality` rather than `hyperloglogplus`, i.e. `HyperLogLogPlusSerialiser` should be in the package `uk.gov.gchq.gaffer.sketches.clearspring.cardinality.serialisation` rather than `uk.gov.gchq.gaffer.sketches.clearspring.hyperloglogplus.serialisation`. This is more consistent with the naming for the sketches under `datasketches` and it's more consistent with the internal structure of the clearspring library which has `HyperLogLogPlus` in a `cardinality` package.

library/sketches-library/src/main/java/uk/gov/gchq/gaffer/sketches/clearspring/cardinality/serialisation/json/hyperloglogplus/HyperLogLogPlusJsonDeserialiser.java
```diff
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-package uk.gov.gchq.gaffer.sketches.serialisation.json.hyperloglogplus;
+package uk.gov.gchq.gaffer.sketches.clearspring.cardinality.serialisation.json.hyperloglogplus;
```
p013570: I don't think we need the hyperloglogplus at the end of the package name.

PR #1170, opened 2017-08-09
Gh 1165 graph config

p013570:


core/graph/src/main/java/uk/gov/gchq/gaffer/graph/Graph.java
```diff
@@ -305,59 +305,131 @@ public StoreProperties getStoreProperties() {
...
-        private String graphId;
-        private GraphLibrary library;
         private StoreProperties properties;
         private Schema schema;
-        private View view;
-        private List<GraphHook> graphHooks = new ArrayList<>();
         private String[] parentSchemaIds;
         private String parentStorePropertiesId;
 
         public Builder graphId(final String graphId) {
```
t616178: Does this method need to be deprecated too?

core/graph/src/main/java/uk/gov/gchq/gaffer/graph/GraphConfig.java
```diff
@@ -0,0 +1,267 @@
...
+            return this;
+        }
+
+        public Builder addHooks(final GraphHook... graphHooks) {
+            Collections.addAll(this.config.getHooks(), graphHooks);
+            return this;
+        }
+
+        public GraphConfig build() {
+            if (null == config.getLibrary()) {
```
t616178: What happens if graphId is null?

p013570: The graph will deal with that - it will attempt to get it from a provided store and if it is still null it will throw an exception.

core/graph/src/main/java/uk/gov/gchq/gaffer/graph/Graph.java
p013570: I have now deprecated that method too.

core/graph/src/main/java/uk/gov/gchq/gaffer/graph/GraphConfig.java
```diff
@@ -0,0 +1,267 @@
...
+import java.io.IOException;
+import java.io.InputStream;
+import java.net.URI;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
+public final class GraphConfig {
```
t616178: Does this class need some Javadoc?

p013570: Thanks, I've added some now.

core/graph/src/main/java/uk/gov/gchq/gaffer/graph/GraphConfig.java
```diff
@@ -91,36 +100,6 @@ public void setHooks(final List<GraphHook> hooks) {
     }
 
     @Override
-    public boolean equals(final Object obj) {
```
t616178: Was there a reason for removing the equals() and hashCode() methods?

p013570: View, GraphHooks and GraphLibrary don't implement equals so having a GraphConfig.equals method doesn't really work.

PR #1187, opened 2017-08-11
Gh 1137 override shallow clone

m55624:


core/graph/src/main/java/uk/gov/gchq/gaffer/operation/export/graph/ExportToOtherAuthorisedGraph.java
```diff
@@ -88,6 +88,16 @@ public void setKey(final String key) {
         return (TypeReference) new TypeReferenceImpl.Object();
     }
 
+    @Override
+    public Operation shallowClone() {
```
p013570: Please can you update the return type to be ExportToOtherAuthorisedGraph. Similarly with the other operations.

core/graph/src/test/java/uk/gov/gchq/gaffer/operation/export/graph/ExportToOtherAuthorisedGraphTest.java
```diff
@@ -55,6 +55,17 @@ public void builderShouldCreatePopulatedOperation() {
     }
 
     @Override
+    public void shouldShallowCloneOperation() {
+        // When
+        ExportToOtherAuthorisedGraph clone = (ExportToOtherAuthorisedGraph) op.shallowClone();
+
+        // Then
+        assertEquals("graphId", clone.getGraphId());
```
p013570: For all these tests please can you also add:

assertNotSame(op, clone);

This will check that the instance is actually different.

PR #1208, opened 2017-08-21
Gh 1177 add description to graph config

m55624:


rest-api/core-rest/src/main/java/uk/gov/gchq/gaffer/rest/service/IGraphConfigurationService.java
```diff
@@ -43,6 +43,11 @@
     Schema getSchema();
 
     @GET
+    @Path("/description")
+    @ApiOperation(value = "Gets the Graph description", response = String.class, responseContainer = "list")
```
t616178: There shouldn't be a responseContainer parameter here, since this operation returns a single value and not a list.

core/graph/src/main/java/uk/gov/gchq/gaffer/graph/GraphConfig.java
```diff
@@ -85,6 +86,14 @@ public void setLibrary(final GraphLibrary library) {
         this.library = library;
     }
 
+    public String getDescription() {
+        return description;
+    }
+
+    public void setDescription(final String description) {
+        this.description = description;
```
t616178: Do we need to enforce any restrictions on the characters that are allowed in the description field?

m55624: I think we will leave any restrictions on the description currently as it isn't 'used' anywhere.  We can look into updating this in the future though if we need to.

PR #1209, opened 2017-08-21
Gh 1154 getelements builder error

m607123:
Added null check to the GetElementsHandler.
Updated existing tests to account for this.
Added new tests to check the new validation.


core/store/src/test/java/uk/gov/gchq/gaffer/store/operation/OperationChainValidatorTest.java
```diff
@@ -54,7 +54,7 @@ public void shouldValidateValidOperationChain() {
                 new GetElements(),
                 new ToVertices(),
                 new GetAdjacentIds()
-        )), true);
```
t616178: Could you add in a test which results in a valid OperationChain being successfully validated?

Issue #1096, opened 2017-07-26
Upgrade Spark version to 2.2.0


ac74475:
This will update the Parquet version to 1.8.2 which will require a few changes to the ParquetStore.

t616178:
Merged into develop.

p013570:
# Migration Steps
Spark has been updated to 2.2.0.

PR #1210, opened 2017-08-22
gh-1096 - spark upgraded to 2.2

m55624:


ac74475:
Can you update the Parquet Store readme to say that it is now using Parquet version 1.8.2

ac74475:
Also the NOTICES file will need updating

PR #1214, opened 2017-08-24
Gh 1213 AddElementsFromHdfs directory check

m607123:
The directory check is now executed before any computation or MapReduce functions are performed - this saves time when a job would fail due to a directory issue, as it will now not waste time Reducing data to simply throw it away.

The check has been moved from the tool to the handler; the configuration constructed as an instance variable and exposed by a getter; tests modified to expect the correct exception.


store-implementation/accumulo-store/src/main/java/uk/gov/gchq/gaffer/accumulostore/operation/hdfs/handler/AddElementsFromHdfsHandler.java
```diff
@@ -68,6 +68,12 @@ public void doOperation(final AddElementsFromHdfs operation,
             operation.setSplitsFilePath(splitsFilePath);
         }
 
+        try {
+            checkHdfsDirectories(operation, store);
+        } catch (IOException e) {
```
t616178: Missing final keyword.

store-implementation/accumulo-store/src/main/java/uk/gov/gchq/gaffer/accumulostore/operation/hdfs/handler/AddElementsFromHdfsHandler.java
```diff
@@ -222,4 +228,22 @@ private void importElements(final AddElementsFromHdfs operation, final AccumuloS
...
+
+    private void checkHdfsDirectories(final AddElementsFromHdfs operation, final AccumuloStore store) throws IOException {
+        final AddElementsFromHdfsTool tool = new AddElementsFromHdfsTool(new AccumuloAddElementsFromHdfsJobFactory(), operation, store);
+
+        LOGGER.info("Checking that the correct HDFS directories exist");
+        final FileSystem fs = FileSystem.get(tool.getConfig());
+
+        final Path outputPath = new Path(operation.getOutputPath());
+        LOGGER.info("Ensuring output directory {} doesn't exist", outputPath);
+        if (fs.exists(outputPath)) {
```
t616178: Does this still result in the correct behaviour if the output path does not exist in the filesystem?

m607123: I've added a test that should hopefully cover this - seems to report that this is working as intended however.

Issue #1097, opened 2017-07-26
Improve the serialisation of Gaffer Objects to improve performance with Spark


ac74475:
Register all the core/type/types classes into the uk/gov/gchq/gaffer/spark/serialisation/kryo/Registrator class, which will speed up the serialisation of those classes when performing Spark operations.

t616178:
Merged into develop

PR #1225, opened 2017-08-30
gh-1097 - Added and Registered core/type/types Kryo Serializers

m607123:
For each class in core/type/types, a Kryo serializer was created and registered in the Registrator.
IntegerFreqMap has been omitted due to its deprecated status.


m607123:
Closing while tests added

library/spark/spark-library/src/test/java/uk/gov/gchq/gaffer/spark/serialisation/kryo/impl/EdgeKryoSerializerTest.java
```diff
@@ -36,7 +36,7 @@ public void setup() {
     }
 
     @Test
-    public void testEdge() {
+    public void testEdgeKryoSerializer() {
```
t616178: These test methods should begin with the keyword "should" and should describe the expected behaviour of the test, for example:

```java
@Test
public void shouldSerialiseDeserialiseEdge() {
    // Test logic here
}
```

library/spark/spark-library/src/test/java/uk/gov/gchq/gaffer/spark/serialisation/kryo/impl/FreqMapKryoSerializerTest.java
```diff
@@ -0,0 +1,55 @@
+/*
```
t616178: Is there an opportunity here to create a KryoSerializerTest which is extended by the other test classes? The implementation details in each of the test methods are very similar.

PR #1228, opened 2017-08-31
Gh 489 remove deprecated classes

t616178:


rest-api/core-rest/src/main/java/uk/gov/gchq/gaffer/rest/factory/DefaultGraphFactory.java
```diff
@@ -104,37 +108,12 @@ public void setSingletonGraph(final boolean singletonGraph) {
...
-        if (null != graphConfigPath) {
-            builder.config(Paths.get(graphConfigPath));
-        }
+        builder.config(getConfigPath());
 
         for (final Path path : getSchemaPaths()) {
             builder.addSchema(path);
         }
 
-        final String graphId = System.getProperty(SystemProperty.GRAPH_ID);
```
p013570: Might be worth leaving this bit in for a bit longer as it was only recently deprecated.

Issue #1188, opened 2017-08-11
Add GraphLibrary to FederatedStore


GCHQDeveloper404:

p013570:
Merged into develop.

PR #1240, opened 2017-09-05
gh-1188 - Graphs filtered by Operation Options

m607123:
Operation now extends Options
The contents of Options.Builder moved to Operation.Builder
Modified any resulting hierarchical/inheritance issues

FederatedOperationHandler now returns only graphs specified in the Options of the Operation if defined, otherwise returns all graphs as before


store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/FederatedStore.java
```diff
@@ -456,4 +456,18 @@ protected Object doUnhandledOperation(final Operation operation,
                                           final Context context) {
         throw new UnsupportedOperationException();
     }
+
+    public Collection<Graph> getGraphs(final String[] split) {
```
p013570: Please can you rename split to graphIds.

Issue #1267, opened 2017-09-08
Improve the example button in the REST api swagger ui


p013570:
With the new rest api version the example button for the operations isn't as useful as it was in the previous rest api version. As there is only 1 operation end point it is more difficult to get examples of the different operations.

For v2, we could add a select/dropdown next to the example button to choose the operation you wish to get the json for.

p013570:
This new feature will be helpful when writing operation chains in json.

To use it
- go the the rest api swagger ui
- go to operations/execute
- in the dropdown select an operation
- click add example
- select another operation
- click add example again

This will add 2 examples in an operation chain.

p013570:
Merged into develop.

PR #1268, opened 2017-09-08
gh-1267 - updated example button in swagger

p013570:
To test this:
- start the example rest api
- go to operations/execute
- in the dropdown select an operation
- click add example
- select another operation
- click add example again

This will add 2 examples in an operation chain.


t616178:
This is great - just a few general points:

* Is there any reason behind the ordering of operations in the dropdown list? Alphabetical ordering or placing subheadings in the dropdown might be better.
* Some of the operation examples might need better generation logic if they are to be used without any additional input. For example, trying to add a ```Sort``` operation will result in an error because no value is provided for the comparators field.

p013570:
Thanks @t616178. I've updated it to sort the operations. I've created another issue to improve the logic used to generate the examples - #1281

Issue #1275, opened 2017-09-12
GetAdjacentIds with a view containing entity filters is invalid


p013570:
We should validate the GetAdjacentIds operations and if the view contains any entities with filters then we should throw an exception.

When an GetAdjacentIds operation is handled it does not load any entities so no entity filters can be applied.

m607123:
There only seems to be a test for the MapStore implementation of the GetAdjacentIds, @p013570 would you like similar tests writing for the other Store implementations as well?

m607123:
# Migration
When building a GetAdjacentIds operation, please ensure that entities **do not have any**:
- preAggregationFilter
- postAggregationFilter
- postTransformFilter

attached to them whatsoever, else the operation will fail validation and throw an IllegalArgumentException.

p013570:
Merged into develop.

PR #1277, opened 2017-09-12
Gh-1275 GetAdjacentIds Validation

m607123:
If the View of an Operation contains any entities with aggregation or transform filters, an exception will now be thrown.

Tests updated to reflect changes.

### DIFF
--- a/core/data/src/main/java/uk/gov/gchq/gaffer/data/elementdefinition/view/View.java
+++ b/core/data/src/main/java/uk/gov/gchq/gaffer/data/elementdefinition/view/View.java
@@ -25,7 +25,6 @@ import org.apache.commons.lang3.builder.HashCodeBuilder;
 
 import uk.gov.gchq.gaffer.commonutil.CommonConstants;
 import uk.gov.gchq.gaffer.commonutil.ToStringBuilder;
-import uk.gov.gchq.gaffer.commonutil.iterable.ChainedIterable;
 import uk.gov.gchq.gaffer.data.elementdefinition.ElementDefinitions;
 import uk.gov.gchq.gaffer.data.elementdefinition.exception.SchemaException;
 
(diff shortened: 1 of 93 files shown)


core/data/src/main/java/uk/gov/gchq/gaffer/data/elementdefinition/view/View.java
```diff
@@ -132,6 +131,18 @@ public boolean hasPostTransformFilters() {
         return hasFilters(ViewElementDefinition::hasPostTransformFilters);
     }
 
+    public boolean hasAnyEntityFilter() {
```
p013570: Perhaps rename this to 'hasEntityFilters'

core/data/src/main/java/uk/gov/gchq/gaffer/data/elementdefinition/view/View.java
```diff
@@ -132,6 +131,18 @@ public boolean hasPostTransformFilters() {
...
+        return hasEntityFilters(ViewElementDefinition::hasPostAggregationFilters)
+                || hasEntityFilters(ViewElementDefinition::hasPostTransformFilters)
+                || hasEntityFilters(ViewElementDefinition::hasPreAggregationFilters);
+    }
+
+    public boolean hasAnyEdgeFilter() {
```
p013570: Perhaps rename this to 'hasEdgeFilters'

store-implementation/map-store/src/test/java/uk/gov/gchq/gaffer/mapstore/impl/GetAdjacentIdsTest.java
```diff
@@ -280,4 +281,39 @@ public void testGetElementsByEntityIdWithViewRestrictedByGroupAndAPostAggregatio
         expectedResults.remove(new EntitySeed("Y2"));
         assertEquals(expectedResults, resultsSet);
     }
+
+    @Test
+    public void shouldFailValidationWhenEntityHasFilter() throws OperationException {
```
p013570: Please could you also write some unit tests to specifically check the validation doesn't fail when there are entities, but without filters.

Issue #1291, opened 2017-09-14
Create Filter operation


p013570:
Create a new Filter operation that takes an iterable of elements and ElementFilters. The fields should be similar to the View, with edges, entities, globalElements, globalEdges, globalEntities.

p013570:
Merged into develop.

p013570:
Please note, this operation will note be available be default until this bug is fixed: https://github.com/gchq/Gaffer/issues/1341

PR #1293, opened 2017-09-14
gh-1291 - Filter function added, FilterStream functionality added

m607123:
Filter function added, taking in an iterable of Elements and an ElementFilter, returning a filtered iterable of Elements (or the original iterable if the ElementFilter is null).

For utility purposes, a FilterStreamSupplier and StreamFitlerIterable were written to facilitate the new functionality. (with thanks to @t616178! )


m607123:
Closing while features changed/added

PR #1300, opened 2017-09-19
Gh 1289 update accumulo

m55624:
Before merging this in we need to check the gaffer-tools 'slider' implementation will still work.


store-implementation/accumulo-store/README.md
```diff
@@ -55,13 +55,13 @@ Gaffer's `AccumuloStore` is particularly well-suited to graphs where the propert
 Accumulo set up
 -----------------------------------------------
 
-Gaffer has been extensively tested with Accumulo version 1.7.2. It should work with any of the 1.7.* versions of Accumulo.
+Gaffer has been extensively tested with Accumulo version 1.8.1. It is recommended to use this version, although it should work with any of the 1.7.* versions of Accumulo as well.
```
p013570: This should say:

should work with any of the 1.8.* versions of Accumulo as well

Issue #1310, opened 2017-09-21
When schemas are merged groupBy and description are overridden


p013570:
Currently the logic for merging schemas always overrides the groupBy and description fields. This should be updated to only override if the new values have been set (not null).

This causes issues when loading a schema folder as the schema files may be loaded and merged in together different orders.

p013570:
Merged into develop.

PR #1311, opened 2017-09-21
gh-1310 - updating schema merging of groupBy and description

p013570:


core/store/src/main/java/uk/gov/gchq/gaffer/store/schema/SchemaElementDefinition.java
```diff
@@ -704,9 +704,18 @@ public CHILD_CLASS merge(final ELEMENT_DEF elementDef) {
             elDef.ingestAggregatorCache = null;
             elDef.queryAggregatorCacheMap.clear();
 
-            elDef.groupBy = new LinkedHashSet<>(elementDef.groupBy);
-            elDef.parents = null != elementDef.parents ? new LinkedHashSet<>(elementDef.parents) : null;
-            elDef.description = elementDef.description;
+            if (!elementDef.groupBy.isEmpty()) {
+                elDef.groupBy = new LinkedHashSet<>(elementDef.groupBy);
```
m55624: Can elementDef.groupBy be null? if so you will get a NPE thrown.  Might be worth adding the null != elementDef.groupBy check?

Issue #1303, opened 2017-09-19
Add support in the Parquet Store to natively handle the AgeOff predicate


p013570:
Without the native support of the AgeOff Predicate in the Parquet Store, all queries will need to be re-validated client side and it will make queries significantly slower.

gaffer01:
The way this should work is:

- Validation should be applied to all elements when new data is added (as all the data has to be read anyway) - this should be optional though in case your application knows for sure that no invalid data has been added.
- There should be a `RemoveInvalidElements` operation that reprocesses the entire graph and removes invalid elements.
- All queries against the Parquet store should have an _optional_ option to validate the data before it is returned. There are scenarios where you know there isn't any invalid data, e.g. you do a nightly update to the graph which adds new data and removes old data and you then know that all queries that run in the next 24 hours won't see any invalid data (because all the data in the graph is now valid for the next 24 hours). In this case you want to avoid the expense of validation (the additional expense of doing validation could be very high if you're getting a data frame and then only reading one column - doing validation may require reading all the columns, converting to a Java `Element` object and applying a predicate).

ac74475:
This looks like the same issue as #1100 to me. I agree with @gaffer01's explanation of how this should be implemented.

p013570:
Yes I agree with @gaffer01. I've added this comment to #1100:

Currently the Parquet Store has special implementations for primitive predicates for some of Gaffer/Koryphe's core predicates like IsMoreThan. See ParquetFilterUtils.addPrimitiveFilter. I believe this means the IsMoreThan predicate can be applied more efficiently. AgeOff is just another predicate that is quite simple and #1303 is to add specific logic in ParquetFilterUtils to deal with the AgeOff predicate efficiently in a similar way to IsMoreThan.

So this issue is to simply add a bit of logic to handle the AgeOff Predicate in a similar way to IsMoreThan in ParquetFilterUtils.

p013570:
This has been done as part of #1100.

PR #1320, opened 2017-09-21
Gh 1100 parquet age off

m55624:


gaffer01:
Before this is merged can someone comment on the question raised in the issue about how this relates to #1303 ?

store-implementation/parquet-store/src/main/java/uk/gov/gchq/gaffer/parquetstore/operation/getelements/impl/ParquetElementRetriever.java
```diff
@@ -188,12 +196,15 @@ public Element next() throws NoSuchElementException {
                 e = queue.poll();
                 if (null != e) {
                     if (needsValidation) {
```
ac74475: All validation and filters should really be applied by the RetrieveElementsFromFile class so that the functionality is computed by the threads per file rather than by a single thread.

ac74475:
The Parquet Store README will need updating to reflect these changes

store-implementation/parquet-store/src/main/java/uk/gov/gchq/gaffer/parquetstore/operation/getelements/impl/RetrieveElementsFromFile.java
```diff
@@ -87,23 +89,24 @@ public RetrieveElementsFromFile(final Path filePath, final FilterPredicate filte
...
     @Override
     public OperationException call() throws Exception {
+        if (null == elementFilter) {
+            elementFilter = new ViewElementDefinition.Builder().json(elementDefinitionJson).build().getPreAggregationFilter();
+        }
         try {
             final ParquetReader<Element> fileReader = openParquetReader();
             Element e = fileReader.read();
-            while (null != e) {
+            while (e != null) {
```
ac74475: We usually try to have the constant first so "null != e" would be prefered

PR #1323, opened 2017-09-21
Gh 1260 road example

t616178:


example/road-traffic/road-traffic-generators/src/main/java/uk/gov/gchq/gaffer/traffic/generator/RoadTrafficElementGenerator.java
```diff
@@ -0,0 +1,79 @@
...
+                .property("count", 1L)
+                .build();
+    }
+
+    protected Date getDate(final String dCountString, final String hour) {
+        Date dCount = null;
+        try {
+            dCount = new SimpleDateFormat("dd/MM/yyyy HH:mm").parse(dCountString);
+        } catch (final ParseException e) {
+            // incorrect date format
```
javadev001001: If we're seeing incorrect date formats, is it not worth logging this/reporting it to the user?  Otherwise it's going to look like data is loading, where actually it's being dropped because of the timestamp

t616178: Agreed that this is a potential issue - I'll add an exception here if it was not possible to parse one of the date strings.

Issue #1329, opened 2017-09-22
GetDataFrameOfElementsHandler doesn't pass through options


gaffer01:
The `GetDataFrameOfElementsHandler` in the spark-accumulo module doesn't pass the options on the operation on to the underlying `GetRDDOfAllElements` operation. This means that the Hadoop configuration cannot be passed into this operation.

If you're running the code with the `hadoop` command or within a Spark job configured with a Hadoop configuration then it will work, but you can't override any parts of the configuration.

gaffer01:
An initial attempt to unit test this using the following code in `GetRDDOfElementsHandlerTest` fails - the value set on the configuration isn't present in the configuration obtained from the RDD:

    @Test
    public void checkHadoopConfIsPassedThrough() throws OperationException, IOException {
        final Graph graph1 = new Graph.Builder()
                .config(new GraphConfig.Builder()
                        .graphId("graphId")
                        .build())
                .addSchema(getClass().getResourceAsStream("/schema/elements.json"))
                .addSchema(getClass().getResourceAsStream("/schema/types.json"))
                .addSchema(getClass().getResourceAsStream("/schema/serialisation.json"))
                .storeProperties(getClass().getResourceAsStream("/store.properties"))
                .build();
        final User user = new User();
        final SparkSession sparkSession = SparkSessionProvider.getSparkSession();
        final Configuration conf = new Configuration();
        conf.set("AN_OPTION", "A_VALUE");
        final String encodedConf = AbstractGetRDDHandler.convertConfigurationToString(conf);
        final GetRDDOfElements rddQuery = new GetRDDOfElements.Builder()
                .input(new EdgeSeed("1", "B", false))
                .sparkSession(sparkSession)
                .option(AbstractGetRDDHandler.HADOOP_CONFIGURATION_KEY, encodedConf)
                .build();
        final RDD<Element> rdd = graph1.execute(rddQuery, user);

        assertEquals("A_VALUE", rdd.sparkContext().hadoopConfiguration().get("AN_OPTION"));
    }

p013570:
Merged into develop.

PR #1334, opened 2017-09-25
gh-1329 - all options now passed through to underlying Operations

m55624:


gaffer01:
I think this also needs fixing in `GetRDDOfAllElementsHandler`. And the issue also asks for the `GetDataFrameOfElementsHandler` class to pass the option containing the encoded `Configuration` on to the operations that return the RDD.

Issue #1346, opened 2017-09-27
Filter, Transform, and Aggregate operations should be validated


m607123:
When the above operations are used, it is difficult to discern any potential issues with the query, so adding validation to each of them would aid utilisation of them. The validation should be similar to the validation done for an operation View.

p013570:
Merged into develop.

PR #1361, opened 2017-10-02
Gh-1346 - Function Validation

m607123:
Validation added for the Aggregate, Filter, and Transform functions.
Now requires:
An operation that is not null
A schema that is not null
For each element group defined in the "Edges" or "Entities" map of the operation, they must be defined in the schema
Functions to execute should be not null
Any operations/predicates should be applicable to the property upon which they are to be executed


core/operation/src/main/java/uk/gov/gchq/gaffer/operation/impl/function/Function.java
```diff
@@ -0,0 +1,34 @@
...
+package uk.gov.gchq.gaffer.operation.impl.function;
+
+import uk.gov.gchq.gaffer.data.element.Element;
+import uk.gov.gchq.gaffer.operation.Operation;
+import uk.gov.gchq.gaffer.operation.io.InputOutput;
+import uk.gov.gchq.gaffer.operation.io.MultiInput;
+
+import java.util.Map;
+
+public abstract class Function implements
```
p013570: Please could you make this an interface we are trying to avoid lots of abstract operation classes.

core/data/src/main/java/uk/gov/gchq/gaffer/data/element/function/ElementFilter.java
```diff
@@ -31,7 +31,7 @@
 import java.util.function.Predicate;
 
 /**
- * An {@code ElementFiler} is a {@link Predicate} which evaluates a condition against
+ * An <code>ElementFilter</code> is a {@link Predicate} which evaluates a condition against
```
p013570: Please can you use {@code }

core/store/src/main/java/uk/gov/gchq/gaffer/store/operation/handler/function/FilterHandler.java
```diff
@@ -50,6 +51,13 @@
             operation.setEdges(edgeMap);
         }
 
+        final FunctionValidator<Filter> validator = new FunctionValidator<>();
```
p013570: Rather than creating a new instance of this for every query, the validator could be a 'private final' field.

m607123:
Closing while changes made

core/operation/src/main/java/uk/gov/gchq/gaffer/operation/impl/function/Function.java
```diff
@@ -0,0 +1,26 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package uk.gov.gchq.gaffer.operation.impl.function;
+
+import java.util.Map;
+
+public interface Function {
```
t616178: Missing class level and method level javadoc.

core/operation/src/main/java/uk/gov/gchq/gaffer/operation/impl/function/Filter.java
```diff
@@ -39,10 +39,10 @@
  * For more complex queries, a {@link Map} of groups to {@link ElementFilter}s can also be provided to either
  * {@link  uk.gov.gchq.gaffer.data.element.Edge}s or {@link uk.gov.gchq.gaffer.data.element.Entity}s.
  */
-public class Filter implements
+public class Filter implements Function,
         Operation,
         InputOutput<Iterable<? extends Element>, Iterable<? extends Element>>,
-        MultiInput<Element> {
+        MultiInput<Element>  {
```
t616178: Try to avoid adding extra unnecessary whitespace.

core/operation/src/main/java/uk/gov/gchq/gaffer/operation/impl/function/Filter.java
```diff
@@ -39,10 +39,10 @@
  * For more complex queries, a {@link Map} of groups to {@link ElementFilter}s can also be provided to either
  * {@link  uk.gov.gchq.gaffer.data.element.Edge}s or {@link uk.gov.gchq.gaffer.data.element.Entity}s.
  */
-public class Filter implements
+public class Filter implements Function,
         Operation,
```
t616178: This is a redundant interface implementation, since the InputOutput<?> interface extends Operation.

core/store/src/main/java/uk/gov/gchq/gaffer/store/operation/handler/function/FunctionValidator.java
```diff
@@ -0,0 +1,300 @@
...
+        if (null != aggregator) {
+            final List<TupleAdaptedBinaryOperator<String, ?>> components = aggregator.getComponents();
+
+            for (final TupleAdaptedBinaryOperator<String, ?> component : components) {
+                final String[] selection = component.getSelection();
+
+                for (final SchemaElementDefinition elementDef : elements.values()) {
+                    final Class[] selectionClasses = Arrays.stream(selection).map(elementDef::getPropertyClass).toArray(Class[]::new);
+                    final Map<String, String> properties = elementDef.getPropertyMap();
+                    if (properties.size() > 0) {
```
t616178: We should prefer the use of ```!Map.isEmpty()``` here.

core/store/src/main/java/uk/gov/gchq/gaffer/store/operation/handler/function/FunctionValidator.java
```diff
@@ -0,0 +1,300 @@
...
+                }
+            }
+        } else {
+            for (final SchemaElementDefinition elementDef : elements.values()) {
+                final List<TupleAdaptedBinaryOperator<String, ?>> components = elementDef.getOriginalAggregateFunctions();
+                for (final TupleAdaptedBinaryOperator<String, ?> component : components) {
+                    final String[] selection = component.getSelection();
+                    final Class[] selectionClasses = Arrays.stream(selection).map(elementDef::getPropertyClass).toArray(Class[]::new);
+                    final Map<String, String> properties = elementDef.getPropertyMap();
+                    if (properties.size() > 0) {
```
t616178: We should prefer the use of ```!Map.isEmpty()``` here.

core/store/src/main/java/uk/gov/gchq/gaffer/store/operation/handler/function/FunctionValidator.java
```diff
@@ -0,0 +1,300 @@
...
+        final ValidationResult result = new ValidationResult();
+
+        final List<TupleAdaptedPredicate<String, ?>> components = filter.getComponents();
+        for (final TupleAdaptedPredicate<String, ?> component : components) {
+            final String[] selection = component.getSelection();
+
+            for (final SchemaElementDefinition elementDef : elements.values()) {
+                final Class[] selectionClasses = Arrays.stream(selection).map(elementDef::getPropertyClass).toArray(Class[]::new);
+                final Map<String, String> properties = elementDef.getPropertyMap();
+                if (properties.size() > 0) {
```
t616178: We should prefer the use of ```!Map.isEmpty()``` here.

core/store/src/main/java/uk/gov/gchq/gaffer/store/operation/handler/function/FunctionValidator.java
```diff
@@ -0,0 +1,300 @@
...
+        final List<TupleAdaptedFunction<String, ?, ?>> components = transformer.getComponents();
+        for (final TupleAdaptedFunction<String, ?, ?> component : components) {
+            final String[] selection = component.getSelection();
+            final String[] projection = component.getProjection();
+
+            for (final SchemaElementDefinition elementDef : elements.values()) {
+                final Class[] selectionClasses = Arrays.stream(selection).map(elementDef::getPropertyClass).toArray(Class[]::new);
+                final Class[] projectionClasses = Arrays.stream(projection).map(elementDef::getPropertyClass).toArray(Class[]::new);
+                final Map<String, String> properties = elementDef.getPropertyMap();
+                if (properties.size() > 0) {
```
t616178: We should prefer the use of ```!Map.isEmpty()``` here.

core/store/src/main/java/uk/gov/gchq/gaffer/store/operation/handler/function/FunctionValidator.java
```diff
@@ -0,0 +1,300 @@
...
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.function.BinaryOperator;
+import java.util.stream.Collectors;
+
+public class FunctionValidator<T extends Function> {
```
p013570: A lot of the logic in this class is duplicated in ViewValidator. Could you make a utility class for the common logic?

core/store/src/main/java/uk/gov/gchq/gaffer/store/operation/handler/function/FunctionValidator.java
```diff
@@ -0,0 +1,300 @@
...
+        }
+
+        result.add(validateOperation(operation, schema));
+
+        return result;
+    }
+
+    private ValidationResult validateOperation(final T operation, final Schema schema) {
+        final ValidationResult result = new ValidationResult();
+        final boolean opIsAggregate = operation instanceof Aggregate;
```
p013570: Rather than doing this, it would be better to have separate classes: AggregateValidator, FilterValidator and TransformFilter, which extends FunctionValidator.

m607123:
Closing while changes made

p013570:
When using the Aggregator:

If you don't supply any groups then it should default to using all groups, similar to the Filter Operation (perhaps also check the Transform Operation)
If you aggregate a transient property then the validation fails as it can't find the class of the transient property. Perhaps just skip the validation for transient properties.
I don't think we need the 'else' block on line 138 of the AggregateValidator.

Please can you check all 3 operations in the REST API too.

p013570:

m607123:
Checked working as intended in the REST API, with the Road Traffic Demo.

Issue #1379, opened 2017-10-05
Change All Spark Operations to accept Spark Session


james010101101:
Some spark based operations accept sparksessions, while some acceptJavaSparkContext's, this should be standardised so all operations accept sparksessions, 

A better method of supplying the spark sessions to the operations is needed, currently a spark session is provided as a mandatory option on each operation, which has the side effect that spark operations can't be run over the rest service.

I propose we change this so that you create a gaffer "spark context" (inheriting from the gaffer context object)  in your store if you enable spark operations, which will then contain a spark session.which can be retrieved in the handler

p013570:
#Migration Steps
The GraphHook interface has changed so it now takes a Context object instead of just the user object.

p013570:
Merged into develop.

PR #1385, opened 2017-10-05
Gh1379 change all spark operations to accept spark session

james010101101:


p013570:
Rather than always creating a SparkSession, if possible we should do it lazily. If you do a GetElements operation you might not need a spark session so it seems wasteful to add it.

I've refactored the code and pushed it to branch: gh-1379-Change-All-Spark-Operations-to-accept-Spark-Session-lazy. Instead the handlers now just ask the context for a spark session, if one doesn't exist it creates one and adds to back to the context so it can be retrieved again later.

I've raised a separate pull request - https://github.com/gchq/Gaffer/pull/1387

@james010101101 let me know what you think. I'll close this pull request for now, but please reopen it if you don't like my changes.

Issue #605, opened 2016-12-14
Integrate release scripts with Travis


t616178:
Integrate the release scripts so that Travis handles all aspects of the release process.

p013570:
Merged into develop.

PR #1388, opened 2017-10-06
gh-605 - added deployment scripts to tag master branch

p013570:


p013570:
Tested on forked Gaffer repo and it seems to work ok.

PR #1399, opened 2017-10-10
Gh 1398 npe when set id is null store properties

m55624:


core/graph/src/main/java/uk/gov/gchq/gaffer/graph/Graph.java
```diff
@@ -851,8 +851,10 @@ private void updateStore(final GraphConfig config) {
             if (null != properties) {
                 if (null == mergedStoreProperties) {
                     mergedStoreProperties = properties;
+                    mergedStoreProperties.setId(properties.getId());
```
p013570: This line is not required.

core/graph/src/main/java/uk/gov/gchq/gaffer/graph/Graph.java
```diff
@@ -851,8 +851,10 @@ private void updateStore(final GraphConfig config) {
...
                 } else {
                     mergedStoreProperties.getProperties().putAll(properties.getProperties());
+                    mergedStoreProperties.setId(mergedStoreProperties.getId() + "," + properties.getId());
```
p013570: mergedStoreProperties.getId() and/or properties.getId() could be null.

Issue #1400, opened 2017-10-10
Update logos with ascii art


GCHQDeveloper404:

p013570:
Merged into develop.

PR #1401, opened 2017-10-10
gh-1400-ascii-logo

GCHQDeveloper404:


logos/asciiLogo.txt
```diff
@@ -0,0 +1,16 @@
...
+           _>  / '     \             o
+          _> ,/ .  @_^`^)                      O
+         -   |.   /_,__ )     o
+         _> | /   '  (.       ________        _____  __O__
+          >_(/ _    (_ \     /  _____/_____ _/ ____\/ ____\___________
+            /.'    (  `.\   /   \  ___\__  \\   __\\   __\/ __ \_  ___\
+              (   (        \    \_\  \/ __ \|  |   |  | \  ___/|  |
+               (   (        \________(__ __ /__|   |__|  \_____|__|
+                `( `l./^^>              /  \
+                  `l.  /                \__/
```
gaffer01: Can the bubble under the 'a' be moved a bit lower? I know it's like that in the png version but here it looks like the bubble is part of the 'a' which is confusing.

PR #1403, opened 2017-10-10
Gh 1381 time series property

gaffer01:


library/time-library/src/main/java/uk/gov/gchq/gaffer/time/LongTimeSeries.java
```diff
@@ -0,0 +1,225 @@
...
+
+/**
+ * This is a time series where the values are {@link Long}s. When the time series
+ * is created, a {@link TimeBucket} is specified. When timestamps are added, they
+ * are rounded to the nearest bucket, e.g. if a {@link TimeBucket} of
+ * <code>MINUTE</code> is specified, and a timestamp of January 1st 2017, 12:34:56
+ * is added then the seconds are removed so that the value is associated to
+ * 12:34.
+ */
+public class LongTimeSeries implements TimeSeries<Long> {
```
p013570: This class doesn't look like it will serialise to/from JSON properly. You have 2 getters for returning the time series information: getInstants and getTimeSeries, one of these needs to be annotated with @JsonIgnore otherwise you will get the same data twice in JSON. You also need a default constructor and a setter for the timeSeries and timeBucket fields. 

As a lot of Gaffer interactions are done via REST in JSON, it would be great if this class worked over REST. 

Your LongTimeSeriesTest should then extend JSONSerialisationTest and that will add a json serialisation test.

PR #1432, opened 2017-10-16
Gh 1301 path operation

t616178:
First attempt at implementing a "Path" operation for Gaffer. There are some known issues to be fixed soon:

* Implementation of operation scoring for the Path operation
* Fine-grained filtering based on entity groups or properties
* Performance testing and tweaks


core/store/src/main/java/uk/gov/gchq/gaffer/store/operation/handler/PathHandler.java
```diff
@@ -0,0 +1,175 @@
...
+import java.util.ArrayList;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+import java.util.Stack;
+import java.util.stream.Collectors;
+
+public class PathHandler implements OutputOperationHandler<Path, Iterable<Iterable<Edge>>> {
+
+    private final Set<EntitySeed> visitedSeeds = new HashSet<>();
```
p013570: If we made this a Set of just the vertex values instead of the wrapper EntitySeed object then it would mean we could be slightly more efficient with our memory usage.

core/store/src/main/java/uk/gov/gchq/gaffer/store/operation/handler/PathHandler.java
```diff
@@ -0,0 +1,175 @@
...
+    private int pathVertices;
+    private int hops;
+
+    @Override
+    public Iterable<Iterable<Edge>> doOperation(final Path operation, final Context context, final Store store) throws OperationException {
+
+        // Check input
+        if (null == operation.getInput() || !operation.getInput().iterator().hasNext()) {
```
p013570: We really need to avoid doing operation.getInput().iterator().hasNext(). It is really inefficient for when run on results from accumulo and it causes the connection to accumulo to not be closed properly.

p013570: If the input is empty then I think we should return an empty iterable to make it consistent with other operations.

core/store/src/main/java/uk/gov/gchq/gaffer/store/operation/handler/PathHandler.java
```diff
@@ -0,0 +1,175 @@
...
+            return null;
+        }
+
+        // Check operations input
+        final Iterable<GetElements> operations = operation.getOperations();
+
+        if (null == operations || Iterables.isEmpty(operations)) {
+            return null;
```
p013570: Should this return an empty iterable?

core/store/src/main/java/uk/gov/gchq/gaffer/store/operation/handler/PathHandler.java
```diff
@@ -0,0 +1,175 @@
...
+        }
+
+        // Validate the View objects
+        for (final GetElements op : operations) {
```
p013570: This validation could go in the Path.validate method. That would mean the validation would be applied before the Operation Chain is started to be executed.

core/store/src/main/java/uk/gov/gchq/gaffer/store/operation/handler/PathHandler.java
```diff
@@ -0,0 +1,175 @@
...
+            if (null != op.getView() && op.getView().hasEntities()) {
+                throw new OperationException("The view for operation " + op + " must not contain Entities.");
```
p013570: GetElements.toString() isn't very useful. Perhaps you could say something like:

The GetElement operations must not contain Entities. Please fix the view for the third GetElements operation.

core/store/src/main/java/uk/gov/gchq/gaffer/store/operation/handler/PathHandler.java
```diff
@@ -0,0 +1,175 @@
...
+            }
+        }
+
+        Iterable<? extends EntitySeed> seeds = operation.getInput();
+
+        hops = Iterables.size(operation.getOperations());
```
p013570: operations is just a list so you can do operation.getOperations().size()

core/store/src/main/java/uk/gov/gchq/gaffer/store/operation/handler/PathHandler.java
```diff
@@ -0,0 +1,175 @@
...
+        for (final GetElements op : operations) {
+
+            final Set<EntitySeed> opSeeds = difference(visitedSeeds, Sets.newHashSet(seeds));
+
+            op.setInput(opSeeds);
+            final Iterable<? extends Element> results = store.execute(op, context);
+            visitedSeeds.addAll(opSeeds);
+
+            // Cache results
+            Lists.newArrayList(results).stream()
```
p013570: I think we should avoid doing: Lists.newArrayList(results)
As it will bring all the results into memory. Instead can we do: Streams.toStream(results)

core/store/src/main/java/uk/gov/gchq/gaffer/store/operation/handler/PathHandler.java
```diff
@@ -0,0 +1,175 @@
...
+                    .map(e -> (Edge) e)
```
p013570: I think you may as well remove this map phase and just cast the results iterable to Iterable<Edge>.

core/store/src/main/java/uk/gov/gchq/gaffer/store/operation/handler/PathHandler.java
p013570: Actually you could just do this logic in a for loop instead of using streams.

core/store/src/main/java/uk/gov/gchq/gaffer/store/operation/handler/PathHandler.java
```diff
@@ -0,0 +1,175 @@
...
+                    .forEach(e -> {
+                        if (null != edgeTable.get(e.getMatchedVertexValue(), e.getAdjacentMatchedVertexValue())) {
```
p013570: You could store this as a local variable to avoid having to retrieve it from the edgeTable twice:

edgeTable.get(e.getMatchedVertexValue(), e.getAdjacentMatchedVertexValue()

core/store/src/main/java/uk/gov/gchq/gaffer/store/operation/handler/PathHandler.java
```diff
@@ -0,0 +1,175 @@
...
+                            final Set<Edge> set = edgeTable.get(e.getMatchedVertexValue(), e.getAdjacentMatchedVertexValue());
+                            set.add(e);
+                            edgeTable.put(e.getMatchedVertexValue(), e.getAdjacentMatchedVertexValue(), set);
```
p013570: I'm assuming this is a limitation of the HashBasedTable that we have to reinsert the value?

core/store/src/main/java/uk/gov/gchq/gaffer/store/operation/handler/PathHandler.java
```diff
@@ -0,0 +1,175 @@
...
+            final OperationChain<Set<? extends EntitySeed>> opChain = new OperationChain.Builder()
+                    .first(new ToVertices.Builder()
+                            .input(results)
+                            .edgeVertices(ToVertices.EdgeVertices.DESTINATION)
+                            .build())
+                    .then(new ToEntitySeeds())
+                    .then(new ToSet<>())
+                    .build();
+
+            seeds = store.execute(opChain, context);
```
p013570: You could probably optimise this and just do the conversion yourself here. The ToVertices handler has extra logic you could skip.

core/store/src/main/java/uk/gov/gchq/gaffer/store/operation/handler/PathHandler.java
```diff
@@ -0,0 +1,175 @@
...
+                            final Set<Edge> set = new HashSet<>();
+                            set.add(e);
+                            edgeTable.put(e.getMatchedVertexValue(), e.getAdjacentMatchedVertexValue(), set);
+                        }
+                    });
+
+            final OperationChain<Set<? extends EntitySeed>> opChain = new OperationChain.Builder()
+                    .first(new ToVertices.Builder()
+                            .input(results)
+                            .edgeVertices(ToVertices.EdgeVertices.DESTINATION)
```
p013570: I think we would want to extract the opposite MatchedVertex vertex.

core/store/src/main/java/uk/gov/gchq/gaffer/store/operation/handler/PathHandler.java
```diff
@@ -0,0 +1,175 @@
...
+                            .build())
+                    .then(new ToEntitySeeds())
```
p013570: Perhaps we could skip the ToEntitySeeds operation here and just convert them to entity seeds as we execute the next operation (if there is even a next operation)

core/store/src/main/java/uk/gov/gchq/gaffer/store/operation/handler/PathHandler.java
```diff
@@ -0,0 +1,175 @@
...
+            paths.add(edges);
+        } else {
+            for (int i = 1; i < path.size(); i++) {
+                final List<Edge> edgeList = new ArrayList<>(graph.get(path.get(i - 1), path.get(i)));
+
+                if (edgeList.size() == 1) {
+                    edges.add(edgeList.get(0));
+                } else {
+                    for (final Edge edge : edgeList) {
+                        final List<Edge> tmp = new ArrayList<>(edges);
```
p013570: Can we avoid creating a new ArrayList here?

core/store/src/main/java/uk/gov/gchq/gaffer/store/operation/handler/PathHandler.java
```diff
@@ -0,0 +1,175 @@
...
+    private <T, U> List<List<T>> doPath(final T source, final HashBasedTable<T, T, ?> graph, final Stack<T> path) {
+        final List<List<T>> paths = new ArrayList<>();
+
+        path.push(source);
+
+        if (pathVertices == path.size()) {
+            paths.add(new ArrayList<>(path));
+        } else {
+            for (final T obj : graph.row(source).keySet()) {
+                paths.addAll(new ArrayList<>(doPath(obj, graph, path)));
```
p013570: Again, perhaps we could remove the new ArrayList construction here?

core/store/src/main/java/uk/gov/gchq/gaffer/store/operation/handler/PathHandler.java
```diff
@@ -0,0 +1,175 @@
...
+        return paths;
+    }
+
+    private <T, U> List<List<T>> doPath(final T source, final HashBasedTable<T, T, ?> graph, final Stack<T> path) {
+        final List<List<T>> paths = new ArrayList<>();
+
+        path.push(source);
+
+        if (pathVertices == path.size()) {
+            paths.add(new ArrayList<>(path));
```
p013570: Is this new ArrayList required?

t616178:
Closing for updates.

t616178:
Reopened after changes applied.

core/operation/src/main/java/uk/gov/gchq/gaffer/operation/impl/GetWalks.java
```diff
@@ -0,0 +1,137 @@
...
+import uk.gov.gchq.gaffer.operation.serialisation.TypeReferenceImpl;
+import uk.gov.gchq.koryphe.ValidationResult;
+
+import java.util.Arrays;
+import java.util.List;
+import java.util.ListIterator;
+import java.util.Map;
+
+/**
+ * A {@code GetWalks} class is used to retrieve all of the paths in a graph
```
gaffer01: "all of the paths" should be "all of the walks"

store-implementation/accumulo-store/src/main/java/uk/gov/gchq/gaffer/accumulostore/utils/IngestUtils.java
```diff
@@ -63,7 +63,7 @@ private IngestUtils() {
      * @param conn       - An existing connection to an Accumulo instance
      * @param table      - The table name
      * @param fs         - The FileSystem in which to create the splits file
-     * @param splitsFile - A Path for the output splits file
+     * @param splitsFile - A GetWalks for the output splits file
```
gaffer01: This should be changed back to Path

store-implementation/accumulo-store/src/main/java/uk/gov/gchq/gaffer/accumulostore/utils/IngestUtils.java
```diff
@@ -128,7 +128,7 @@ public static int createSplitsFile(final Connector conn, final String table, fin
      * Read a splits file and get the number of split points within
      *
      * @param fs         - The FileSystem in which to create the splits file
-     * @param splitsFile - A Path for the output splits file
+     * @param splitsFile - A GetWalks for the output splits file
```
gaffer01: This should be changed back to Path

store-implementation/accumulo-store/src/main/java/uk/gov/gchq/gaffer/accumulostore/utils/IngestUtils.java
```diff
@@ -152,7 +152,7 @@ public static int getNumSplits(final FileSystem fs, final Path splitsFile) throw
      * <p>
      *
      * @param fs      - The FileSystem in which to create the splits file
-     * @param dirPath - The Path to the directory
+     * @param dirPath - The GetWalks to the directory
```
gaffer01: This should be changed back to Path

store-implementation/hbase-store/src/main/java/uk/gov/gchq/gaffer/hbasestore/utils/IngestUtils.java
```diff
@@ -45,7 +45,7 @@ private IngestUtils() {
      * <p>
      *
      * @param fs      - The FileSystem in which to create the splits file
-     * @param dirPath - The Path to the directory
+     * @param dirPath - The GetWalks to the directory
```
gaffer01: This should be changed back to Path

core/store/src/main/java/uk/gov/gchq/gaffer/store/operation/handler/GetWalksHandler.java
```diff
@@ -0,0 +1,144 @@
...
+                } else {
+                    final Set<Edge> set = new HashSet<>();
+                    set.add(e);
+                    subgraph.put(e.getMatchedVertexValue(), e.getAdjacentMatchedVertexValue(), set);
+                }
+            }
+
+            final OperationChain<Iterable<? extends EntitySeed>> opChain = new OperationChain.Builder()
+                    .first(new ToVertices.Builder()
+                            .input(results)
```
gaffer01: It looks like the `Iterable` `results` will be consumed twice causing the same edges to be pulled back from the database twice. Can this be avoided? The results are being cached in memory anyway. Maybe `results` should be cached?

p013570:
Closing while further changes are made.

t616178:
Reopening after changes applied.

core/store/src/main/java/uk/gov/gchq/gaffer/store/operation/handler/GetWalksHandler.java
```diff
@@ -0,0 +1,147 @@
...
+
+/**
+ * An operation handler for {@link GetWalks} operations.
+ * <p>
+ * Currently the handler only supports creating {@link Walk}s which contain
+ * {@link Edge}s.
+ */
+public class GetWalksHandler implements OutputOperationHandler<GetWalks, Iterable<Walk>> {
+
+    private int hops;
```
p013570: Perhaps add a field for maxHops and default it to null. If you ensure this class is json serialisable then this would allow a sys admin to set a maximum number of hops. Then when the operation is handled we can verify that the number of hops doesn't exceed the maximum allowed.

core/store/src/main/java/uk/gov/gchq/gaffer/store/operation/handler/GetWalksHandler.java
```diff
@@ -0,0 +1,147 @@
...
+        return Streams.toStream(getWalks.getInput())
+                .map(seed -> walk(seed.getVertex(), null, adjacencyLists, new LinkedList<>()))
+                .flatMap(List::stream)
+                .collect(Collectors.toList());
+    }
+
+    private List<Walk> walk(final Object curr, final Object prev, final List<AdjacencyList<Object, Edge>> adjacencyLists, final LinkedList<Set<Edge>> queue) {
+        final List<Walk> walks = new ArrayList<>();
+
+        if (null != prev && hops != queue.size()) {
```
p013570: Perhaps store queue.size() as a variable called hop. To help document that you are using it to calculate which hop you are looking at.

core/store/src/main/java/uk/gov/gchq/gaffer/store/operation/handler/GetWalksHandler.java
```diff
@@ -0,0 +1,147 @@
...
+        if (hops == queue.size()) {
+            final Walk.Builder builder = new Walk.Builder();
+            for (final Set<Edge> edgeSet : queue) {
+                builder.edges(edgeSet);
+            }
+            final Walk walk = builder.build();
+            walks.add(walk);
+        } else {
+            for (final Object obj : adjacencyLists.get(queue.size()).getDestinations(curr)) {
+                walks.addAll(walk(obj, curr, adjacencyLists, queue));
```
p013570: I'm concerned that we could get a StackOverflowError here if there are a large number of edges coming off a single vertex.

integration-test/src/test/java/uk/gov/gchq/gaffer/integration/impl/StoreValidationIT.java
```diff
@@ -23,7 +24,7 @@
 
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertTrue;
-
+@Ignore
```
p013570: Remove this ignore annotation.

p013570:
Closing while further testing is done.

core/store/src/main/java/uk/gov/gchq/gaffer/store/operation/handler/GetWalksHandler.java
t616178: This is difficult to do since the size of the queue can change after that variable would have been assigned (we'd have to track all changes to the size of the queue). I'll add some documentation to make this clearer.

core/store/src/main/java/uk/gov/gchq/gaffer/store/operation/handler/GetWalksHandler.java
t616178: The number of stack frames is correlated with the number of hops, rather than the number of vertices, so this shouldn't be an issue.

core/store/src/main/java/uk/gov/gchq/gaffer/store/operation/handler/GetWalksHandler.java
t616178: Done

integration-test/src/test/java/uk/gov/gchq/gaffer/integration/impl/StoreValidationIT.java
t616178: Done

Issue #1433, opened 2017-10-16
Allow group and property names to use hyphens


p013570:
Currently we validate group and property names using the regex: [a-zA-Z0-9|]*

It would be useful to also allow hyphens.

This would mean updating the regex in GroupUtil and PropertiesUtil to include hyphens. Also need some tests for it.

p013570:
Merged into develop.

PR #1441, opened 2017-10-17
gh-1433 - Updated regex and unit tests

m607123:
Regex updated to allow the use of hyphens
Updated unit tests to include hyphens


m607123:
Currently the regex allows groups and properties to also start with either the pipe symbol and now a hyphen - is this behaviour acceptable?

p013570:
Ideally we would probably not want to allow names to start with a | or a - however, making that change now could result in a breaking change for some users. So for now we will just have to allow it, so what you have done looks good.

Issue #1418, opened 2017-10-13
Update description at top of REST API Swagger UI


p013570:
Currently the description is:

```
Wiki
https://github.com/gchq/Gaffer/wiki
Created by Gaffer Developers
See more at https://github.com/gchq/Gaffer
Apache 2.0
```

The wiki link should be removed and changed to the gaffer-doc website.

p013570:
Merged into develop.

PR #1442, opened 2017-10-17
gh-1418 - Updated wiki link

m607123:
Wiki link updated
Checked working in REST API


rest-api/core-rest/src/main/java/uk/gov/gchq/gaffer/rest/service/v1/SwaggerDefinitionConfig.java
```diff
@@ -45,7 +45,7 @@
         consumes = {APPLICATION_JSON},
         produces = {APPLICATION_JSON},
         schemes = {SwaggerDefinition.Scheme.HTTP, SwaggerDefinition.Scheme.HTTPS},
-        externalDocs = @ExternalDocs(value = "Wiki", url = "https://github.com/gchq/Gaffer/wiki")
+        externalDocs = @ExternalDocs(value = "Wiki", url = "https://gchq.github.io/gaffer-doc/")
```
p013570: Please can you change the name of it to "Documentation" instead of "Wiki". Perhaps you could create a constant for the value so you don't have it hardcoded in 2 places.

PR #1447, opened 2017-10-17
Gh 1360 validate user operations

PammyS:


p013570:
Hi @PammyS,

Sorry I hadn't seen the commit @m607123 had made on his branch until now. It looks like you may have been led down the wrong path. I think the validation should really have gone in the Graph class in one place at the starting point for all interactions with Gaffer.

I think we need to add something like the following to line 199 in Graph, just after the null check for the operationChain parameter:

```java
if(null == context) {
    throw new IllegalArgumentException("A context containing a user is required");
}

if(null == context.getUser()) {
    throw new IllegalArgumentException("The context does not contain a user");
}
```

The only other thing required would be some simple unit tests to check the validation. The testing should go in GraphTest. There are several different entry points to test, each with a null context or null user.
graph.execute(OperationChain, Context)
graph.execute(OperationChain, User)
graph.executeJob(OperationChain, Context)
graph.executeJob(OperationChain, User)

An example test for graph.execute(OperationChain, Context) is:

```java
@Test
public void shouldThrowExceptionOnExecuteWithANullContext() throws OperationException {
    // Given
    final Context context = null;
    final OperationChain opChain = mock(OperationChain.class);

    final Graph graph = new Graph.Builder()
            .config(new GraphConfig.Builder()
                    .graphId(GRAPH_ID)
                    .build())
            .storeProperties(StreamUtil.storeProps(getClass()))
            .addSchemas(StreamUtil.schemas(getClass()))
            .build();

    // When / Then
    try {
        graph.execute(opChain, context);
        fail("Exception expected");
    } catch (final IllegalArgumentException e) {
        assertEquals("A context containing a user is required", e.getMessage());
    }
}
```

I'll leave you to test the other bits.

PammyS:
@p013570 

Thanks a lot for your feedback! I have made appropriate changes and added tests cases as well but pushing code in a new branch as I realised there were some merge conflicts present and, the prior changes were also not required

New PR : 1448

p013570:
Closing this pull request as this has been replaced by #1448.

PR #1448, opened 2017-10-17
Fix/user null check

PammyS:
Added user null check and test cases for the same

### DIFF
--- a/core/graph/src/main/java/uk/gov/gchq/gaffer/graph/Graph.java
+++ b/core/graph/src/main/java/uk/gov/gchq/gaffer/graph/Graph.java
@@ -233,7 +233,9 @@ public final class Graph {
 
     private void updateOperationChainView(final OperationChain<?> operationChain) {
         for (final Operation operation : operationChain.getOperations()) {
-            if (operation instanceof OperationView) {
+            if (operation instanceof OperationChain) {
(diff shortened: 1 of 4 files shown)


p013570:
The tests look good. Just noticed a couple of issues highlighted by the Travis CI build:

- our checkstyle has failed due to missing whitespace around 'if' in the code you added to Graph.
  - you can run checkstyle on the project using the command "mvn clean verify -Panalyze". If you just want to run checkstyle on one module, e.g graph, then you can do "mvn clean verify -Panalyze -pl :graph".
- As a result of your change it highlights a couple of issues in our code elsewhere which caused tests to fail:
  - in GafferResultCacheExporter can you change "context.getUser()" to just "context" on lines 121 and 138.
  - in FederatedOperationHandler can you also change "context.getUser()" to just "context" on line 46.

That should get the tests to pass. You can verify the code builds successfully using: 
"mvn clean install", although this may take about 10 minutes to run. If you want to just test the modules you have changed (e.g graph and federated-store) you can do something like:
"mvn clean install -Pquick" followed by "mvn clean install -pl :graph,:federated-store"

PammyS:
@p013570 
Thanks again for the feedback, I have made the changes and check the tests, its working!

PammyS:
@p013570 
Thanks a lot! Sure will pick them one by one

PR #1453, opened 2017-10-18
Gh 1243 operation chain handler

p013570:


p013570:
Just need to make a couple of changes.

Issue #1452, opened 2017-10-18
Add ToMap example to ExamplesFactory


p013570:
Add a new interface method to ExamplesFactory:

```java
ToMap toMap();
```

Add a new implementation to DefaultExamplesFactory. See [doc](https://gchq.github.io/gaffer-doc/getting-started/operation-examples.html#tomap-example) for some examples of using the ToMap operation.

Hook the example up in the DefaultExamplesFactory.generateExample method, i.e. add a new if block for:
```java
else if (operation instanceof ToMap) {
    return toMap(); 
}
```

t616178:
Merged into develop.

Issue #1451, opened 2017-10-18
Add Min example to ExamplesFactory


p013570:
Add a new interface method to ExamplesFactory:

```java
Min min();
```

Add a new implementation to DefaultExamplesFactory. See [doc](https://gchq.github.io/gaffer-doc/getting-started/operation-examples.html#min-example) for some examples of using the Min operation.

Hook the example up in the DefaultExamplesFactory.generateExample method, i.e. add a new if block for:
```java
else if (operation instanceof Min) {
    return min(); 
}
```

t616178:
Merged into develop.

Issue #1450, opened 2017-10-18
Add Max example to ExamplesFactory


p013570:
Add a new interface method to ExamplesFactory:

```java
Max max();
```

Add a new implementation to DefaultExamplesFactory. See [doc](https://gchq.github.io/gaffer-doc/getting-started/operation-examples.html#max-example) for some examples of using the Max operation.

Hook the example up in the DefaultExamplesFactory.generateExample method, i.e. add a new if block for:
```java
else if (operation instanceof Max) {
    return max(); 
}
```

t616178:
Merged into develop.

PR #1460, opened 2017-10-19
added examples for max min toMap

PammyS:
Issues refereed #1450, #1451, #1452.


p013570:
They all look good and seem to work nicely. If you haven't already you can see these examples in action if you start up the rest api using:
```bash
mvn clean install -Pquick -Proad-traffic-demo -pl :road-traffic-demo -am
```

Then you will see the swagger ui if you go to:  http://localhost:8080/rest/

Then if you expand the "Operation" then "/graph/operations/execute" section you will see a dropdown menu. Within that dropdown list there are all the Operations. If you select Min or Max or ToMap then click "Add Example" you will see that your example gets serialised into json and added to the input box to the left. This allows users to get an example operation template that they can then manipulate and use.

As an improvement, it would be nice to avoid hardcoding the strings "entity", "edge" and "count". It would be better to use the getAnEntityGroup() and getAnEntityPropertyName() methods to dynamically inspect the current Graph and select a valid string based on the schema. 

So instead of:
```java
final Max max = new Max.Builder()
                .comparators(new ElementPropertyComparator.Builder()
                        .groups("entity", "edge")
                        .property("count")
                        .build())
                .build();
```

we could have:

```java
final Max max = new Max.Builder()
                .comparators(new ElementPropertyComparator.Builder()
                        .groups(getAnEdgeGroup())
                        .property(getAnEdgePropertyName())
                        .build())
                .build();
```

This same improvement could be applied to Min, Max, ToMap and Sort. If we make this change then when we start the Road Traffic Demo Rest Api (as described above) you will see it uses a valid group name like "RoadUse" and valid property name like "startDate".

p013570:
@PammyS the values for reverse, deduplicate, resultLimit can just be hardcoded - they are just example values and would work for any Gaffer Graph, they don't rely on a specific schema.

to get an example vertex you can use:
```java
getExampleVertex(entityDef.getIdentifierClass(IdentifierType.VERTEX), 1)
```

to get an example source:
```java
getExampleVertex(edgeDef.getIdentifierClass(IdentifierType.SOURCE), 2)
```

PammyS:
@p013570 
for both cases its't giving error as both the  getExampleVertex(entityDef.getIdentifierClass(IdentifierType.VERTEX), 1)
getExampleVertex(edgeDef.getIdentifierClass(IdentifierType.SOURCE), 2)
return type is object and vertex and source input type is Sting. 
Should I use toString on these return object of getExampleVertex function?

p013570:
@PammyS - I've just taken a look at your ToMap example and seen where you are trying to insert the source and vertex values. The "source" string is actually a string representing the key in the output map, so it can be any value you want. So if you put "some string" as the source value the output map would contain:

{
    "some string": theSourceValue
}

So I would recommend something like:

```java
 final ToMap toMap = new ToMap.Builder()
                .generator(new MapGenerator.Builder()
                        .group(getAnEdgeGroup())
                        .source("the edge source")
                        .property(getAnEdgePropertyName(), "the edge property " + getAnEdgePropertyName())
                        .build())
                .build();
```


Thanks for highlighting the lack of documentation around those classes. I've added a new issue #1463 to update the documentation for the MapGenerator class to inform users how to configure it.

PammyS:
@p013570 
thanks! I have updated and re-based  with develop before pushing. Please take a look.

t616178:
@PammyS Thanks for adding in those changes. Everything looks good from here, so I've merged the changes into our fork.

PR #1461, opened 2017-10-19
Gh 1431 merging schema and storeprops ids

p013570:


p013570:
Closing whilst we consider removing schema and store properties IDs. See #1477.

PR #1471, opened 2017-10-25
Added a /properties endpoint, with the ability to configure which pro…

james010101101:
…perties are visible


rest-api/core-rest/src/main/java/uk/gov/gchq/gaffer/rest/ServiceConstants.java
```diff
@@ -32,6 +32,7 @@
 
     public static final String CLASS_NOT_FOUND = "Class not found";
     public static final String FUNCTION_NOT_FOUND = "Function not found";
+    public static final String PROPERTY_NOT_FOUND = "Property not found";
```
javadev001001: Is this a duplicated line?

rest-api/core-rest/src/main/java/uk/gov/gchq/gaffer/rest/service/v2/PropertiesServiceV2.java
```diff
@@ -0,0 +1,56 @@
...
+
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * An implementation of {@link IPropertiesServiceV2} that gets the configured system properties
+ */
+public class PropertiesServiceV2 implements IPropertiesServiceV2 {
+    private static final Map<String, String> PROPERTIES = new HashMap<>();
```
p013570: I think it might be better to not cache the properties in case any of the properties change over time. In the future, if we find it is really slow then we could consider adding a daily cache or something similar.

rest-api/core-rest/src/main/java/uk/gov/gchq/gaffer/rest/ServiceConstants.java
p013570: They look like different lines to me - one is used when a Function isn't found and one is used when a Property is not found.

Issue #1477, opened 2017-10-30
Deprecate Schema & Properties Id field


GCHQDeveloper404:
Related to issues and questions raised from #1472.

Adding Schema/Props to GraphLibrary required the modification of the Id field. This would/could lead to equality issues after adding a Schema/Props as the Json stored would be different to what was given.

This ticket is to remove/depreciate Id fields and add the Id when adding to the GraphLibrary. This removes some of the need to maintain ID's from #1431.

The fields are still required when adding to GraphLibrary so that the Schema/Props keyed with same Id save on storage space.

GCHQDeveloper404:
waiting on #1472

p013570:
This will be done as part of #1472.

p013570:
Merged into develop.

PR #1480, opened 2017-10-31
Gh 1472 graphlibrary adding graphs overwrite bug

m55624:
This also includes fixes for https://github.com/gchq/Gaffer/issues/1477.


core/graph/src/test/java/uk/gov/gchq/gaffer/graph/GraphTest.java
```diff
@@ -1595,6 +1625,132 @@ public void shouldBuildGraphFromConfigAndOverrideFields() throws Exception {
...
+                .addToLibrary(true)
+                .parentStorePropertiesId("storePropertiesId1")
+                .storeProperties(graphStoreProperties)
+                .addParentSchemaIds(SCHEMA_ID_1)
+                .addSchemas(graphSchema)
+                .build();
+
+        // Then
+        assertEquals(graphId1, graph1.getGraphId());
+        assertTrue(JsonUtil.equals(library.getSchema(SCHEMA_ID_1).toJson(false), librarySchema.toJson(false)));
```
p013570: Instead of assertTrue you can use:
JsonAssert.assertEquals()

core/store/src/main/java/uk/gov/gchq/gaffer/store/schema/Schema.java
```diff
@@ -289,6 +301,12 @@ protected BaseBuilder(final Schema schema) {
             super(schema);
         }
 
+        /**
+         *
+         * @param id the id.
+         * @return this builder.
+         * @deprecated the ID should not be used. The ID should be supplied to the graph library separately
+         */
         public CHILD_CLASS id(final String id) {
```
p013570: This method needs to be annotated with "Deprecated"

p013570: There are quite a few tests that still use the schema id. If you look at the call hierarchy of this id method then you will see them all. If the test is specifically testing the deprecated id field then it should be updated.

core/graph/src/test/java/uk/gov/gchq/gaffer/operation/export/graph/handler/ExportToOtherGraphHandlerTest.java
```diff
@@ -62,19 +62,21 @@
 
     private static final String GRAPH_ID = "graphId";
     private static final String STORE_PROPS_ID = "storePropsId";
+    public static final String STORE_PROPS_ID_1 = STORE_PROPS_ID + 1;
     private static final String SCHEMA_ID = "schemaId";
+    public static final String SCHEMA_ID_2 = SCHEMA_ID + 2;
+    public static final String SCHEMA_ID_1 = SCHEMA_ID + 1;
     private static final String ID = "gaffer.store.id";
```
p013570: Can this constant be removed?

core/store/src/main/java/uk/gov/gchq/gaffer/store/StoreProperties.java
```diff
@@ -247,6 +276,7 @@ public String getId() {
      * Set the ID for the StoreProperties
      *
      * @param id the value of the ID
+     * @deprecated the ID should be supplied to the graph library separately
      */
     public void setId(final String id) {
```
p013570: This method needs to be annotated with "Deprecated".

p013570: There are still a few usages of this method. Please can we update them unless they are specifically for the old id logic. For example in StorePropertiesTest.shouldRemovePropertyWhenPropertyValueIsNull  and in FederatedGetSchemaHandlerTest.

PR #1499, opened 2017-11-06
Gh 1322 federated store add graphs schema leak

m55624:


store-implementation/federated-store/src/test/java/uk/gov/gchq/gaffer/federatedstore/FederatedStoreAuthTest.java
```diff
@@ -33,63 +35,94 @@
...
                         .graphId(EXPECTED_GRAPH_ID)
-                        .schema(expectedSchema)
-                        .storeProperties(storeProperties)
+                        .schema(schema)
+                        .storeProperties(graphStoreProperties)
                         .graphAuths("auth1")
                         .build(),
-                new Context(testUser),
-                store);
+                new Context(authUser),
```
GCHQDeveloper404: should be testUser.

m55624:
Need to check on the catch(Exception e) from the put method in GraphStorage that no Schema info can be leaked.

Issue #1421, opened 2017-10-13
If Jobs are not enabled then the /job endpoint should be disabled and removed from Swagger


p013570:
If jobs are not enabled then the JobService classes should not be added in the ApplicationConfig classes.

You may need to add a method to Graph/Store 'areJobsEnabled' to return true/false based on if the jobTracker has been configured.

p013570:
Should be done after #1513.

p013570:
Merged into develop.

PR #1502, opened 2017-11-06
gh-1421 - added JavaScript to conditionally hide Job endpoint

m607123:
Job endpoint in the Swagger UI is now "hidden" if jobs are not enabled. This is checked by firing off a GET request to check if a job-related operation is available, and if the statusText is not "success", then the endpoint is hidden.


rest-api/core-rest/src/main/webapp/lib/gaffer.js
```diff
@@ -135,6 +135,15 @@ function init(onSwaggerComplete){
           log("Loaded swagger");
               $('pre code').each(function(i,e){hljs.highlightBlock(e)});
               addExampleButtons();
+              $.get(
+                    getVersion() + '/graph/operations/uk.gov.gchq.gaffer.operation.impl.job.GetJobDetails',
+                    null,
+                    function(response) {
+                        if(200 != response.response_code) {
```
m55624: This callback will only run on success, not sure you really need the check.  If you want it is possible to use function(responseData, statusText) {if(statusText !== "Success")()}.  Better way to check the response.

m607123:
Closing while changes made

p013570:
Closing until #1513 is fixed.

PR #1511, opened 2017-11-08
add setCacheProperties(Class cacheServiceClass)

GCHQDeveloper404:
FederatedStoreProperties.java add setCacheProperties(final Class cacheServiceClass)


GCHQDeveloper404:
probably needs null check, or if null clear the key.

Issue #1485, opened 2017-11-01
Investigate possible performance improvements for the GetWalks operation


t616178:
Usage of GetWalks operation has the potential to bring lots of data in memory. The operation implementation should be checked to ensure that typical usage does not result in issues with excessive memory usage.

t616178:
Several possible avenues for performance improvement:

* pruning of orphaned edges while the elements are being added to the temporary store (reduces overall memory consumption and reduces the amount of data to process when determining the walks)
* Perform walk creation in reverse (starting from known walk destinations and tracing back to start may be more efficient than starting with the initial seed vertices).

p013570:
Merged into develop.

p013570:
The GetWalks handler now prunes the cached results after each hop to reduce the amount of memory held, however this can be disabled.

PR #1512, opened 2017-11-09
gh-1485 - Added options for pruning the data stored in the intermediate graph

t616178:
Added some functionality to allow the user to specify pruning of orphaned edges when creating a ```Walk``` using the ```GetWalks``` operation.

This has required making some changes to the ```AdjacencyMap``` class (and related classes) to increase the flexibility.

### DIFF
--- a/core/data/src/test/java/uk/gov/gchq/gaffer/data/graph/adjacency/SimpleAdjacencyMapsTest.java
+++ b/core/data/src/test/java/uk/gov/gchq/gaffer/data/graph/adjacency/SimpleAdjacencyMapsTest.java
@@ -1,0 +1,21 @@
+/*
+ * Copyright 2017 Crown Copyright
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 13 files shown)


core/data/src/test/java/uk/gov/gchq/gaffer/data/graph/adjacency/SimpleAdjacencyMapsTest.java
```diff
@@ -0,0 +1,21 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package uk.gov.gchq.gaffer.data.graph.adjacency;
+
+public class SimpleAdjacencyMapsTest {
```
p013570: Please can you add some tests in this class?

core/operation/src/main/java/uk/gov/gchq/gaffer/operation/impl/GetWalks.java
```diff
@@ -54,6 +54,7 @@
     private Iterable<? extends EntityId> input;
     private Map<String, String> options;
     private Integer resultsLimit = 1000000;
+    private boolean prune = false;
```
p013570: Please can you make sure there is an issue in gaffer-tools to add this prune field to the python shell.

p013570: This should not be exposed to users. It should be in the handler.

p013570:

PR #1515, opened 2017-11-13
Gh 1490 improve logic and documentation add elements from hdfs

m55624:


gaffer01:
Do we need an issue about to gaffer-doc to improve the documentation of min and max there, and note that setting the number of reducers has been deprecated?

m55624:
closing for fixes.

PR #1516, opened 2017-11-13
gh-1485 - Added options for pruning the data stored in the intermedite graph

t616178:


core/store/src/main/java/uk/gov/gchq/gaffer/store/operation/handler/GetWalksHandler.java
```diff
@@ -45,23 +48,27 @@
...
- * Currently the handler only supports creating {@link Walk}s which contain
- * {@link Edge}s.
- * </p>
+ *
+ * Currently the handler only supports creating {@link Walk}s which contain {@link Edge}s.
+ *
+ * This operation handler can be modified by supplying an operationDeclarations.json
+ * file in order to limit the maximum number of hops permitted.
  */
 public class GetWalksHandler implements OutputOperationHandler<GetWalks, Iterable<Walk>> {
```
p013570: Please can you add a some information to explain when pruning should be enabled and the effects it will have.

core/data/src/main/java/uk/gov/gchq/gaffer/data/graph/adjacency/PrunedAdjacencyMaps.java
```diff
@@ -0,0 +1,96 @@
...
+            final AdjacencyMap<V, E> prev = maps.get(maps.size() - 1);
+
+            final Set<V> prevDestinations = prev.getAllDestinations();
+
+            final List<V> verticesToRemove = new ArrayList<>();
+
+            // Build up the list of destination vertices in the previous map which
+            // do not connect to any source vertices in the current map
+            for (final V dest : prevDestinations) {
+                if (!curr.getAllSources().contains(dest)) {
```
p013570: Don't call curr.getAllSources() every time as it create lots of objects.

core/data/src/main/java/uk/gov/gchq/gaffer/data/graph/adjacency/PrunedAdjacencyMaps.java
```diff
@@ -0,0 +1,96 @@
...
+        }
+    }
+
+    @Override
+    public List<AdjacencyMap<V, E>> asList() {
+        return adjacencyMaps;
+    }
+
+    @Override
+    public String toString() {
```
p013570: The toString method could be added to the parent class and then it could be removed from here and SimpleAdjacencyMaps

core/data/src/main/java/uk/gov/gchq/gaffer/data/graph/adjacency/PrunedAdjacencyMaps.java
t616178: Done.

core/store/src/main/java/uk/gov/gchq/gaffer/store/operation/handler/GetWalksHandler.java
t616178: Added more description to the class level Javadoc.

Issue #1469, opened 2017-10-25
SplitStore from Iterable of Elements


gaffer01:
The only way of adding split points to an Accumulo table is to use `SampleDataForSplitPoints` followed by `SplitStore`. This is difficult to use if the source data is only available as an iterable, e.g. if it's on Kafka.

We should create a way for a user to create some split points based on an iterable of elements.

p013570:
# Migration Steps
There is no breaking change and no immediate migration required. However, the SplitStore operation has been deprecated and we recommend you update to use SplitStoreFromFile when possible.

p013570:
Merged into develop.

PR #1517, opened 2017-11-13
gh-1469 - added sample elements for split points operation

p013570:
There is a bit of extra complexity in the handler as I have attempted to make an abstract class that is generic so that it is easy for different Stores to implement it. The operation SampleElementsForSplitPoints returns a List<T> where T is the split point type. For Accumulo and HBase this is a String, but other stores may use a different type. For Accumulo and HBase the split points are Base64 encoded.


core/store/src/main/java/uk/gov/gchq/gaffer/store/operation/handler/AbstractSampleElementsForSplitPointsHandler.java
```diff
@@ -0,0 +1,118 @@
...
+import java.util.stream.Stream;
+
+public abstract class AbstractSampleElementsForSplitPointsHandler<T, S extends Store> implements OutputOperationHandler<SampleElementsForSplitPoints<T>, List<T>> {
+    public static final int MAX_SAMPLED_ELEMENTS_DEFAULT = 10000000;
+
+    private int maxSampledElements = MAX_SAMPLED_ELEMENTS_DEFAULT;
+
+    @Override
+    public List<T> doOperation(final SampleElementsForSplitPoints<T> operation, final Context context, final Store store) throws OperationException {
+        S typedStore = (S) store;
```
gaffer01: Should be final I think

store-implementation/hbase-store/src/main/java/uk/gov/gchq/gaffer/hbasestore/operation/handler/SampleElementsForSplitPointsHandler.java
```diff
@@ -0,0 +1,74 @@
...
+            numSplits = getNumAccumuloSplits(store);
+        }
+        return numSplits;
+    }
+
+    private int getNumAccumuloSplits(final HBaseStore store) {
+        int numberRegions;
+        try {
+            numberRegions = store.getConnection().getAdmin().getTableRegions(store.getTableName()).size();
+            LOGGER.debug("Number of tablet servers is {}", numberRegions);
```
gaffer01: Pedantic but should be "region servers" not "tablet servers"

gaffer01:
I've made a couple of minor comments.

Main comment is how are the list of split points applied to a table? There is a `SplitStore` operation but that takes a path to a file of split points, whereas the new operation added here outputs a list of split points. Is the intention that the split points will be written to file and then read by the `SplitStore` operation? Should we add the ability to split a store directly from a list of split points?

p013570:
@gaffer01 - added made your suggested changes. I've deprecated SplitStore and created SplitStoreFromFile and SplitStoreFromIterable. The SplitStoreFromFile reads the file locally or in HDFS then delegates to SplitStoreFromIterable.

Issue #1417, opened 2017-10-13
Add configurable logo to REST API Swagger UI


p013570:
This will help to brand the REST API for different systems

p013570:
Merged into develop.

PR #1560, opened 2017-11-28
Gh-1417 UI Configurable Logo

m607123:
Added the ability to customise the logo and favicons individually, as well as the logo's hyperlink.
If either or both of the favicons are not provided, it will just default to using the logo image.


example/road-traffic/road-traffic-demo/pom.xml
```diff
@@ -83,6 +83,18 @@
                                     href='http://www.nationalarchives.gov.uk/doc/open-government-licence/version/3'
                                     &gt;Open Government Licence&lt;/a&gt;.
                                 </gaffer.properties.app.description>
+                                <gaffer.properties.app.logo.link>
+                                    https://github.com/gchq/Gaffer
+                                </gaffer.properties.app.logo.link>
+                                <gaffer.properties.app.logo.src>
+                                    images/iconCircle.png
```
p013570: Please could you rename the iconCircle.png file to logo.png and move it into src/main/webapp/images/

example/road-traffic/road-traffic-demo/pom.xml
```diff
@@ -83,6 +83,18 @@
...
+                                </gaffer.properties.app.logo.src>
+                                <gaffer.properties.app.logo.favicon.small>
```
p013570: Please could you add Gaffer favicon-16x16 and favicon-32x32?

example/road-traffic/road-traffic-demo/pom.xml
p013570: Please could you remove the images in dependencies/swagger/images that we are not using.

rest-api/core-rest/src/main/webapp/lib/gaffer.js
```diff
@@ -208,6 +211,27 @@ function updateDocUrl(properties) {
...
 
+function updateLogoLink(properties) {
+    updateElementWithId('logo', 'gaffer.properties.app.logo.link', properties, function(value, id) {
+        $('#' + id).attr("href", value);
+    });
+}
+
+function updateLogo(properties) {
+    updateElementWithId('logo', 'gaffer.properties.app.logo.src', properties, function(value, id) {
+        $('#' + id).prepend("<img class='logo__img' alt='swagger' height='30' width='30' src='" + value + "'/>");
```
p013570: The Gaffer logo you have as a default is not square so setting height and width to 30 squashes it. Perhaps you should adjust the logo so it is square.

p013570: If you switch to version v2 in the Swagger ui, then this causes the logo to be added multiple times. We also had a problem like that with the banner, but that has since been fixed.

p013570:

m607123:
Changes made. Thanks to @t616178 for the new icons.

Issue #1527, opened 2017-11-17
AddElementsFromHdfs deduplicates Elements that have aggregation disabled and identical keys


supamanda:
Hi again,

I'm still looking at the netflow data I've mentioned in other questions. I specified in the schema that Netflow edges are not to be aggregated and each edge represents a single netflow event in the original csv with probably different properties.

When I use the startTime property as the timestamp (I have to nano precision, but storing at millis), only one record is stored in accumulo even if I have >1 record with the same source+destination+timestamp. I've also tried using the default timestamp and this is resulting in more missing records. I've checked the element generator, and all the elements are being generated correctly.

Here is some example records from the CTU13 dataset (previously mentioned) that demonstrates the problem.
These two records have the same source (::) and destination (ff02::1:ff97:a3eb), and the timestamps both map to 2011/08/12 15:59:53.173 (converted to a long), and only the first one was in accumulo
```
[2011/08/12 15:59:53.173326,0.0,ipv6-icmp,::,130,   ->,ff02::1:ff97:a3eb,null,MRQ,0,null,1,86,86,flow=Background]
[2011/08/12 15:59:53.173341,90.500267,ipv6-icmp,::,135,   ->,ff02::1:ff97:a3eb,null,NNS,0,null,6,468,468,flow=Background]
```
These two records have the same source (::) and destination (ff02::1:ff9d:312f), and the timestamps map to different values: 2011/08/12 16:09:48.275 and 2011/08/12 16:09:48.276, and both records are in accumulo
```
[2011/08/12 16:09:48.275895,0.500047,ipv6-icmp,::,135,   ->,ff02::1:ff9d:312f,null,NNS,0,null,2,156,156,flow=Background]
[2011/08/12 16:09:48.276904,0.0,ipv6-icmp,::,130,   ->,ff02::1:ff9d:312f,null,MRQ,0,null,1,86,86,flow=Background]
```

Do you have any suggestions about how I can change my config / ingest to ensure that all my events are stored separately?

Thanks,
Amanda

gaffer01:
The `timestamp` isn't necessary. Specifying the timestamp as the startTime causes the same records to be deduped in Accumulo as you found. We'd recommend not using the timestamp (using it in situations where it is safe to do so might cause very minor performance benefits).

If by "default timestamp" you mean not specifying the timestamp field then that is odd, as that should work. The following code shows how the same edge will appear multiple times in the data if no aggregation is specified:

    Schema schema = new Schema.Builder()
[...]

PR #1563, opened 2017-11-29
gh-1527 - added test for adding duplicate elements with the same time…

p013570:
…stamp


p013570:
Closing whilst bug #1565 is fixed.

PR #1568, opened 2017-11-29
Gh 1565 hbase timestamps

p013570:


p013570:
Closing while tests are fixed.

core/common-util/src/main/java/uk/gov/gchq/gaffer/commonutil/LongUtil.java
```diff
@@ -0,0 +1,81 @@
...
+     * <li>the result bytes would be: [e,f,h,i,j,k,l,m]</li>
+     * </ul>
+     *
+     * @return the time based random number
+     */
+    public static long getTimeBasedRandom() {
+        long time = System.currentTimeMillis();
+        int randomVal = RANDOM.nextInt();
+
+        return getRandom(time, randomVal);
```
gaffer01: This can be replaced by  `return randomVal | time << 32`, and the whole of the method below can be deleted.

store-implementation/hbase-store/README.md
```diff
@@ -185,7 +187,20 @@ See [the aggregation example](https://gchq.github.io/gaffer-doc/getting-started/
...
+
+For elements with the same Key (group, vertex/source/destination/drection, groupBy), with aggregation disabled, the timestamp value must be unique.
+If it is not unique then elements will be deduplicated and deleted, even if some properties are different.
+
+If you choose to set the timestampProperty in the schema and populate the timestamps yourself then you must ensure the timestamps are unique.
+If you don't set the timestamp values then HBase will attempt to create a unique timestamp value.
+There is a very very small chance that it will fail to create a completely unique value and data could be lost.
+The auto generated timestamp is made up partly of the current system time and party with a random number.
+So if you ingest lots of similar elements at exactly the same time (within the same millisecond) the chance of a non-unique value is higher.
+
```
gaffer01: I'm not sure we need to say this. If you're ingesting the same elements from the same JVM then it's the same instance of SplittableRandom being used and this doesn't give repeated elements until at least 2^64 applications. We're using nextInt not nextLong so the chance of a repeat is higher but still very low.

If you're ingesting elements from different JVMs then different instances of SplittableRandom are used but the default constuctor of that says it varies "across program invocations".

I would replace the end of this paragraph with "If you don't set the timestamp values then HBase will attempt to create a unique timestamp value. This value is created from a combination of the current time and a randomly generated number. The chances of this not being unique are negligible."

p013570:
This also fixes #1575.

Issue #1436, opened 2017-10-16
GraphFrames integration


t616178:
Currently, the Accumulo store supports create DataFrames from stored data. It should be straightforward to also create GraphFrames which would allow for analysis of graphs using the methods available on the GraphFrames API.

p013570:
Merged into develop.

PR #1571, opened 2017-11-30
Gh 1507 graphframe element generator

t616178:
Initial ideas for converting a GraphFrame (or a DataFrame) to an Iterable of Element objects. The implementation here uses two operations to achieve this:

* GraphFrameToIterable
* GenerateElements (with a RowToElementGenerator)

Unfortunately it's not straightforward to use a generator as is, since the bulk of the generators in Gaffer at the moment are geared up for converting Iterable<OBJ> -> Iterable<Element>. So in order to use the existing generator-based approach, the GraphFrame must be first converted into an Iterable<Row>.

There are a few caveats here based on some limitations of the use of DataFrames:

* All elements are pulled into a single DataFrame/GraphFrame, so it will not always be possible to distinguish between edge properties and entity properties. If an edge group and an entity group are present that share a property name, it will not always be possible to reconstruct the original elements - these properties must be of the same type (this is noted in the AccumuloStoreRelation javadoc).
* The implementation relies on the use of several reserved words (src, dst, id, vertex etc). If any of these words are used as property names, then this approach will not work.

Any suggestions would be welcome - should we stick with using a Generator approach or shall I wrap these into a single operation?

This should be merged after gh-1436.


library/spark/spark-accumulo-library/src/main/java/uk/gov/gchq/gaffer/sparkaccumulo/operation/handler/graphframe/GraphFrameAsIterableHandler.java
```diff
@@ -0,0 +1,47 @@
...
+        if (null == operation.getInput()) {
+            throw new OperationException("Input must not be null");
+        }
+
+        final GraphFrame graphFrame = operation.getInput();
+
+        final Iterator<Row> vertices = graphFrame.vertices().toLocalIterator();
+        final Iterator<Row> edges = graphFrame.edges().toLocalIterator();
+
+        return () -> Iterators.concat(vertices, edges);
```
p013570: I haven't checked, but it looks like you won't be able to iterator around the iterable more than once. You will need something like:

```java
return () -> {
    final Iterator<Row> vertices = graphFrame.vertices().toLocalIterator();
    final Iterator<Row> edges = graphFrame.edges().toLocalIterator();

    return Iterators.concat(vertices, edges);
};
```

Does the iterable need to be closeable?  Is there any way to close the graph frame?

library/spark/spark-accumulo-library/src/main/java/uk/gov/gchq/gaffer/sparkaccumulo/operation/handler/graphframe/GetGraphFrameOfElementsHandler.java
```diff
@@ -0,0 +1,90 @@
...
+ * @see uk.gov.gchq.gaffer.sparkaccumulo.operation.handler.dataframe.GetDataFrameOfElementsHandler
+ */
+public class GetGraphFrameOfElementsHandler implements OutputOperationHandler<GetGraphFrameOfElements, GraphFrame> {
+
+    @Override
+    public GraphFrame doOperation(final GetGraphFrameOfElements operation, final Context context, final Store store) throws OperationException {
+        return doOperation(operation, context, (AccumuloStore) store);
+    }
+
+    public GraphFrame doOperation(final GetGraphFrameOfElements operation, final Context context,
```
p013570: I think we may need to add some validation to ensure the elements don't have properties with names that will conflict with 'vertex', 'id', 'group'. Similarly with SchemaToStructTypeConverter constants.

library/spark/spark-accumulo-library/src/main/java/uk/gov/gchq/gaffer/sparkaccumulo/operation/handler/graphframe/GetGraphFrameOfElementsHandler.java
```diff
@@ -0,0 +1,90 @@
...
+        // Create a DataFrame of Edges - must add an "id" column which we fill with
+        // the row number. We add a partitionBy on group to avoid creating a single
+        // partition for all data.
+        final Dataset<Row> edges = sparkSession.sql("select * from elements where group in " + edgeGroups)
+                .withColumn("id", functions.row_number().over(Window.orderBy("group").partitionBy("group")));
+
+        // Create a DataFrame of Entities
+        final Dataset<Row> entities = sparkSession.sql("select * from elements where group in " + entityGroups);
+
+        return GraphFrame.apply(entities.withColumnRenamed("vertex", "id"), edges);
```
p013570: Have we checked this is ok with multiple Entities with the same vertex? What happens?

library/spark/spark-library/pom.xml
```diff
@@ -132,6 +138,11 @@
             <artifactId>spark-sql_${scala.version}</artifactId>
             <version>${spark.version}</version>
         </dependency>
+        <dependency>
```
p013570: Please can you update the NOTICES file.

p013570:

PR #1583, opened 2017-12-06
Gh 1564 hbase reducer

p013570:


library/hdfs-library/src/main/java/uk/gov/gchq/gaffer/hdfs/operation/reducer/GafferReducer.java
```diff
@@ -0,0 +1,112 @@
+/*
+ * Copyright 2016 Crown Copyright
```
t616178: Copyright header should be updated - same goes for the rest of the new files in this PR.

PR #1589, opened 2017-12-07
Gh 1547 create named view resolver graphhook

m55624:


core/data/src/main/java/uk/gov/gchq/gaffer/data/elementdefinition/view/View.java
```diff
@@ -386,6 +386,10 @@ public CHILD_CLASS json(final byte[]... jsonBytes) throws SchemaException {
         @Override
         @JsonIgnore
         public CHILD_CLASS merge(final View view) {
+            if (view instanceof NamedView && !(getThisView() instanceof NamedView)) {
```
p013570: The View class should not reference the NamedView class. It causes a cyclic dependency between the 2 classes. Perhaps you could create a method:
```java
public boolean canBeMerged(final View destinationView, final View sourceView);
```

In named view you could override it to to check the destinationView is a NamedView.

Then in the merge method here you could do:
```java
this.canBeMerged(this, view) && view.canBeMerged(this, view);
```

That would check the View is happy for them to be merged and the NamedView class is happy.

core/graph/src/main/java/uk/gov/gchq/gaffer/graph/hook/NamedViewResolver.java
```diff
@@ -0,0 +1,90 @@
...
+import uk.gov.gchq.gaffer.operation.Operations;
+import uk.gov.gchq.gaffer.operation.graph.OperationView;
+import uk.gov.gchq.gaffer.store.Context;
+import uk.gov.gchq.gaffer.store.operation.handler.named.cache.NamedViewCache;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * A {@link GraphHook} to resolve named operations.
```
p013570: Update javadoc.

core/graph/src/main/java/uk/gov/gchq/gaffer/graph/hook/NamedViewResolver.java
```diff
@@ -0,0 +1,90 @@
...
+
+    @Override
+    public <T> T onFailure(final T result, final OperationChain<?> opChain, final Context context, final Exception e) {
+        return null;
+    }
+
+    private void resolveViewsInOperations(final Operations<?> operations) {
+        final List<Operation> updatedOperations = new ArrayList<>(operations.getOperations().size());
+        for (final Operation operation : operations.getOperations()) {
+            if (operation instanceof OperationView) {
```
p013570: We recently just fixed a bug in NamedOperationResolver. Please can you apply the same recursion: https://github.com/gchq/Gaffer/pull/1588. I.e check if the operation is an instanceof Operations.

core/graph/src/main/java/uk/gov/gchq/gaffer/graph/hook/NamedViewResolver.java
```diff
@@ -0,0 +1,90 @@
...
+                if (((OperationView) operation).getView() instanceof NamedView) {
+                    final NamedView resolvedNamedView = resolveNamedViewInOperation((NamedView) ((OperationView) operation).getView());
+
+                    final NamedView mergedNamedView = new NamedView.Builder().name(resolvedNamedView.getName()).merge(resolvedNamedView).merge(((OperationView) operation).getView()).build();
```
p013570: Perhaps add some line breaks.

core/graph/src/main/java/uk/gov/gchq/gaffer/graph/hook/NamedViewResolver.java
```diff
@@ -0,0 +1,90 @@
...
+            updatedOperations.add(operation);
+        }
+        operations.getOperations().clear();
+        operations.getOperations().addAll((List) updatedOperations);
+    }
+
+    private NamedView resolveNamedViewInOperation(final NamedView opView) {
+        NamedView.Builder newNamedView;
+        try {
+            newNamedView = new NamedView.Builder().name(opView.getName()).merge(cache.getNamedView(opView.getName()));
```
p013570: I don't think you need to create a new NamedView object here. Can you create a View object instead?

core/graph/src/main/java/uk/gov/gchq/gaffer/graph/hook/NamedViewResolver.java
```diff
@@ -0,0 +1,90 @@
...
+            for (final String name : opView.getMergedNamedViewNames()) {
+                newNamedView.merge(cache.getNamedView(name));
```
p013570: Before you merge cache.getNamedView(name) in, you should recursively check it doesn't refer to any other named views.

core/graph/src/main/java/uk/gov/gchq/gaffer/graph/hook/NamedViewResolver.java
m55624: I need to keep the name, so when the merge is run it will add the name to the mergedNamedViewNames, otherwise it will still work but the user won't be able to keep track of what the current NamedView consists of.

core/graph/src/main/java/uk/gov/gchq/gaffer/graph/hook/NamedViewResolver.java
m55624: I don't think this is necessary.  If a NamedView is nested, then the parent will contain the NamedView name of the merged NamedViews, even if they are nested on several levels.  There is a test in to check.

PR #1591, opened 2017-12-12
Gh 1436 graphframes implementation

t616178:


core/data/src/main/java/uk/gov/gchq/gaffer/data/elementdefinition/view/View.java
```diff
@@ -143,6 +144,15 @@ public boolean hasEdgeFilters() {
                 || hasEdgeFilters(ViewElementDefinition::hasPreAggregationFilters);
     }
 
+    @JsonIgnore
+    public Map<String, ViewElementDefinition> getElements() {
```
p013570: This is only used in one place. It might be better to put this where it is used, so we don't encourage others to use it without realising a new hash map is created with every call.

library/spark/spark-accumulo-library/src/main/java/uk/gov/gchq/gaffer/sparkaccumulo/operation/handler/dataframe/GetDataFrameOfElementsHandler.java
```diff
@@ -44,6 +45,14 @@
         } else {
             operationOptions = new HashMap<>();
         }
+
+        for (final SchemaEntityDefinition entityDef : store.getSchema().getEntities().values()) {
```
p013570: This could be annoying for people. Could you use the keyword VERTEX instead? Do you also need to check GROUP, SOURCE, DESTINATION and DIRECTED too?

library/spark/spark-library/src/main/java/uk/gov/gchq/gaffer/spark/operation/graphframe/GetGraphFrameOfElements.java
```diff
@@ -0,0 +1,152 @@
...
+    public void setDirectedType(final DirectedType directedType) {
+        this.directedType = directedType;
+    }
+
+    @Override
+    public ValidationResult validate() {
+        final ValidationResult result = Output.super.validate();
+
+        if (!view.hasEntities() || !view.hasEdges()) {
+            result.addError("Cannot create a Graphframe unless the View contains both edges and entities.");
```
p013570: Why does it need entities? It is possible to create Gaffer graphs without any entities at all.

library/spark/spark-library/src/main/java/uk/gov/gchq/gaffer/spark/operation/graphframe/GetGraphFrameOfElements.java
```diff
@@ -0,0 +1,152 @@
...
+        }
+
+        final List<String> reservedProperties = Lists.newArrayList("id", "vertex", "group");
```
p013570: See other comment. Could we change these to ID, VERTEX, GROUP?

core/data/src/main/java/uk/gov/gchq/gaffer/data/elementdefinition/view/View.java
t616178: Ok - fixed now.

library/spark/spark-accumulo-library/src/main/java/uk/gov/gchq/gaffer/sparkaccumulo/operation/handler/dataframe/GetDataFrameOfElementsHandler.java
t616178: This issue builds on the implementation of the GetDataFrameOfElements operation which relies on these keywords to create Rows (for the DataFrame) from the Elements in the Gaffer graph. Any changes applied here would also need to be made to the GetDataFrameOfElements operation.

library/spark/spark-library/src/main/java/uk/gov/gchq/gaffer/spark/operation/graphframe/GetGraphFrameOfElements.java
t616178: See my comment on your other comment.

library/spark/spark-library/src/main/java/uk/gov/gchq/gaffer/spark/operation/graphframe/GetGraphFrameOfElements.java
t616178: It doesn't *need* entities strictly speaking, but a GraphFrame object created without any entity objects will not be query-able due to the internal representation used by the Graphframes library. If a user has a graph (or creates a query) which has only Edges, they would expect to be able to carry out operations such as PageRank or ShortestPath, as neither of these algorithms would require the Entities to be present. The options are: prevent a GraphFrame from being created without Entities being present (as I have done here) *or* add dummy entities to the GraphFrames object when the user does not specify any entities (this will enable users to run more queries on the GraphFrames object). Do you have any preference?

core/data/src/main/java/uk/gov/gchq/gaffer/data/element/ReservedPropertyNames.java
```diff
@@ -0,0 +1,58 @@
...
+import java.util.stream.Collectors;
+
+/**
+ * Enumeration of all protected property names.
+ */
+public enum ReservedPropertyNames {
+
+    VERTEX("vertex"),
+    SOURCE("src"),
+    DESTINATION("dst"),
```
p013570: I'm not sure about putting this class in the data module and making it part of core Gaffer. Those names are only reserved when using spark. However, it would be useful for users to be aware of these reserved names when they first create a Gaffer schema even before they thing about using spark.

I would be happy to keep this class here if we included things like: "vertex", "VERTEX", "src", "SRC", "source", "SOURCE", etc. Otherwise we will soon end up having another library that needs to use a slightly different keywords. However, if we are saying these property names are reserved then we really should validate schemas and tell users not to use these names at all. We can't really fail schema validation as we would break any existing systems using these names. Perhaps we include some validation in the core schema validation for these property names, but just log warnings if any of these names are used. In version 2.0.0 we could change the warning to be an error.

library/spark/spark-accumulo-library/pom.xml
```diff
@@ -65,6 +67,43 @@
 
     <build>
         <plugins>
+            <plugin>
+                <groupId>net.alchim31.maven</groupId>
```
p013570: What is this plugin used for? It will need to be added to the NOTICES file if we need it.

library/spark/spark-library/src/main/java/uk/gov/gchq/gaffer/spark/operation/dataframe/converter/schema/SchemaToStructTypeConverter.java
```diff
@@ -152,35 +153,40 @@ private void buildSchema() {
...
-                                propertyType = converter.convertedType();
-                                converterByProperty.put(property, converter);
-                                structFieldList.add(new StructField(property, propertyType, true, Metadata.empty()));
-                                LOGGER.info("Property {} of type {} will be converted by {} to {}",
-                                        property,
-                                        propertyClass,
-                                        converter.getClass().getName(),
-                                        propertyType);
-                                break;
+                if (!ReservedPropertyNames.contains(property)) {
```
p013570: Please can we make sure this is documented somewhere so users are aware that the reserved properties are skipped.

library/spark/spark-library/src/main/java/uk/gov/gchq/gaffer/spark/operation/dataframe/ConvertElementToRow.java
```diff
@@ -84,6 +85,8 @@ public Row apply(final Element element) {
                     final Object value = element.getProperties().get(property);
                     if (null == value) {
                         fields.appendElem(null);
+                    } else if (ReservedPropertyNames.contains(property)) {
```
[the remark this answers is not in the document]
t616178: Done.

core/data/src/main/java/uk/gov/gchq/gaffer/data/element/ReservedPropertyNames.java
t616178: Done.

library/spark/spark-accumulo-library/pom.xml
t616178: Done.

library/spark/spark-library/src/main/java/uk/gov/gchq/gaffer/spark/operation/dataframe/converter/schema/SchemaToStructTypeConverter.java
t616178: Added documentation to the class describing the new behaviour.

library/spark/spark-library/src/main/java/uk/gov/gchq/gaffer/spark/operation/graphframe/GetGraphFrameOfElements.java
```diff
@@ -110,21 +104,7 @@ public ValidationResult validate() {
...
-        for (final ViewElementDefinition viewElementDefinition : viewElementDefinitions) {
-            if (null != viewElementDefinition && null != viewElementDefinition.getProperties()) {
-                final Collection<String> repeatedProperties = CollectionUtils.intersection(viewElementDefinition.getProperties(), GetDataFrameOfElements.RESERVED_FIELDS);
-                if (!repeatedProperties.isEmpty()) {
-                    result.addError("Cannot create a GraphFrame using the current View - the properties: [" +
-                            String.join(", ", repeatedProperties) + "] are reserved and must not be in the View."
-                    );
-                }
-            }
+            result.addError("Cannot create a GraphFrame unless the View contains edges.");
```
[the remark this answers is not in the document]
t616178: Updated to allow the creation of GraphFrames with empty edge DataFrames, though some of the analytics will not work correctly if no edges are present - perhaps we need to validate the GraphFrame when executing specific algorithms?

Issue #1607, opened 2018-01-02
Time based random number should always be positive


p013570:
The time based random number in LongUtil can sometimes be negative. We should enforce the number is always positive.

This was introduced in v1.2.0-RC1.

p013570:
Merged into develop.

PR #1608, opened 2018-01-02
gh-1607 - time based random will now always be positive

p013570:


p013570:
I'm going to merge this into develop, to fix develop. @gaffer01 if you have any comments on it I will make the changes separately.

PR #1626, opened 2018-01-17
Gh 1542 named views

m55624:


m55624:
```This example explains how to configure your Gaffer Graph to allow NamedViews to be used. 
Named views enable encapsulation of a View, or multiple Views, into a new single NamedView.
A NamedView can be used in conjunction with other Views.
There are various possible uses for NamedViews, including:
 * Making it simpler to reuse frequently used Views.
 * Share commonly used Views with other users.
 
In addition to the NamedView there are a set of operations which manage named views (AddNamedView, GetAllNamedViews, DeleteNamedView).

## Configuration
You will need to configure what cache to use for storing NamedViews. For more information on the cache service, see [Cache](cache.md).

## Using Named Views
We will now go through some examples using NamedViews.

We will use the same basic schema and data from the first developer walkthrough.

Start by creating your user instance and graph:

${START_JAVA_CODE}
${USER_SNIPPET}
${END_CODE}

${START_JAVA_CODE}
${GRAPH_SNIPPET}
${END_CODE}


Then add a NamedView to the cache with the AddNamedView operation:

${START_JAVA_CODE}
${ADD_NAMED_VIEW_SNIPPET}
${END_CODE}

For the next step we need to create a GetElements operation and use
the NamedView previously added to the cache, using the name specified.
After this we can execute it:

${START_JAVA_CODE}
${CREATE_GET_ELEMENTS_OPERATION_SNIPPET}
${END_CODE}

${START_JAVA_CODE}
${EXECUTE_GET_ELEMENTS_OPERATION_SNIPPET}
${END_CODE}

The results are:

```
${OPERATION_USING_NAMED_VIEW_RESULTS}
```

NamedViews can also take parameters, allowing values to be configured by the user when the NamedView is used.
When adding a NamedView with parameters the View must be specified as a JSON string, with
parameter names enclosed with '${' and '}'. For each parameter, a ViewParameterDetail object must be created which gives a description, a class type
and an optional default for the Parameter, and also indicates whether the parameter must be provided (ie. there is no default).

The following code adds a NamedView with a 'countIsMoreThan' parameter that allows the minimum edge count for the View to be set, along
with a 'selectionParam' parameter that allows the user to set the selection for the View:

${START_JAVA_CODE}
${ADD_NAMED_VIEW_WITH_PARAMETERS_SNIPPET}
${END_CODE}

A view can then be created, with a value provided for the 'countIsMoreThan' and 'selectionParam' parameter:

${START_JAVA_CODE}
${CREATE_NAMED_VIEW_WITH_PARAMETERS_SNIPPET}
${END_CODE}

and executed:

${START_JAVA_CODE}
${EXECUTE_GET_ELEMENTS_OPERATION_WITH_PARAMETERS_SNIPPET}
${END_CODE}

giving these results:

```
${OPERATION_USING_NAMED_VIEW_WITH_PARAMETERS_RESULTS}
``````

File: library/cache-library/jcs-cache-service/src/main/resources/cache.ccf [hunk 1]
```diff
@@ -16,6 +16,9 @@ jcs.auxiliary.DC.attributes.OptimizeOnShutdown=true
 # Region for named operations
 jcs.region.namedOperationsRegion=DC
 
+# Region for named views
+jcs.region.namedViewRegion=DC
```
p013570: Please could you add a migration step to the issue to tell users that if they want to use NamedViews they will need to add this to any custom cache.ccf files they have. Also, for backwards compatibility please can you make sure you that if a user doesn't set this, then the app can handle it and doesn't crash.

core/graph/src/main/java/uk/gov/gchq/gaffer/graph/Graph.java
```diff
@@ -860,6 +864,16 @@ private void updateGraphHooks(final GraphConfig config) {
                     config.getHooks().add(0, new NamedOperationResolver());
                 }
             }
+            boolean hasNamedViewHook = false;
+            for (final GraphHook graphHook : config.getHooks()) {
```
p013570: You could merge this for loop and the one before together to avoid having to iterate over the hooks twice.

core/graph/src/main/java/uk/gov/gchq/gaffer/graph/Graph.java
```diff
@@ -860,6 +864,16 @@ private void updateGraphHooks(final GraphConfig config) {
...
+                if (NamedViewResolver.class.isAssignableFrom(graphHook.getClass())) {
+                    hasNamedViewHook = true;
+                    break;
+                }
+            }
+            if (!hasNamedViewHook) {
+                config.getHooks().add(0, new NamedViewResolver());
```
p013570: I think we need the NamedViewResolver to be after the NamedOperationResolver. After named operations have been resolved, we then need to resolve any named views that exist within the named operations.

core/graph/src/main/java/uk/gov/gchq/gaffer/graph/hook/NamedViewResolver.java
```diff
@@ -0,0 +1,110 @@
...
+        return result;
+    }
+
+    @Override
+    public <T> T onFailure(final T result, final OperationChain<?> opChain, final Context context, final Exception e) {
+        return result;
+    }
+
+    private void resolveViewsInOperations(final Operations<?> operations) {
+        final List<Operation> updatedOperations = new ArrayList<>(operations.getOperations().size());
```
p013570: I don't think we need this updatedOperations list. We are just modifying the existing operations.

core/data/src/main/java/uk/gov/gchq/gaffer/data/elementdefinition/view/NamedView.java
```diff
@@ -0,0 +1,196 @@
...
+        @Override
+        public CHILD_CLASS merge(final View view) {
+            if (null != view) {
+                if (view instanceof NamedView) {
+                    NamedView namedViewInstance = (NamedView) view;
+                    if (null != namedViewInstance.getName() && !namedViewInstance.getName().isEmpty()) {
+                        if (null == self().getElementDefs().getName() || self().getElementDefs().getName().isEmpty()) {
+                            self().name(namedViewInstance.getName());
+                        } else {
+                            if (self().getElementDefs().getName() != ((NamedView) view).getName()) {
```
p013570: You need to use .equals for string comparisons.

core/graph/src/main/java/uk/gov/gchq/gaffer/graph/hook/NamedViewResolver.java
```diff
@@ -0,0 +1,110 @@
...
+            View resolvedCachedView = cachedNamedView.getView(namedView.getParameters());
+            newNamedView = new NamedView.Builder()
+                    .name(namedView.getName())
+                    .merge(resolvedCachedView);
+
+            if (resolvedCachedView instanceof NamedView && null != ((NamedView) resolvedCachedView).getMergedNamedViewNames()) {
+                for (final String name : ((NamedView) resolvedCachedView).getMergedNamedViewNames()) {
+                    final NamedViewDetail nestedCachedNamedView = cache.getNamedView(name);
+                    final NamedView resolvedNestedCacheNamedView = new NamedView.Builder().name(null).merge(nestedCachedNamedView.getView(namedView.getParameters())).build();
+                    newNamedView.merge(resolvedNestedCacheNamedView);
```
p013570: I think you need another layer of recursion to check that the resolved named view doesn't need further resolving.

core/data/src/main/java/uk/gov/gchq/gaffer/data/elementdefinition/view/NamedView.java
```diff
@@ -0,0 +1,196 @@
...
+        public CHILD_CLASS json(final byte[]... jsonBytes) throws SchemaException {
+            return json(NamedView.class, jsonBytes);
+        }
+
+        @Override
+        public CHILD_CLASS merge(final View view) {
+            if (null != view) {
+                if (view instanceof NamedView) {
+                    NamedView namedViewInstance = (NamedView) view;
+                    if (null != namedViewInstance.getName() && !namedViewInstance.getName().isEmpty()) {
```
p013570: This is a bit tricky to read - perhaps use StringUtils.isEmpty(namedViewInstance.getName()) to simplify your logic.

core/graph/src/main/java/uk/gov/gchq/gaffer/graph/hook/NamedViewResolver.java
m55624: This won't be necessary, as NamedViews are merged all the lists are concatenated so the top level NamedView list of mergedNamedViewNames will contain all child NamedView names.  Tests have been added to check expected behaviour is correct.

core/graph/src/test/java/uk/gov/gchq/gaffer/graph/GraphTest.java
```diff
@@ -1438,7 +1438,7 @@ public void shouldAddHookAndGetGraphHooks() throws Exception {
     }
 
     @Test
-    public void shouldAddNamedOperationResolverHookFirst() throws Exception {
+    public void shouldAddNamedOperationResolverHookAfterNamedOperationResolver() throws Exception {
```
p013570: Minor comment, but I think this method name should be:
shouldAddNamedViewResolverHookAfterNamedOperationResolver

core/data/src/main/java/uk/gov/gchq/gaffer/data/elementdefinition/view/NamedViewDetail.java
```diff
@@ -0,0 +1,250 @@
...
+import java.nio.charset.Charset;
+import java.util.Map;
+import java.util.Set;
+
+/**
+ * Simple POJO containing the details associated with a {@link NamedView}.
+ */
+public class NamedViewDetail implements Serializable {
+    private static final long serialVersionUID = -8354836093398004122L;
+    private static final String CHARSET_NAME = CommonConstants.UTF_8;
```
p013570: I know this CHARSET_NAME was copied from NamedOperationDetail, but it would be simpler to remove it and just use our string util: StringUtil.toString().

PR #1648, opened 2018-01-29
Gh 1646 named view resolver needs to check for merged named view names

m55624:


core/graph/src/main/java/uk/gov/gchq/gaffer/graph/hook/NamedViewResolver.java
```diff
@@ -78,26 +79,43 @@ private void resolveViewsInOperations(final Operations<?> operations) {
...
+                    }
+                } else {
+                    fullyResolvedView.merge(resolvedCachedView);
                 }
             }
         } catch (final CacheOperationFailedException e) {
-            // failed to find the namedView in the cache
-            throw new RuntimeException(e);
+            // failed to find the namedView in the cache.
+            // ignore this as users might know it isn't in there.
```
p013570: Ignoring this error means if a user tries to use a named view that doesn't exist it will just not add the view and the user may be unaware that it has failed.

m55624:
All looks good

PR #1661, opened 2018-02-19
Gh 1649 if operation

m607123:
If operation added.
Accepts either a boolean or a predicate as the condition, and will then execute one of two provided operations depending on the condition.

### DIFF
--- a/core/operation/src/main/java/uk/gov/gchq/gaffer/operation/impl/If.java
+++ b/core/operation/src/main/java/uk/gov/gchq/gaffer/operation/impl/If.java
@@ -32,8 +32,10 @@ import java.util.Collection;
 import java.util.Map;
 import java.util.function.Predicate;
 
-
-// TODO Write JavaDoc
+/**
+ * A {@code If} is an {@link Operation} which will execute one of two Operations, based on the result of testing an input Object against a provided {@link Predicate}.
+ * A simple boolean, or anything that resolves to a boolean, can also be used in place of the Predicate.
+ */
 @JsonPropertyOrder(value = {"input", "condition", "predicate", "then", "otherwise", "options"}, alphabetic = true)
 public class If implements InputOutput<Object, Object>, Operations {
 
--- a/core/operation/src/main/java/uk/gov/gchq/gaffer/operation/impl/If.java
+++ b/core/operation/src/main/java/uk/gov/gchq/gaffer/operation/impl/If.java
@@ -1,0 +1,121 @@
+/*
+ * Copyright 2017 Crown Copyright
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package uk.gov.gchq.gaffer.operation.impl;
+
+import com.fasterxml.jackson.core.type.TypeReference;
+import org.apache.commons.lang3.exception.CloneFailedException;
+
+import uk.gov.gchq.gaffer.operation.Operation;
+import uk.gov.gchq.gaffer.operation.Operations;
+import uk.gov.gchq.gaffer.operation.io.InputOutput;
+import uk.gov.gchq.gaffer.operation.serialisation.TypeReferenceImpl;
+
+import java.util.Collection;
+import java.util.Map;
+import java.util.function.Predicate;
+
+public class If implements InputOutput<Object, Object>, Operations {
+
+    private Object input;
+    private Map<String, String> options;
+    private Boolean condition;
+    private Operation then;
+    private Operation otherwise;
+    private Predicate<Object> predicate;
+
+    @Override
+    public Object getInput() {
+        return input;
+    }
+
+    @Override
+    public void setInput(final Object input) {
+        this.input = input;
+    }
+
+    @Override
+    public TypeReference<Object> getOutputTypeReference() {
+        return new TypeReferenceImpl.Object();
+    }
+
+    @Override
+    public Operation shallowClone() throws CloneFailedException {
+        return null;
+    }
+
+    @Override
+    public Map<String, String> getOptions() {
+        return options;
+    }
+
+    @Override
+    public void setOptions(final Map<String, String> options) {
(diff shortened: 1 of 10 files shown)


core/operation/src/main/java/uk/gov/gchq/gaffer/operation/impl/If.java
```diff
@@ -0,0 +1,190 @@
...
+import java.util.Collection;
+import java.util.Map;
+import java.util.function.Predicate;
+
+/**
+ * A {@code If} is an {@link Operation} which will execute one of two Operations, based on the result of testing an input Object against a provided {@link Predicate}.
+ * A simple boolean, or anything that resolves to a boolean, can also be used in place of the Predicate.
+ */
+@JsonPropertyOrder(value = {"input", "condition", "predicate", "then", "otherwise", "options"}, alphabetic = true)
+public class If implements InputOutput<Object, Object>, Operations {
```
p013570: I think this may need to be generic similar to Limit<I>. I.e If<I, O> implements InputOutput<I, O>. It might help when using the OperationChain builder.

core/store/src/main/java/uk/gov/gchq/gaffer/store/operation/handler/IfHandler.java
```diff
@@ -0,0 +1,82 @@
...
+        boolean computedCondition;
+
+        if (null == operation.getCondition()) {
+            computedCondition = null != operation.getPredicate()
+                    && operation.getPredicate().test(input);
+        } else {
+            computedCondition = operation.getCondition();
+        }
+
+        if (computedCondition) {
```
p013570: Perhaps change this logic to:

```java
        final Operation nextOp = computedCondition ? operation.getThen() : operation.getOtherwise();
        final Object result;
        if (null == nextOp) {
            result = input;
        } else {
            updateOperationInput(nextOp, input);
            if (nextOp instanceof Output) {
                result = store.execute((Output) nextOp, context);
            } else {
                store.execute(nextOp, context);
                result = null;
            }
        }
        return result;
```

core/operation/src/main/java/uk/gov/gchq/gaffer/operation/impl/If.java
```diff
@@ -0,0 +1,190 @@
...
+    }
+
+    @Override
+    public void setOptions(final Map<String, String> options) {
+        this.options = options;
+    }
+
+    @Override
+    public Collection getOperations() {
+        return null;
```
p013570: This needs fixing. This is required to allow things like the AddOperationsToChain and OperationChainLimiter graph hooks to work. I'm not sure what the best solution is. Perhaps something like this, but it is a bit fragile and would need a fair bit of testing:
```java
@Override
public Collection<Operation> getOperations() {
    if (null == then) {
        if (null == otherwise) {
            return Collections.emptyList();
        }
        return Lists.newArrayList(OperationChain.wrap(otherwise));
    }
    if (null == otherwise) {
        return Lists.newArrayList(OperationChain.wrap(then));
    }
    return Lists.newArrayList(OperationChain.wrap(then), OperationChain.wrap(otherwise));
}

@Override
public void updateOperations(final Collection<Operation> operations) {
    final Iterator<Operation> itr = operations.iterator();
    if (null != then) {
        if (!itr.hasNext()) {
            throw new IllegalArgumentException("Cannot update operations there are not enough operations to set 'then'");
        }
        then = itr.next();
    }
    if (null != otherwise) {
        if (!itr.hasNext()) {
            throw new IllegalArgumentException("Cannot update operations there are not enough operations to set 'otherwise'");
        }
        otherwise = itr.next();
    }
    if (itr.hasNext()) {
        throw new IllegalArgumentException("Cannot update operations there are too many operations: " + operations.size());
    }
}
```

To calculate the score of the If operation correctly, we need to calculate the score of the then and otherwise operations and take the largest score. This logic should be done by creating an implementation of ScoreResolver.

core/store/src/main/java/uk/gov/gchq/gaffer/store/operation/handler/IfHandler.java
```diff
@@ -0,0 +1,82 @@
...
+            throw new OperationException("Input cannot be null");
+        }
+
+        final Object input = operation.getInput();
+
+        boolean computedCondition;
+
+        if (null == operation.getCondition()) {
+            computedCondition = null != operation.getPredicate()
+                    && operation.getPredicate().test(input);
```
p013570: I think it would be worth putting a try/catch around 'operation.getPredicate().test(input)' as the predicate may not accept the input type. We should catch any ClassCastExceptions and throw a useful error message for users.

p013570:
Please can you also register the operation with the handler in the Store.

core/store/src/main/java/uk/gov/gchq/gaffer/store/operation/handler/IfHandler.java
```diff
@@ -0,0 +1,82 @@
...
+import uk.gov.gchq.gaffer.store.Store;
+
+/**
+ * An operation handler for {@link If} operations.
+ * If {@link If#getThen()} or {@link If#getOtherwise()} returns a null operation, then the input object will simply be returned.
+ */
+public class IfHandler implements OutputOperationHandler<If, Object> {
+    @Override
+    public Object doOperation(final If operation, final Context context, final Store store) throws OperationException {
+        if (null == operation.getInput()) {
```
p013570: null inputs should be allowed.

core/operation/src/main/java/uk/gov/gchq/gaffer/operation/impl/If.java
```diff
@@ -0,0 +1,190 @@
...
+                .toString();
+    }
+
+    public static final class Builder extends Operation.BaseBuilder<If, Builder>
+        implements InputOutput.Builder<If, Object, Object, Builder> {
+        public Builder() {
+            super(new If());
+        }
+
+        public Builder condition(final boolean condition) {
```
p013570: I think you'll need to change the parameter to a Boolean, otherwise your shallowClone method will fail if 'condition' is not set.

core/operation/src/test/java/uk/gov/gchq/gaffer/operation/impl/IfTest.java
```diff
@@ -0,0 +1,75 @@
...
+
+        // Then
+        assertThat(filter.getInput(), is(notNullValue()));
+        assertTrue(filter.getCondition());
+        assertTrue(filter.getThen() instanceof GetElements);
+        assertTrue(filter.getOtherwise() instanceof GetAllElements);
+    }
+
+    @Override
+    public void shouldShallowCloneOperation() {
```
p013570: We need another test to check shallow clone when a condition is not set.

m607123:
Closing until #1672 has been completed.

Issue #1662, opened 2018-02-20
A new user context is required for every Graph execution


p013570:
The Graph class has various methods for executing operations. Some of those methods take a User Context object. A new context object must be used every time an operation is executed on a Graph, which includes a new job id. This is not currently enforced. Users can currently reuse the same context for multiple operations. This will cause problems with clashing job IDs and some of the Context fields will be overridden.

p013570:
Merged into develop.

PR #1666, opened 2018-02-21
gh-1662 - forces the context to be updated

p013570:
When operations are executed on a graph a new context object will be created, containing a new jobId


core/graph/src/test/java/uk/gov/gchq/gaffer/graph/GraphTest.java
```diff
@@ -1844,54 +1948,6 @@ public void shouldThrowExceptionOnExecuteJobWithANullContext() throws OperationE
         }
     }
 
-    @Test
-    public void shouldThrowExceptionOnExecuteWithANullUserInContext() throws OperationException {
```
p013570: These tests have been removed as a context can no longer be created without a user.

Issue #1659, opened 2018-02-19
Add Since annotation to Operation implementations


p013570:
We need to document what version Operations were added in. We should annotate all Operation implementations with 
```java
@Since(version=<Gaffer Version>). 
```
If the version dates back prior to version 1.0, then we should just annotate it with version 1.0.

Once this has been completed we need to update the Gaffer Doc to include this information.

p013570:
Merged into develop.

PR #1674, opened 2018-02-26
gh-1659 - Annotated all concrete implementations of Operation with @Since annotation.

t616178:
Sample implementation of a ```@Since``` annotation for Gaffer operations.


core/operation/src/test/java/uk/gov/gchq/gaffer/operation/AbstractOperationsVersionTest.java
```diff
@@ -0,0 +1,64 @@
...
+
+        // Then
+        reflections.getSubTypesOf(Operation.class)
+                .stream()
+                .filter(cls -> !cls.isInterface())
+                .filter(cls -> !cls.isAnnotationPresent(TestOperation.class))
+                .forEach(cls -> {
+                    final Since since = cls.getAnnotation(Since.class);
+                    if (null == since) {
+                        fail("Class: " + cls.getCanonicalName() + " does not have a version annotation.");
```
p013570: Rather than using this mechanism to test classes have the annotation, I think it would be better to use the mechanism we have for checking classes are annotated with JsonPropertyOrder. This is done by adding shouldHaveJsonPropertyAnnotation test to the JSONSerialisationTest abstract class. In a similar way we could add a test to the abstract OperationTest class. Also, for backwards compatibility, we should avoid failing the test if an operation doesn't have the annotation. This is achieved in the JsonPropertyOrder case by using assumeTrue, so the test is highlighted as being 'skipped' but doesn't cause the build to fail.

core/operation/src/test/java/uk/gov/gchq/gaffer/operation/TestOperation.java
```diff
@@ -0,0 +1,32 @@
...
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Annotation used to denote {@link Operation} implementations which are used by
+ * unit tests.
+ */
+@Target(ElementType.TYPE)
+@Retention(RetentionPolicy.RUNTIME)
+public @interface TestOperation {
```
p013570: I don't think we need this annotation.

core/operation/src/main/java/uk/gov/gchq/gaffer/operation/Since.java
```diff
@@ -0,0 +1,32 @@
...
+import java.lang.annotation.Target;
+
+/**
+ * Annotation used to define the Gaffer version at which an {@link Operation}
+ * implementation was added.
+ */
+@Target(ElementType.TYPE)
+@Retention(RetentionPolicy.RUNTIME)
+public @interface Since {
+    String version();
```
p013570: If you rename version to value, you don't need to include the version field name when you use it. I.e. 
```java
@Since(1.0.0)
``` 
instead of
```java
@Since(version=1.0.0)
```

Issue #1654, opened 2018-02-02
REST end point /graph/config/serialisedFields/{className}/classes may return incorrect classes


p013570:
The mechanism for determining the field classes doesn't seem to return the class type that is used in the setter method. Instead it uses the field class.

For example in the IsA predicate, the 'type' field is a Class, but the JSON setter takes a String. Therefore the serialised field class should be 'String', not 'Class'.

p013570:
Merged into develop.

PR #1675, opened 2018-02-26
Gh-1654 Rest SerialisedFields bug

m607123:
A somewhat fragile but functioning solution to the SerialisedFieldClasses endpoint returning incorrect class types for the serialised fields of a given class.
This cannot currently handle parameterised types (they show simply as "T", for example), but most other cases are handled without issue.
Enums are described as a simple String, since a String will be deserialised correctly.

Added a few Json-related annotations to properties which should not be serialised.


m607123:
Changes look great, thanks @p013570.

p013570:
This was merged in as part of https://github.com/gchq/Gaffer/pull/1700

Issue #1649, opened 2018-01-31
Create an If operation


p013570:
The operation would be an input/output operation and have 4 main fields:
'condition': boolean
'then': Operation
'otherwise': Operation
'input': Object

We should also allow a 'predicate' field to be set instead of the 'condition'. The logic would then look something like:

```java
boolean computedCondition;
if(null == condition) {
    computedCondition = null != predicate && predicate.test(input);
} else {
    computedCondition = condition;
}

if(computedCondition) {
    if(null == then) {
        return input;
    }
    // set the input on the 'then' operation
    // see OperationChainHandler.updateOperationInput
    return store.execute(then, context);
}

if(null == otherwise) {
    return input;
}
// set the input on the 'else' operation
// see OperationChainHandler.updateOperationInput
return store.execute(otherwise, context);
```

This operation should implement Operations and the AddOperationsToChain graph hook should be tested with this operation.

p013570:
Merged into develop.

PR #1676, opened 2018-02-26
Gh-1649 If Operation

m607123:
Added the IfScoreResolver to the ScoreOperationChainHandler's default ScoreResolvers.
Updated test to reflect this change.


core/operation/src/main/java/uk/gov/gchq/gaffer/operation/impl/If.java
```diff
@@ -0,0 +1,244 @@
...
+                return Collections.emptyList();
+            }
+            return Collections.singletonList(OperationChain.wrap(otherwise));
+        }
+
+        if (null == otherwise) {
+            return Collections.singletonList(OperationChain.wrap(then));
+        }
+
+        return Lists.newArrayList(OperationChain.wrap(then), OperationChain.wrap(otherwise));
```
p013570: This needs to include the Conditional operation.

core/operation/src/main/java/uk/gov/gchq/gaffer/operation/impl/If.java
```diff
@@ -0,0 +1,244 @@
...
+    }
+
+    @Override
+    public void updateOperations(final Collection operations) {
+        final Iterator<Operation> itr = operations.iterator();
```
p013570: This should include the conditional operation.

core/store/src/main/java/uk/gov/gchq/gaffer/store/operation/handler/IfHandler.java
```diff
@@ -0,0 +1,100 @@
...
+    @Override
+    public Object doOperation(final If operation, final Context context, final Store store) throws OperationException {
+        final Object input = operation.getInput();
+
+        boolean computedCondition;
+
+        if (null == operation.getCondition()) {
+            final Object intermediate;
+
+            if (null == operation.getConditional().getTransform()) {
```
p013570: You need a null check for operation.getConditional().

core/store/src/main/java/uk/gov/gchq/gaffer/store/operation/resolver/IfScoreResolver.java
```diff
@@ -0,0 +1,52 @@
...
+            final Collection<Operation> ops = operation.getOperations();
+            Integer maxScore = DEFAULT_OPERATION_SCORE;
+
+            for (final Operation op : ops) {
+                final Integer score = defaultScoreResolver.getScore(op);
+                maxScore = Integer.max(score, maxScore);
+            }
+
+            return maxScore;
+        } else {
```
p013570: You don't this this else block. Just do:
```java
if(null != operation) {... return maxScore}

LOGGER.warn("Cannot score a null operation");
return DEFAULT_OPERATION_SCORE;
```

core/store/src/main/java/uk/gov/gchq/gaffer/store/operation/resolver/ScoreResolver.java
```diff
@@ -43,7 +43,7 @@
      * @param defaultScoreResolver the default score resolver to look up scores for nested operations
      * @return the score for the operation, otherwise null if not found
      */
-    default Integer getScore(final T operation, final ScoreResolver defaultScoreResolver) {
+    default Integer getScore(final T operation, final ScoreResolver<T> defaultScoreResolver) {
```
p013570: I don't think you should change this interface method signature, as it will cause breaking changes.

core/store/src/main/java/uk/gov/gchq/gaffer/store/operation/resolver/IfScoreResolver.java
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    @Override
+    public Integer getScore(final If operation, final ScoreResolver defaultScoreResolver) {
+        if (null != operation) {
+            final Collection<Operation> ops = operation.getOperations();
+            Integer maxScore = DEFAULT_OPERATION_SCORE;
+
+            for (final Operation op : ops) {
+                final Integer score = defaultScoreResolver.getScore(op);
```
p013570: The logic will need to be updated when the Conditional operation is added. The conditional operation is always run so the score should always include that. Then the max score from the then/otherwise operation.
Rather than using operation.getOperations(). You can access the conditional, then and otherwise operations directly.

core/store/src/main/java/uk/gov/gchq/gaffer/store/operation/resolver/IfScoreResolver.java
```diff
@@ -0,0 +1,52 @@
...
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import uk.gov.gchq.gaffer.operation.Operation;
+import uk.gov.gchq.gaffer.operation.impl.If;
+
+import java.util.Collection;
+
+public class IfScoreResolver implements ScoreResolver<If> {
```
p013570: Please could you add some tests for this resolver.

m607123:
Closing while changes made

core/operation/src/main/java/uk/gov/gchq/gaffer/operation/impl/If.java
```diff
@@ -0,0 +1,244 @@
...
+ * @see If.Builder
+ */
+@JsonPropertyOrder(value = {"input", "condition", "predicate", "then", "otherwise", "options"}, alphabetic = true)
+public class If<I, O> implements InputOutput<I, O>,
+        Operations {
+
+    private I input;
+    private Map<String, String> options;
+    private Boolean condition;
+    private Operation then;
```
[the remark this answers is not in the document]
m607123: Potentially, but I think this is too restrictive - this forces both <code>then</code> and <code>otherwise</code> to be of the same input/output type when in reality they may be very different - I've also just tried playing around with having them as a generic <code>InputOutput</code> but this also excludes operations such as <code>GetAllElements</code>, as well as <code>OperationChain</code> because they're only implementations of <code>Output</code>.

PR #1678, opened 2018-02-27
Gh 1668 repeat operation

m607123:
Added Repeat operation, operation handler, custom ScoreResolver implementation, and tests for all 3.

### DIFF
--- a/core/operation/src/main/java/uk/gov/gchq/gaffer/operation/impl/Repeat.java
+++ b/core/operation/src/main/java/uk/gov/gchq/gaffer/operation/impl/Repeat.java
@@ -29,6 +29,13 @@ import java.util.Collection;
 import java.util.List;
 import java.util.Map;
 
+/**
+ * A {@code Repeat} is a Gaffer {@link Operation} for executing a delegate <code>Operation</code>
+ * for a specified number of repeats.
+ * <p>The delegate operation can also be an implementation of {@link Operations},
+ * so for example, a small {@link uk.gov.gchq.gaffer.operation.OperationChain} could be delegated
+ * to the <code>Repeat</code>, and repeated a number of times.</p>
+ */
 @JsonPropertyOrder(value = {"class", "input", "operation", "times"}, alphabetic = true)
 public class Repeat implements InputOutput<Object, Object>,
         Operations<Operation> {
--- a/core/operation/src/main/java/uk/gov/gchq/gaffer/operation/impl/Repeat.java
+++ b/core/operation/src/main/java/uk/gov/gchq/gaffer/operation/impl/Repeat.java
@@ -18,9 +18,12 @@ package uk.gov.gchq.gaffer.operation.impl;
 import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.annotation.JsonPropertyOrder;
 import com.fasterxml.jackson.core.type.TypeReference;
+import com.google.common.collect.Iterables;
+import com.google.common.collect.Lists;
 import org.apache.commons.lang3.exception.CloneFailedException;
 
 import uk.gov.gchq.gaffer.operation.Operation;
+import uk.gov.gchq.gaffer.operation.OperationChain;
 import uk.gov.gchq.gaffer.operation.Operations;
 import uk.gov.gchq.gaffer.operation.io.InputOutput;
 import uk.gov.gchq.gaffer.operation.serialisation.TypeReferenceImpl;
@@ -107,6 +110,15 @@ public class Repeat implements InputOutput<Object, Object>,
             ops.add(operation);
         }
         return ops;
+    }
+
+    @Override
+    public void updateOperations(final Collection<Operation> operations) {
+        if (operations.size() > 1) {
+            operation = new OperationChain(Lists.newArrayList(operations));
+        } else {
+            operation = Iterables.get(operations, 0);
+        }
     }
 
     public static final class Builder
--- a/core/operation/src/main/java/uk/gov/gchq/gaffer/operation/impl/Repeat.java
+++ b/core/operation/src/main/java/uk/gov/gchq/gaffer/operation/impl/Repeat.java
@@ -15,6 +15,7 @@
  */
 package uk.gov.gchq.gaffer.operation.impl;
 
+import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.annotation.JsonPropertyOrder;
 import com.fasterxml.jackson.core.type.TypeReference;
 import org.apache.commons.lang3.exception.CloneFailedException;
@@ -23,6 +24,8 @@ import uk.gov.gchq.gaffer.operation.Operation;
 import uk.gov.gchq.gaffer.operation.Operations;
 import uk.gov.gchq.gaffer.operation.io.InputOutput;
 import uk.gov.gchq.gaffer.operation.serialisation.TypeReferenceImpl;
+import uk.gov.gchq.gaffer.operation.util.OperationConstants;
+import uk.gov.gchq.koryphe.Since;
 
 import java.util.ArrayList;
 import java.util.Collection;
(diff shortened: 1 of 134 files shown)


core/operation/src/main/java/uk/gov/gchq/gaffer/operation/impl/Repeat.java
```diff
@@ -0,0 +1,133 @@
...
+import uk.gov.gchq.gaffer.operation.io.InputOutput;
+import uk.gov.gchq.gaffer.operation.serialisation.TypeReferenceImpl;
+
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * A {@code Repeat} is a Gaffer {@link Operation} for executing a delegate <code>Operation</code>
```
t616178: Usage of ```<code>``` tags should be replaced with ```{@code}``` markdown.

core/operation/src/main/java/uk/gov/gchq/gaffer/operation/impl/Repeat.java
```diff
@@ -0,0 +1,133 @@
...
+ * for a specified number of repeats.
+ * <p>The delegate operation can also be an implementation of {@link Operations},
+ * so for example, a small {@link uk.gov.gchq.gaffer.operation.OperationChain} could be delegated
+ * to the <code>Repeat</code>, and repeated a number of times.</p>
+ */
+@JsonPropertyOrder(value = {"class", "input", "operation", "times"}, alphabetic = true)
+public class Repeat implements InputOutput<Object, Object>,
```
t616178: Does this only make sense if the operation being repeated also implements ```InputOutput<I, O>```? If so, the class signature could be amended to:

`java
public class Repeat<I, O> implements InputOutput<I, O>, Operations<InputOutput<I,O>
`

This then would indicate that the Input/Output types are the same:

`java
public class Repeat<T> implements InputOutput<T, T>, Operations<InputOutput<T, T>
`

Perhaps there is room for two of these operations, one which is more general and one which provides a bit more type safety? Or does that introduce unnecessary complication?

core/store/src/main/java/uk/gov/gchq/gaffer/store/operation/handler/RepeatHandler.java
```diff
@@ -0,0 +1,75 @@
...
+ *
+ * If the delegate operation is an implementation of {@link Output}, the output value from each
+ * operation is passed as the input to the next. Otherwise, the delegate operation is just
+ * executed on the store for the configured number of repeats.
+ */
+public class RepeatHandler implements OutputOperationHandler<Repeat, Object> {
+    private int maxTimes;
+
+    public RepeatHandler() {
+        this(20);
```
t616178: We don't need these constructors - the default value of ```maxTimes``` can be set when the field is declared. This would would only be overridden by an operation declaration JSON file. (Like we currently do for ```GetWalksHandler```.

core/store/src/test/java/uk/gov/gchq/gaffer/store/operation/handler/RepeatHandlerTest.java
```diff
@@ -0,0 +1,109 @@
...
+import uk.gov.gchq.gaffer.operation.io.Output;
+import uk.gov.gchq.gaffer.store.Context;
+import uk.gov.gchq.gaffer.store.Store;
+
+import static org.junit.Assert.assertTrue;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
+
+public class RepeatHandlerTest {
```
t616178: Would it also be useful to have a simple integration test which demonstrates this operation working against a store which contains elements?

core/operation/src/main/java/uk/gov/gchq/gaffer/operation/impl/Repeat.java
p013570: I kind of agree with: public class Repeat<T> implements InputOutput<T, T>. However, if the operation is designed to repeat, then I think the input type must be the same as the output type. However It is possible to repeat a GetAdjacentIds operation. The input and output types are actually slightly different. Similarly with GetElements. So we probably will need to use public class Repeat<I, O> implements InputOutput<I, O>

Regarding this: Operations<InputOutput<I,O>   we do this repeat to work with any type of operation. For example repeatedly doing AddElements (no output). Or repeatedly running GetAllElements (no input). Or it could be a chain of operations, consisting of some operations with an input, some with an output and some with both. So I think we will have to leave it as Operations<Operation>.

core/operation/src/main/java/uk/gov/gchq/gaffer/operation/impl/Repeat.java
```diff
@@ -0,0 +1,133 @@
...
+        Operations<Operation> {
+    private Operation operation;
+    private int times;
+    private Object input;
+    private Map<String, String> options;
+
+    public Repeat() {
```
p013570: These constructors should be removed to encourage users to use the builder.

core/operation/src/main/java/uk/gov/gchq/gaffer/operation/impl/Repeat.java
```diff
@@ -0,0 +1,133 @@
...
+ * for a specified number of repeats.
+ * <p>The delegate operation can also be an implementation of {@link Operations},
+ * so for example, a small {@link uk.gov.gchq.gaffer.operation.OperationChain} could be delegated
+ * to the <code>Repeat</code>, and repeated a number of times.</p>
+ */
+@JsonPropertyOrder(value = {"class", "input", "operation", "times"}, alphabetic = true)
+public class Repeat implements InputOutput<Object, Object>,
+        Operations<Operation> {
+    private Operation operation;
+    private int times;
```
p013570: Perhaps create a constant - TIMES_DEFAULT, equal to 1. And then set 'times' to this default.

core/operation/src/main/java/uk/gov/gchq/gaffer/operation/impl/Repeat.java
```diff
@@ -0,0 +1,133 @@
...
+
+    public Operation getOperation() {
+        return operation;
+    }
+
+    public void setTimes(final int times) {
+        this.times = times;
+    }
+
+    public int getTimes() {
```
p013570: Annotate getTimes with @JsonInclude(NON_DEFAULT)

core/operation/src/main/java/uk/gov/gchq/gaffer/operation/impl/Repeat.java
```diff
@@ -0,0 +1,133 @@
...
+    }
+
+    @Override
+    public void setOptions(final Map<String, String> options) {
+        this.options = options;
+    }
+
+    @Override
+    public Collection<Operation> getOperations() {
+        final List<Operation> ops = new ArrayList<>();
```
p013570: just return Collections.singletonList(operation)

core/operation/src/main/java/uk/gov/gchq/gaffer/operation/impl/Repeat.java
```diff
@@ -0,0 +1,133 @@
...
+        for (int i = 0; i < times; i++) {
+            ops.add(operation);
+        }
+        return ops;
+    }
```
p013570: You will also need to implement updateOperations. This should have some logic like:
if(operations.size() > 1) {
operation = new OperationChain(operations);
} else {
operation = operations.get(0);
}

core/store/src/main/java/uk/gov/gchq/gaffer/store/operation/handler/RepeatHandler.java
p013570: Perhaps create a constant MAX_TIMES_DEFAULT and set it to 100.

core/store/src/main/java/uk/gov/gchq/gaffer/store/operation/resolver/RepeatScoreResolver.java
```diff
@@ -0,0 +1,55 @@
...
+ * multiplied by the score of the delegate operation,
+ * or if the delegate is an implementation of {@link Operations},
+ * the sum of the scores of the operations contained within.</p>
+ */
+public class RepeatScoreResolver implements ScoreResolver<Repeat> {
+    @Override
+    public Integer getScore(final Repeat repeat, final ScoreResolver defaultScoreResolver) {
+        int score = 0;
+        if (null != repeat) {
+            final Operation delegate = repeat.getOperation();
```
p013570: I think this can be simplified to:

```java
if(null == repeat || null == repeat.getOperation()) {
return 0;
}

return repeat.getTimes() * defaultScoreResolver.getScore(repeat.getOperation());
```

core/operation/src/main/java/uk/gov/gchq/gaffer/operation/impl/Repeat.java
t616178: Is the point not that we need to create a Collection with the right number of operations? If so we could do:

```Collections.nCopies(operation)```

m607123:
Closing while changes made

m607123:
Closing until 1.5.0.

PR #1707, opened 2018-03-09
Gh 1696 add graph with hooks

GCHQDeveloper404:


core/operation/src/main/java/uk/gov/gchq/gaffer/operation/Operation.java
```diff
@@ -182,7 +188,17 @@ default void close() throws IOException {
...
-        for (final Field field : getClass().getDeclaredFields()) {
+
+        HashSet<Field> fields = Sets.<Field>newHashSet();
+        Class<?> currentClass = this.getClass();
+        String packageName;
+        do {
+            fields.addAll(Arrays.asList(currentClass.getDeclaredFields()));
+            currentClass = currentClass.getSuperclass();
+            packageName = currentClass.getPackage().getName();
+        } while (packageName.startsWith(UK_GOV_GCHQ_GAFFER));
```
p013570: I don't think we should limit it to the Gaffer package. Can we just look at all super classes? Just keep iterating up until currentClass is null.

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/operation/AddGraphWithHooks.java
```diff
@@ -0,0 +1,72 @@
...
+
+@JsonPropertyOrder(
+        value = {"class", "graphId"},
+        alphabetic = true
+)
+
+public class AddGraphWithHooks extends AddGraph {
+    private GraphHook[] hooks;
+//    @Required
+//    private String graphId;
```
p013570: This comment can be removed.

core/operation/src/main/java/uk/gov/gchq/gaffer/operation/Operation.java
GCHQDeveloper404: I thought that was excessive.

PR #1719, opened 2018-03-14
Gh 1458 cache generics mismatch

GCHQDeveloper404:


GCHQDeveloper404:
Not an optimal solution.

PR #1730, opened 2018-03-20
Gh 1729 federated op validator

p013570:


store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/FederatedOperationChainValidator.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016-2018 Crown Copyright
+ * Copyright 2016 Crown Copyright
```
m607123: This should be changed back to the correct date.

PR #1746, opened 2018-03-28
Gh 1743 do not provide defaults for data directory in parquet store

gaffer01:


gaffer01:
A note on the migration steps has been added to the issue.

I haven't listed the packages that need to be renamed as the renamed ones are only internal to the Parquet store; clients of the Parquet store shouldn't be using these classes directly.

PR #1751, opened 2018-04-03
Gh 1697 row id aggregator assumes column family is group

m55624:


p013570:

Issue #1762, opened 2018-04-09
NPE in Parquet store when adding element with group that is not in the schema


gaffer01:
The following code produces a NPE:

    Schema s1 = new Schema.Builder()
                .type("string", String.class)
                .type("long", Long.class)
                .edge("X", new SchemaEdgeDefinition.Builder().source("string").destination("string").property("long", "long").aggregate(false).build())
                .build();
    Graph g = new Graph.Builder()
                .addSchema(s1)
                .storeProperties(TestUtils.getParquetStoreProperties(testFolder))
                .config(new GraphConfig.Builder().graphId("g").build())
                .build();

    User user = new User();
    Edge undirectedEdge = new Edge.Builder()
                .group("Y")
                .source("A")
                .dest("B")
                .directed(false)
                .property("long", 2L)
                .build();
    g.execute(
                new AddElements.Builder()
                        .input(undirectedEdge)
                        .build(),
                user
        );

gaffer01:
Merged into develop.

PR #1774, opened 2018-04-17
gh-1762 - check added to check group of element being added is in schema

m55624:


gaffer01:
Don't think this is the right way to do it.

- This consumes the entire iterable of elements to be added to check whether the groups are in the schema - it then needs to be consumed again to add the elements - this is very expensive and there's no guarantee that the iterable can be consumed twice.
- It iterates through all the elements and if any one has a group in the schema then it says it's fine and will then attempt to add all the elements - you'll still get a NPE if one of the elements has a group that's in the schema and one doesn't.

The check that the group is valid needs to be done per element just before the element is added - if the group is not in the schema then a warning should be logged.

store-implementation/parquet-store/src/test/java/uk/gov/gchq/gaffer/parquetstore/ParquetStoreTest.java
```diff
@@ -16,18 +16,67 @@
...
+                    .aggregate(false)
+                    .build())
+            .entity(TestGroups.ENTITY, new SchemaEntityDefinition.Builder()
+                    .property(TestPropertyNames.PROP_1, TestTypes.PROP_INTEGER)
+                    .vertex(TestTypes.ID_STRING)
+                    .aggregate(false)
+                    .build())
+            .build();
+    private final Edge testEdge = new Edge.Builder()
+            .group(TestGroups.EDGE_2)
```
p013570: Perhaps rename this to unknownEdge. Similarly for the entity.

store-implementation/parquet-store/src/main/java/uk/gov/gchq/gaffer/parquetstore/operation/handler/utilities/WriteUnsortedData.java
```diff
@@ -84,17 +84,17 @@ private void _writeElements(final Iterator<? extends Element> elements) throws O
...
             } else {
                 splitToWriter = new HashMap<>();
                 groupSplitToWriter.put(group, splitToWriter);
             }
             if (schemaUtils.getEntityGroups().contains(group)) {
                 writer = getWriter(splitToWriter, groupToSplitPoints.get(group), ((Entity) element).getVertex(), group, ParquetStoreConstants.VERTEX);
-            } else {
+            } else if (schemaUtils.getEdgeGroups().contains(group)) {
                 writer = getWriter(splitToWriter, groupToSplitPoints.get(group), ((Edge) element).getSource(), group, ParquetStoreConstants.SOURCE);
             }
```
gaffer01: Can we have a final else to log the fact that the element is ignored if it's not in schema - probably DEBUG level logging?

store-implementation/parquet-store/src/test/java/uk/gov/gchq/gaffer/parquetstore/ParquetStoreTest.java
```diff
@@ -39,4 +88,36 @@ public void testTraits() throws StoreException {
...
+    }
+
+    @Test
+    public void shouldNotThrowExceptionWhenAddingAMultipleElementsWithGroupsNotInSchema() throws Exception {
+        getGraph().execute(new AddElements.Builder()
+                        .input(testEntity, testEdge)
+                        .build(),
+                new User());
+    }
+
```
gaffer01: Can we have one more test to cover the case when one element is not in the schema, and one is in the schema - in this case the one that is in the schema should still be added.

store-implementation/parquet-store/src/main/java/uk/gov/gchq/gaffer/parquetstore/operation/handler/utilities/WriteUnsortedData.java
m55624: This is on the four lines below, but at a WARN level.

gaffer01: So it is, ignore my comment.

PR #1779, opened 2018-04-24
Gh 1771 multi serialiser

GCHQDeveloper404:


GCHQDeveloper404:
@p013570 
Interesting side effect.

Before:
With an unordered edge int->int the order doesn't matter.

Now:
With an unordered edge **with different vertex types**. int->string the direction of the edge doesn't matter, but the types do matter. I experienced a flip in the scr/dest when string was being placed within the int vertex.

PR #1781, opened 2018-05-01
Gh 1770 add class field to operation field

m55624:


m55624:
Closing for tests to be added.

Issue #1796, opened 2018-05-15
Add a property to enable/disable skipping of unknown properties in JSON


p013570:
To help with federated over different gaffer stores with different versions of Gaffer we should enable the FAIL_ON_UNKNOWN_PROPERTIES feature in Jackson. This will mean, if any new fields are added to classes in Gaffer, it won't cause failures for older versions of Gaffer. We should give users the option to disable this in their Store Properties file.

p013570:
Merged into develop.

p013570:
There is a new store property (or SystemProperty) that you can set to enable/disable the strict mode in json.

PR #1797, opened 2018-05-15
gh-1796 - enabled FAIL_ON_UNKNOWN_PROPERTIES in Jackson

p013570:


core/serialisation/src/main/java/uk/gov/gchq/gaffer/jsonserialisation/JSONSerialiser.java
```diff
@@ -136,6 +140,12 @@ public static void addSimpleClassNames(final boolean includeSubtypes, final Clas
     }
 
     public static void update(final String jsonSerialiserClass, final String jsonSerialiserModules) {
+        update(jsonSerialiserClass, jsonSerialiserModules, null);
```
m55624: an update with strictJson boolean as null is never tested

PR #1810, opened 2018-06-12
Gh 1636 gaffer graph migrations and versioning v2

m55624:


core/operation/src/main/java/uk/gov/gchq/gaffer/operation/function/migration/ToUpperCase.java
```diff
@@ -0,0 +1,34 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package uk.gov.gchq.gaffer.operation.function.migration;
+
+import uk.gov.gchq.koryphe.function.KorypheFunction;
+
+public class ToUpperCase extends KorypheFunction<Object, String> {
```
p013570: Perhaps this could be moved into Koryphe? Similarly with the other new functions. You could keep them in Gaffer for now until Koryphe has been updated, but perhaps just move them to src/test

core/operation/src/main/java/uk/gov/gchq/gaffer/operation/function/migration/ReturnValue.java
```diff
@@ -0,0 +1,43 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package uk.gov.gchq.gaffer.operation.function.migration;
+
+import uk.gov.gchq.koryphe.function.KorypheFunction;
+
+public class ReturnValue extends KorypheFunction<Object, Object> {
```
p013570: Perhaps rename this to Identity and again this could be in Koryphe.

core/data/src/main/java/uk/gov/gchq/gaffer/data/elementdefinition/view/View.java
```diff
@@ -197,6 +198,14 @@ public void expandGlobalDefinitions() {
         }
     }
 
+    public void setHasBeenMigrated(final boolean hasBeenMigrated) {
```
p013570: Would it be possible to avoid this? Perhaps it would be neater to have a field:
private Map<String, String> config;

where you can just set arbitrary config in it. This would avoid haven't to add specific logic for migration into the core code.

core/store/src/main/java/uk/gov/gchq/gaffer/store/schema/ViewValidator.java
```diff
@@ -108,11 +112,15 @@ public ValidationResult validate(final View view, final Schema schema, final Set
                         }
 
                         result.add(validateAgainstStoreTraits(viewElDef, storeTraits));
-                        result.add(validateFunctionArgumentTypes(viewElDef.getPreAggregationFilter(), viewElDef, schemaElDef));
-                        result.add(validateFunctionArgumentTypes(viewElDef.getAggregator(), viewElDef, schemaElDef));
-                        result.add(validateFunctionArgumentTypes(viewElDef.getPostAggregationFilter(), viewElDef, schemaElDef));
-                        result.add(validateFunctionArgumentTypes(viewElDef.getTransformer(), viewElDef, schemaElDef));
-                        result.add(validateFunctionArgumentTypes(viewElDef.getPostTransformFilter(), viewElDef, schemaElDef));
+
+                        if (!view.hasBeenMigrated()) {
```
p013570: Rather than calling it "hasBeenMigrated" perhaps it should be a config item called "skipValidation".

core/operation/src/main/java/uk/gov/gchq/gaffer/operation/function/migration/ToUpperCase.java
m55624: If I move all of these, for example the new Identity function, it will not be accessible in the SchemaMigration where it is currently used.  These, or at least the Identity class will need to stay accessible out of test.

core/graph/src/main/java/uk/gov/gchq/gaffer/graph/hook/migrate/MigrateElement.java
```diff
@@ -0,0 +1,130 @@
...
+        addOldGroupTransform();
+    }
+
+    public String getOldGroup() {
+        return oldGroup;
+    }
+
+    public void setNewGroup(final String newGroup) {
+        this.newGroup = newGroup;
+        addNewGroupTransform();
```
p013570: Unfortunately, automatically adding the newGroup transform function in, means if the class is serialised into json it includes this transform. Then if the json is deserialised, it will include the transform twice. Ideally this newGroup (and oldGroup) transform function should only be added at when needed and shouldn't modify the original list of functions.

core/graph/src/main/java/uk/gov/gchq/gaffer/graph/hook/migrate/SchemaMigration.java
```diff
@@ -0,0 +1,645 @@
...
+    public boolean aggregateAfter = false;
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(SchemaMigration.class);
+
+    private List<MigrateElement> entities = new ArrayList<>();
+    private List<MigrateElement> edges = new ArrayList<>();
+
+    private MigrationOutputType outputType = DEFAULT_OUTPUT_TYPE;
+
+    private Map<String, ElementFilter> entitiesPostAggregationFilterMap = new HashMap<>();
```
p013570: Graph hooks need to be stateless, because the same instance is used for every query for different users. So we can't store these maps as fields, they will need to be local variables.

core/graph/src/main/java/uk/gov/gchq/gaffer/graph/hook/migrate/SchemaMigration.java
```diff
@@ -0,0 +1,645 @@
...
+                if (!entitiesPostTransformFilterMap.isEmpty() || !edgesPostTransformFilterMap.isEmpty()) {
+                    Filter postTransformFilter = new Filter.Builder()
+                            .entities(entitiesPostTransformFilterMap)
+                            .edges(edgesPostTransformFilterMap)
+                            .build();
+                    hookOpList.add(postTransformFilter);
+                }
+
+                if (!hookOpList.isEmpty()) {
+                    afterOpsMap.put(op.getClass().getName(), hookOpList);
```
p013570: We can't do this because if the OperationChain contained:
GetElements
GetElements

We would end up adding the hookOpList after both instances of the GetElements operation.

m55624:
Closing for fixes.

p013570:

Issue #1800, opened 2018-05-22
Create ability to include all groups for Entities/Edges in a View


m55624:
Currently if you have a view, for example:
``"view" : {
    "edges": {
      "group1": {},
      "group2": {}
    }
}``

you must specify all groups you want to return within the view.  The feature of allowing all groups to be included should be added, something like:

``"view" : {
    "edges": {
      "allGroups": "true"
    }
}``

This will become very useful when seedMatching has been deprecated (https://github.com/gchq/Gaffer/issues/1798) for seedMatching EQUALS.

p013570:
Merged into develop.

PR #1814, opened 2018-06-27
gh-1800 - allEntities/allEdges flags added to view and method updated…

m55624:
… to reflect, tests added


core/graph/src/main/java/uk/gov/gchq/gaffer/graph/Graph.java
```diff
@@ -246,11 +246,21 @@ private void updateOperationChainView(final Operations<?> operations) {
                 View opView = ((OperationView) operation).getView();
                 if (null == opView) {
                     opView = config.getView();
-                } else if (!(opView instanceof NamedView) && !opView.hasGroups()) {
+                } else if (!(opView instanceof NamedView) && !opView.hasGroups() && !opView.isAllEdges() && !opView.isAllEntities()) {
                     opView = new View.Builder()
                             .merge(config.getView())
                             .merge(opView)
                             .build();
+                } else if (opView.isAllEdges() || opView.isAllEdges()) {
```
p013570: This should be:

```java
else if (opView.isAllEdges() || opView.isAllEntities())
```

Please can you add another test to check it works for both edges and entities.

Issue #1815, opened 2018-06-28
Add output type to OperationDetail class


p013570:
The rest end point /graph/operations/{className} should also return the output type of an operation (if the operation is an instance of Output).

In the OperationDetail constructor we could add something like:
```java
if(exampleJson instanceof Output) {
    final Type outputType = ((Output) exampleJson).getOutputTypeReference().getType();
    outputClassName = // convert outputType into a string similar to how we do it for the fields.
}

PR #1817, opened 2018-06-28
gh-1815 - outputType added to OperationDetail and tests added

m55624:


rest-api/core-rest/src/main/java/uk/gov/gchq/gaffer/rest/service/v2/OperationServiceV2.java
```diff
@@ -296,6 +298,15 @@ private Operation generateExampleJson(final Class<? extends Operation> opClass)
         return operationFields;
     }
 
+    private String getOperationOutputType(final Operation operation) {
+        if (operation instanceof Output) {
+            final Type outputType = ((Output) operation).getOutputTypeReference().getType();
+            return outputType.getTypeName();
```
p013570: Please can you format the output type name in the same way that is done for JsonSerialisationUtil.getSerialisedFieldClasses

Issue #1811, opened 2018-06-18
Allow the set permissions step to be skipped in the Accumulo import R files job


p013570:
When importing data from HDFS into Accumulo, part of the process sets the permissions on the generated R files. This step isn't always required and we should allow users to disable it if they don't need it.

HDFS permissions can be set globally using:

spark.hadoop.fs.permissions.unmask-mode

and/or

fs.permissions.unmast-mode

Specifically we need to conditionally run IngestUtils.setDirectoryPermsForAccumulo in the ImportElementsToAccumuloTool class.

We could implement this by adding a Gaffer operation option "accumulostore.operation.hdfs.skip_permissions" so advanced users could set this to "true",  which would then skip calling IngestUtils.setDirectoryPermsForAccumulo.

p013570:
Merged into develop.

PR #1822, opened 2018-07-04
gh-1811 - ability to skip permissions on accumulo imports added

m55624:


store-implementation/accumulo-store/src/main/java/uk/gov/gchq/gaffer/accumulostore/operation/hdfs/handler/job/tool/ImportElementsToAccumuloTool.java
```diff
@@ -59,7 +69,10 @@ public int run(final String[] strings) throws Exception {
         fs.delete(new Path(inputPath + "/_SUCCESS"), false);
 
         // Set all permissions
-        IngestUtils.setDirectoryPermsForAccumulo(fs, new Path(inputPath));
+        if (!options.containsKey(AccumuloProperties.HDFS_SKIP_PERMISSIONS) ||
```
p013570: You could get a null pointer exception here if the value of the option is null. It is probably better to do:

```java
if(!Boolean.parseBoolean(options.get(AccumuloProperties.HDFS_SKIP_PERMISSIONS))) {
    IngestUtils.setDirectoryPermsForAccumulo(fs, new Path(inputPath));
}


```

PR #1857, opened 2018-09-11
Gh 1845 create foreach operation

m55624:


core/store/src/main/java/uk/gov/gchq/gaffer/store/operation/handler/ForEachHandler.java
```diff
@@ -0,0 +1,65 @@
...
+
+        if (null == inputs) {
+            throw new OperationException("Inputs cannot be null");
+        }
+
+        for (Object input : inputs) {
+            Operation opInstance = operation.shallowClone();
+            OperationHandlerUtil.updateOperationInput(opInstance, input);
+
+            output.add(store.execute((Output) opInstance, context));
```
p013570: This could actually now result in a class cast exception.

Issue #1859, opened 2018-09-12
Create a Reduce operation


p013570:
The Reduce operation should reduce an iterable into a single value. For example to be able to sum together an iterable of integers. We should be able to use our existing library of BinaryOperators (aggregate functions) with the Reduce opreration.

e.g.

```java
public class Reduce<T> implements InputOutput<Iterable<T>, T>, MultiInput<T> {
    private BinaryOperation<T> aggregateFunction;
}
```

lucean:
I'll pick this up if that's ok?

p013570:
@lucean - sorry just seen your message. Yes, feel free to pick anything up that isn't assigned to someone - just add a comment to say you are working on it.

p013570:
Merged into develop.

p013570:
Merged into develop

PR #1866, opened 2018-09-14
gh-1859 - Implement a Reduce operation.

lucean:
A simple implementation of a `Reduce` operation which should be compatible with the existing Koryphe `BinaryOperator` classes.

There's an issue with generic invariance on the Java 8 implementation of the `Stream#reduce(BinaryOperator<T> binaryOperator)` method.

A workaround is to "coerce" the types of objects in the stream into the target type by using the three argument `reduce(U identity, BiFunction<U,? super T,U> accumulator, BinaryOperator<U> combiner)` method.

I can provide more examples of the `Reduce` operation as an integration test if that would be helpful.

The `@Since` annotations will also need to be updated, depending on your release schedule.


p013570:
@lucean - thank you very much for this pull request, it looks perfect 👍

Issue #1865, opened 2018-09-14
Provide ability to filter FreqMap by regex


javadev001001:
Allow for transforming of Frequency Maps using a Regex to extract the elements of interest, leaving only those that match the regular expression.

mehulumistry:
I am a beginner. I want to start with contributing in this.

m55624:
Merged into develop.

PR #1886, opened 2018-10-02
gh-1865 - Regex and BiPredicate filtering

nikmanG:
In reference to #1865 allows for both regex filtering and bipredicate.

### DIFF
--- a/core/cache/pom.xml
+++ b/core/cache/pom.xml
@@ -3,7 +3,7 @@
     <parent>
         <artifactId>core</artifactId>
         <groupId>uk.gov.gchq.gaffer</groupId>
-        <version>1.7.0</version>
+        <version>1.7.1-SNAPSHOT</version>
     </parent>
     <modelVersion>4.0.0</modelVersion>
 
--- a/core/cache/pom.xml
+++ b/core/cache/pom.xml
@@ -3,7 +3,7 @@
     <parent>
         <artifactId>core</artifactId>
         <groupId>uk.gov.gchq.gaffer</groupId>
-        <version>1.7.0-RC4</version>
+        <version>1.7.0-RC5-SNAPSHOT</version>
     </parent>
     <modelVersion>4.0.0</modelVersion>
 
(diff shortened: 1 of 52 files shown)


d21211122:
Hi,

Thanks for your contribution - the functions look really useful.

We don’t usually add functionality like this directly to our value objects, though, rather we add that logic to external functions. You can see an example of that if you look at the FreqMapAggregator in the functions package. So rather than adding an ‘aggregate’ method to the FreqMap, we’ve created an external function that accepts 2 FreqMaps and merges them together.

Similarly, we’d prefer to create an external FreqMapFilter function, that takes a Predicate as an argument that is used to filter the keys. The Koryphe project already has a Regex matching predicate that you could re-use for adding the regex functionality.

I guess it’d look a little like this:

class FreqMapFilter extends KorypheFunction<FreqMap, FreqMap>
{
    private Predicate<String> keyFilter;

    public FreqMapFilter() {}

    public FreqMapFilter(Predicate<String> p)
    {
        keyFilter = p;
    }

    public FreqMap apply(FreqMap input)
    {
        //iterate through the entries in input FreqMap and remove if keyFilter predicate test fails
    }
}

You can then use the Koryphe Regex predicate to get the specific functionality you want:

FreqMapFilter fmf = new FreqMapFilter(new Regex(“myPattern”));
FreqMap filtered = fmf.apply(myFreqMap);

gaffer01:
@d21211122 Should #1865 be updated then to say what the correct approach is? It implies that a Regex predicate is needed but as you say that already exists in Koryphe. So we just need a FreqMapKeyFilter class which accepts a predicate.

d47853:
Could you change the base from master to develop?

nikmanG:
All done but had trouble serializing because of the predicate.

nikmanG:
Given that the JSONSerialiser needs an empty constructor I think the best course of action would be to change it to return the original freq map (a copy of it anyway). That's two birds with one stone.
The issue I had with serializing the predicate is that it will return
`predicate = {}` even when provided a method hence making that somewhat a lost cause.

d47853:
I guess this ties back to @d21211122's suggestion of using Koryphe predicates which would give you the ability to serialise it.

PR #1888, opened 2018-10-08
Gh 1884 change partitioning in parquet store

gaffer01:


gaffer01:
Failed due to tests taking longer than 45 minutes. Closing for now.

PR #1891, opened 2018-10-09
Gh 1847 create join op

m55624:


core/data/src/main/java/uk/gov/gchq/gaffer/data/element/comparison/ElementEquality.java
```diff
@@ -0,0 +1,100 @@
...
+
+    public ElementEquality() {
+    }
+
+    public ElementEquality(final Set<String> groupByProperties) {
+        this.groupByProperties = groupByProperties;
+    }
+
+    public ElementEquality(final String... groupByProperties) {
+        this.groupByProperties = Sets.newHashSet(groupByProperties);
```
gaffer01: Very minor, but should probably just add these properties to groupByProperties - a set is created at the top of the class so no need to create another one.

core/data/src/main/java/uk/gov/gchq/gaffer/data/element/comparison/ElementEquality.java
```diff
@@ -0,0 +1,100 @@
...
+ */
+public class ElementEquality extends KoryphePredicate2<Element, Element> {
+    private Set<String> groupByProperties = new HashSet<>();
+
+
+    public ElementEquality() {
+    }
+
+    public ElementEquality(final Set<String> groupByProperties) {
+        this.groupByProperties = groupByProperties;
```
gaffer01: Maybe add all these to the existing set? It's safer as then the set won't change if the caller of this constructor changes the set.

core/operation/src/test/java/uk/gov/gchq/gaffer/operation/impl/JoinTest.java
```diff
@@ -0,0 +1,98 @@
...
+                .matchMethod(new TestMatch())
+                .matchKey(MatchKey.LEFT)
+                .joinType(JoinType.INNER)
+                .mergeMethod(new TestMerge())
+                .collectionLimit(10)
+                .build();
+    }
+
+    /**
+     * private copy of the ElementMatch class using the count property to match by.
```
gaffer01: This comment looks wrong.

core/operation/src/test/java/uk/gov/gchq/gaffer/operation/impl/JoinTest.java
```diff
@@ -0,0 +1,98 @@
...
+     */
+    public static class TestMatch implements Match{
+        @Override
+        public List matching(final Object testObject, final List testList) {
+            return testList;
+        }
+    }
+
+    /**
+     * private copy of the ElementMatch class using the count property to match by.
```
gaffer01: Same comment as above.

core/data/src/main/java/uk/gov/gchq/gaffer/data/element/comparison/ElementEquality.java
m55624: Do you mean do something like this? ```this.groupByProperties.addAll(Arrays.asList(groupByProperties));``` Is there any benefit to this?

gaffer01: Oh yes, you need to create a list before you can add it so not worth it - just thought creating another set was unnecessary.

Issue #1921, opened 2018-10-19
ImportRDDOfElements on Parquet store fails when used on a HDFS cluster


gaffer01:
The `ImportRDDOfElements` operation on the Parquet store passes all the unit tests which run on a local file system but fails when run on HDFS running on a cluster. The line

    fs.rename(new Path(tempDir + "/AddElementsFromRDDTemp/sorted_aggregated_new/"), new Path(newDataDir));

in `AddElementsFromRDD` fails to move the directory as intended (the sorted_aggregated_new directory is moved into the new directory). This means that the group directories cannot be found when the partition is calculated.

d47853:
Merged into develop

PR #1922, opened 2018-10-19
gh-1921 Fixed bug with renaming in AddElementsFromRDD

gaffer01:


gaffer01:
That's not changed in this PR though. If we want to do that then that should be a separate issue.

It is a long method but in this case I think breaking it out into smaller methods makes it harder to understand and maintain.

Issue #1927, opened 2018-10-23
ImportRDDOfElements fails with "failed to create file" error on Parquet store


gaffer01:
When `ImportRDDOfElements` is called on the Parquet store it can fail with a "failed to create file" error in the `input.foreachPartition` stage. This is because speculative execution is causing multiple attempts for the same partition to all attempt to create the same file.

There are a few options for fixing this:

- Write each file with a filename that includes the attempt id. Once the task has finished, move it to it's final destination.
- Reuse some of the classes related to `FileCommitProtocol` from Spark.
- Write all the elements into one Dataframe (with merged schema) and then write out bucketed by group.

gaffer01:
Merged into develop.

PR #1938, opened 2018-11-09
gh-1927 Files written by WriteData now use task attempt id in name

gaffer01:


store-implementation/parquet-store/src/main/java/uk/gov/gchq/gaffer/parquetstore/operation/handler/spark/utilities/WriteData.java
```diff
@@ -56,28 +59,47 @@ public WriteData(final Function<String, String> groupToDirectory, final Schema s
...
         }
-        LOGGER.info("Finished writing elements for partition id {} to {}", partitionId, groupToDirectory);
+        LOGGER.info("Finished writing elements for partition id {} and task attempt id {} to {}",
+                partitionId, taskAttemptId, groupToDirectory);
         for (final ParquetWriter<Element> writer : groupToWriter.values()) {
             LOGGER.debug("Closing writer {}", writer);
             writer.close();
         }
+
+        LOGGER.info("Renaming output files from {} to {}",
```
d47853: Not sure this is in the scope of the issue

store-implementation/parquet-store/src/main/java/uk/gov/gchq/gaffer/parquetstore/operation/handler/spark/utilities/WriteData.java
```diff
@@ -56,28 +59,47 @@ public WriteData(final Function<String, String> groupToDirectory, final Schema s
...
         final SchemaUtils schemaUtils = new SchemaUtils(Schema.fromJson(schemaAsJson));
         final int partitionId = TaskContext.getPartitionId();
+        final long taskAttemptId = TaskContext.get().taskAttemptId();
         final Map<String, ParquetWriter<Element>> groupToWriter = new HashMap<>();
+        final Map<String, Path> groupToWriterPath = new HashMap<>();
+
         for (final String group : schemaUtils.getGroups()) {
-            groupToWriter.put(group,
-                    buildWriter(group, groupToDirectory.get(group) + "/input-" + partitionId + ".parquet", schemaUtils));
+            groupToWriterPath.put(group, new Path(groupToDirectory.get(group) + "/input-" + partitionId + "-" + taskAttemptId + ".parquet"));
```
d47853: Can you add a unit test for this please?

store-implementation/parquet-store/src/main/java/uk/gov/gchq/gaffer/parquetstore/operation/handler/spark/utilities/WriteData.java
gaffer01: It's not, but in fixing this issue I found it useful to log that so left it in. I don't see any harm in adding it in this issue.

store-implementation/parquet-store/src/main/java/uk/gov/gchq/gaffer/parquetstore/operation/handler/spark/utilities/WriteData.java
gaffer01: There already are unit tests that the process works, and those still pass. The issue only happens when speculative execution causes multiple attempts of the same task on the same partition. I don't see how it's possible to unit test that scenario. (The title of this PR is misleading - changing the name of the file to include the attempt id is just part of the solution, not the point of the PR). But it is probably possible to test two manual attempts to process the same partition and check that only one works - I'll add that.

store-implementation/parquet-store/src/main/java/uk/gov/gchq/gaffer/parquetstore/operation/handler/spark/utilities/WriteData.java
d47853: Sorry - my comment wasn't very clear. I don't mind the logging but I was really on about renaming the files to remove the attempt Id from the file name. I was wondering whether it was actually nessecary.

gaffer01: Yes, renaming them is critical to fixing the issue. A unique filename needs to initially be created for each call of the function to avoid the failed to create file error but then the result needs to be renamed to the correct file name. Whichever call of the function renames it first wins, other calls just don't do the rename.

PR #1940, opened 2018-11-14
Created AbstractStore without coreOpHandlers.

GCHQDeveloper404:
Implemented Store to extend AbstractStore and add coreOpHandlers.
ProxyStore implements AbstractStore only.


store-implementation/proxy-store/src/main/java/uk/gov/gchq/gaffer/proxystore/ProxyStore.java
```diff
@@ -80,6 +80,10 @@
     private Schema schema;
     private Set<Class<? extends Operation>> supportedOperations;
 
+    public ProxyStore() {
```
m55624: This will be a breaking change. For anyone using it currently it will support the core op handlers, if we add this it will now not.

GCHQDeveloper404:
Switched/renamed to https://github.com/gchq/Gaffer/pull/1949

GCHQDeveloper404:
switched/renamed to #1949

Issue #1945, opened 2018-11-15
Swagger UI doesn't work when REST API is served from /


ctas582:
I'm running the REST API on Tomcat and it's configured to be served from the root i.e. there's no path prefix. I've set the `gaffer.rest-api.basePath` system property to `""`, however the following code forces the URLs that Swagger uses to be prefixed with `//v2/` which causes 404 errors.

https://github.com/gchq/Gaffer/blob/b892624ccf99d501f381d6df5c4efc415902b7f1/rest-api/core-rest/src/main/java/uk/gov/gchq/gaffer/rest/application/ApplicationConfigV2.java#L52-L55

Maybe add a length check too? i.e. `basePath.length() > 0 && !basePath.startsWith("/")`

d47853:
Merged into develop

PR #1958, opened 2018-11-28
gh-1945 - fix to ApplicationConfig empty path and test added

m55624:


rest-api/core-rest/src/test/java/uk/gov/gchq/gaffer/rest/service/v2/ApplicationConfigV2Test.java
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    @After
+    public void tearDown() {
+        if (null != server) {
+            server.shutdownNow();
+            server = null;
+        }
+    }
+
```
d47853: If the system property is set to '/' or end's with a '/', do we still get the same issue?

PR #1923, opened 2018-10-22
Gh 1657 loader tests

m55624:

PR #1959, opened 2018-11-29
running all tests on travis

d47853:
Don't merge. Only doing this for investigatory purposes.


d47853:
This could be merged if it passes CI.

d47853:
This will be done in #1923

Issue #1966, opened 2018-12-05
Update Guava version


m55624:

d47853:
Can't be done until Accumulo version is updated

PR #1967, opened 2018-12-05
gh-1966 - guava version updated

m55624:


m55624:
To be merged after Koryphe version updated with updated Guava version

PR #2003, opened 2018-12-19
Gh 2000 get variables throws npe

benj-fletch:


benj-fletch:
While looking at this area of code, is there benefit in returning results from this operation which are keys mapped to null? (functionality illustrated [in this test](https://github.com/gchq/Gaffer/blob/develop/core/store/src/test/java/uk/gov/gchq/gaffer/store/operation/handler/GetVariablesHandlerTest.java#L83))

Is it more desirable by default to not return these results and add an option to the operation to return the mappings to null if the user desires it?

core/store/src/main/java/uk/gov/gchq/gaffer/store/operation/handler/GetVariablesHandler.java
```diff
@@ -22,13 +22,18 @@
...
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 
 public class GetVariablesHandler implements OperationHandler<GetVariables> {
     @Override
     public Map<String, Object> doOperation(final GetVariables operation, final Context context, final Store store) throws OperationException {
         final Map<String, Object> variableMap = new HashMap<>();
-        for (final String key : operation.getVariableNames()) {
+        List<String> variableNames = operation.getVariableNames();
```
m55624: This will probably read better just as one line:
```List<String> variableNames = operation.getVariableNames() != null ? operation.getVariableNames() : Lists.newArrayList();```
I know returning earlier will be more efficient but the iteration of an empty list should not really affect performance.

m55624:
Your above comment was considered when it was implemented but users may want to use a null value (for whatever reason) and this provides a way to do this without a flag.  Also further flags makes it not possible to use nulls within JSON OperationChains using Variables (which is the final goal for these Variables tickets), again something we don't want to limit users to.

core/store/src/main/java/uk/gov/gchq/gaffer/store/operation/handler/GetVariablesHandler.java
benj-fletch: I agree, it's negligible difference. I have updated

benj-fletch:
@m55624 makes sense, thanks for clarifying

m55624:
@benj-fletch No problem, thanks for the contribution again :)

Issue #1905, opened 2018-10-10
GroupPartitioner in Parquet store has a todo to precompute the list of partitions


gaffer01:

gaffer01:
Merged into develop.

PR #2006, opened 2018-12-19
GH-1905 - add pre-computation of partitions to Parquet GroupPartitioner

benj-fletch:


benj-fletch:
Note there might be some change in functionality here since we are now sharing the same list across any calls to `GroupPartitioner#getPartitions()`.

I also considered merging `splitPoints` and `partitions` into a `LinkedHashMap<PartitionKey, Partition>` where the key is the start of the partition (split point). It got a bit finicky to guarantee order, however, so removed it. Does anyone think there is benefit in merging the two lists?

store-implementation/parquet-store/src/main/java/uk/gov/gchq/gaffer/parquetstore/partitioner/GroupPartitioner.java
```diff
@@ -200,6 +207,8 @@ public String toString() {
         return new ToStringBuilder(this)
                 .append("group", group)
                 .append("splitPoints", splitPoints)
+                .append("partitions", partitions)
+                .append("partitionKeyLength", partitionKeyLength)
```
m55624: This can be removed after above comment.

store-implementation/parquet-store/src/main/java/uk/gov/gchq/gaffer/parquetstore/partitioner/GroupPartitioner.java
```diff
@@ -218,6 +227,8 @@ public boolean equals(final Object obj) {
         return new EqualsBuilder()
                 .append(group, other.group)
                 .append(splitPoints, other.splitPoints)
+                .append(partitions, other.partitions)
+                .append(partitionKeyLength, other.partitionKeyLength)
```
m55624: This can be removed after above comment.

store-implementation/parquet-store/src/main/java/uk/gov/gchq/gaffer/parquetstore/partitioner/GroupPartitioner.java
```diff
@@ -226,6 +237,8 @@ public int hashCode() {
         return new HashCodeBuilder(17, 37)
                 .append(group)
                 .append(splitPoints)
+                .append(partitions)
+                .append(partitionKeyLength)
```
m55624: This can be removed after above comment.

store-implementation/parquet-store/src/main/java/uk/gov/gchq/gaffer/parquetstore/partitioner/GroupPartitioner.java
```diff
@@ -36,12 +36,16 @@
 
     private final String group;
     private final List<PartitionKey> splitPoints;
+    private final List<Partition> partitions;
+    private final int partitionKeyLength;
```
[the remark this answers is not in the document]
benj-fletch: should have realised that, I have moved it.

store-implementation/parquet-store/src/main/java/uk/gov/gchq/gaffer/parquetstore/partitioner/GroupPartitioner.java
```diff
@@ -87,16 +89,17 @@ public int getPartitionId(final PartitionKey partitionKey) {
      * @return a list of all the partition ids which could contain data matching the provided partial key
      */
     public List<Integer> getPartitionIds(final Object[] partialKey) {
+        final int partitionKeyLength = splitPoints.isEmpty() ? 0 : splitPoints.get(0).getLength();
```
m55624: If this is moved below the if(0 == splitPoints.size()) (to be changed in below comment) then it can just be set with ``` final int partitionKetLength = splitPoints.get(0).getLength();``` as we know it will not be empty because we have checked above

store-implementation/parquet-store/src/main/java/uk/gov/gchq/gaffer/parquetstore/partitioner/GroupPartitioner.java
```diff
@@ -87,16 +89,17 @@ public int getPartitionId(final PartitionKey partitionKey) {
...
         if (null == partialKey) {
             throw new IllegalArgumentException("getPartitionIds cannot be called with null partialKey");
         }
         if (0 == splitPoints.size()) {
```
m55624: Can this just be changed to if(splitPoints.isEmpty).  This is a safer way to check an empty List.

benj-fletch: I have updated for both this and the above comment

Issue #2009, opened 2018-12-21
Fix required tests for OperationTests


m55624:
the getTestObject should return just a new instance of the Operation being tested.  Further to this any required fields should be added to the getRequiredFields() method.  This is implemented incorrectly in MapTest and others.  To view a correct implementation see CancelScheduledJobTest.

m55624:
Merged into develop.

PR #2010, opened 2018-12-21
GH-2009 - updated OperationTests to properly use getTestObject and getRequiredFields

benj-fletch:
Note I have also removed some duplicated code in the Operation class

### DIFF
--- a/core/operation/src/main/java/uk/gov/gchq/gaffer/operation/Operation.java
+++ b/core/operation/src/main/java/uk/gov/gchq/gaffer/operation/Operation.java
@@ -201,29 +201,11 @@ public interface Operation extends Closeable {
             final Required[] annotations = field.getAnnotationsByType(Required.class);
             if (null != annotations && annotations.length > 0) {
                 if (field.isAccessible()) {
-                    final Object value;
-                    try {
-                        value = field.get(this);
-                    } catch (final IllegalAccessException e) {
-                        throw new RuntimeException(e);
-                    }
-
-                    if (null == value) {
-                        result.addError(field.getName() + " is required for: " + this.getClass().getSimpleName());
-                    }
+                    validateRequiredFieldPresent(result, field);
                 } else {
                     AccessController.doPrivileged((PrivilegedAction<Operation>) () -> {
                         field.setAccessible(true);
-                        final Object value;
-                        try {
-                            value = field.get(this);
-                        } catch (final IllegalAccessException e) {
-                            throw new RuntimeException(e);
-                        }
-
-                        if (null == value) {
-                            result.addError(field.getName() + " is required for: " + this.getClass().getSimpleName());
-                        }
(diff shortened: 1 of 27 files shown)


m55624:
Hi @benj-fletch, thanks for this.  Sorry I forgot to mention any method that overrides a test, does not also need the test annotation again.  For example in AddNamedOperationTest the shouldJsonSerialiseAndDeserialise method has the Override and Test annotation.  This is however not necessary as the parent method already has the Test annotation.  This can cause issues if we ever start loading the tests in via Reflection, as we do with the ITs currently.  The tests will then be run twice.  Can these be updated to reflect this.

benj-fletch:
@m55624 ah I was not aware that was the behaviour! I have had a look for places that both override and have the test annotation and removed the extra @Test annotation

core/operation/src/main/java/uk/gov/gchq/gaffer/operation/Operation.java
```diff
@@ -233,6 +215,19 @@ default ValidationResult validate() {
         return result;
     }
 
+    default void validateRequiredFieldPresent(final ValidationResult result, final Field field) {
```
m55624: Can this be private please.

core/operation/src/test/java/uk/gov/gchq/gaffer/named/operation/AddNamedOperationTest.java
```diff
@@ -55,7 +55,12 @@ public void shouldJSONSerialiseAndDeserialise() throws SerialisationException, J
                 .build();
 
         // When
-        String json = new String(JSONSerialiser.serialise(addNamedOperation, true));
+        String json = null;
+        try {
```
m55624: Can we use the fromJson and toJson methods that are part of the JSONSerialisationTest class.  Similar to how it is written within ForEachTest.  This avoids the need for the SerialisationException bits.

core/operation/src/main/java/uk/gov/gchq/gaffer/operation/Operation.java
benj-fletch: I can't make this private until java 9. Is there a better place you can think of to place the concrete method?

core/operation/src/test/java/uk/gov/gchq/gaffer/named/operation/AddNamedOperationTest.java
benj-fletch: Updated to use those methods

core/operation/src/main/java/uk/gov/gchq/gaffer/operation/Operation.java
m55624: Ah sorry I forgot this was an interface.  No worries thats fine as it is.

PR #2085, opened 2019-02-08
Gh 2073 implement graph hook test interface on all graph hook tests

n288TJYRX:


core/graph/src/main/java/uk/gov/gchq/gaffer/graph/hook/GraphHook.java
```diff
@@ -15,6 +15,7 @@
  */
 package uk.gov.gchq.gaffer.graph.hook;
 
+import com.fasterxml.jackson.annotation.JsonPropertyOrder;
```
m55624: This line shouldn't be necessary

core/graph/src/test/java/uk/gov/gchq/gaffer/graph/hook/AdditionalOperationsTest.java
```diff
@@ -105,6 +109,7 @@ public void assertClonedOperations(final List<Operation> ops1, final List<Operat
 
     @Override
     protected AdditionalOperations getTestObject() {
-        return new AdditionalOperations();
+        return fromJson(ADDITIONAL_OPERATIONS_RESOURCE_PATH);
```
m55624: These tests objects should normally just be a new instance of the class, in the background for some things it does some checks for required fields etc which if they are populated will not throw exceptions correctly.  Can you change this back to just the ```new AdditionalOperations();```

m55624:
Just noticed also there is a checkstyle issue with your code somewhere.  If you run mvn clean install -Panalyze it will fail on the checkstyle issue and you can fix it.  They are normally quite useful errors but any problems let me know.

Issue #2047, opened 2019-01-28
ViewValidator with static methods


GCHQDev404:
ViewValidator probably could change to use static methods.

### Caution
There is FederatedViewValidator which changes/overrides the behaviour of how view validation is applied.

d47853:
From a first glance I think we can label it as such for now. If a beginner is struggling, they can tell us and we can take it on ourselves.

c015dariu:
Hey I've tried to take a stab at this issue myself but I'm getting nullpointers when running OperationChainValidatorTests, whenever it tries to run validateOperationChain using any(Set.class) it always puts in a null to the method, this only happens when it's made static and I can't seem to figure out why.

GCHQ-83497:
@c015dariu It looks like OperationChainValidator uses a ViewValidator.  If we are making that static, I don't think we should be passing that through.

c015dariu:
If you mean via the constructor I fully agree, I have tried removing all references to instanced View Validators and replaced with static calls (and also leaving them and testing that), however my concern is towards the test which utilises Mockito to test the method parameters. I believe Mockito has issues testing static methods as, despite having no change other than now being static (which I've tried calling via instance or via class), it causes mockito to change its behaviour and fail the test. I would replace that line with a test which similarly covers that behaviour but unfortunately I don't fully understand what it does yet.

Line in question:
`given(ViewValidator.validate(any(View.class), any(Schema.class), anySet())).willReturn(new ValidationResult());`

GCHQ-83497:
Yeah I meant the constructor. In response to your concern, using given is attempting to mock the response of validate() so that it returns a new ValidationResult() every time it is called - we shouldn't be mocking statics.

c015dariu:
My current PR is failing on a test that hid from me before committing which is being evasive but after alot of digging around I'm starting to see the issue with having FederatedViewValidator not able to override ViewValidator (thanks to being static), which means the correct ViewValidator is not being used for Federated requests. A quick fix would have been to check the instance of store or the operation chain when passing into OperationChainValidator to identify federated chains/store and then using the appropriate Validator however to access either of these classes requires dependency on Federated-Store which in turn causes a cyclic dependency loop.

I'm running into a brick wall with my understanding of the Gaffer code, is there a better way or place to make this check/fix?

PR #2086, opened 2019-02-08
gh-2047-ViewValidator with static methods

c015dariu:
- Made all methods in ViewValidator Static.
- FederatedViewValidator method is now static, removed @Override tag.
- Replaced all instance calls of ViewValidator with static calls.
- Removed Mockito test on (now) static method which was failing.
- OperationChainValidator no longer takes ViewValidator in the constructor.
- SchemaOperationChainValidator additionally no longer takes ViewValidator in the constructor.
      - Consequently, replaced super(viewValidator) with just super().
#2047

### DIFF
--- a/core/graph/src/main/java/uk/gov/gchq/gaffer/graph/SchemaOperationChainUtil.java
+++ b/core/graph/src/main/java/uk/gov/gchq/gaffer/graph/SchemaOperationChainUtil.java
@@ -24,7 +24,6 @@ import uk.gov.gchq.gaffer.operation.Operations;
 import uk.gov.gchq.gaffer.operation.graph.OperationView;
 import uk.gov.gchq.gaffer.store.SchemaOperationChainValidator;
 import uk.gov.gchq.gaffer.store.schema.Schema;
-import uk.gov.gchq.gaffer.store.schema.ViewValidator;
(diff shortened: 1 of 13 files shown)


m55624:
I have stopped the build due to this branch failing on a checkstyle issue with your code. If you run mvn clean install -Panalyze it will show you the issue. They are normally quite useful errors but any problems let me know.

c015dariu:
#2047 
Hello @m55624, I've managed to get a solution with checkstyle however whichever way I think to solve it, some issues regarding  will crop up.

Checkstyle is not happy with ViewValidator (now being recognized as an utility class) not being final and having a non-default constructor, so I could only see three solutions but there is probably a much easier one I've missed:

1. Remove inheritance between FederatedViewValidator and ViewValidator, give both private constructors and make Federated call validateEdge/Entities statically. This involves making the two methods public rather than protected.
2. Refactor packages to preserve protected methods, I imagine this will break alot of classes.
3. Merge both Federated and View into one, and change the name of Federated's validate method to validateFederated

The current PR involved method 1.

GCHQDev404:
Please see comments by contributor in PR and #2047.
ticket is being deferred.

Issue #2096, opened 2019-02-19
View's are changed after being filtered


m55624:
Currently the view is filled in with the Schema view, based on the GraphConfig view, within the graph _execute method.  It can then be filtered using the UpdateGraphHook.  After this if everything is filtered out to an empty view the view will be again filled in with the full Schema view.  

The second update needs to be removed, and an empty view should just return nothing.

m55624:
View validation for empty/null views can't be added in the Graph class, as this will introduce large breaking changes to the FederatedStore (and others).

m55624:
Merged into develop.

PR #2097, opened 2019-02-19
gh-2096 - update View then run GraphHook after

m55624:


core/graph/src/main/java/uk/gov/gchq/gaffer/graph/Graph.java
```diff
@@ -236,11 +237,31 @@ public JobDetail executeJob(final Operation operation, final Context context) th
         final OperationChain clonedOpChain = request.getOperationChain().shallowClone();
         O result = null;
         try {
+            // This updates the view if empty or null in order for the
```
d47853: Might be worth adding a todo to remove this in V2.

core/graph/src/test/java/uk/gov/gchq/gaffer/graph/GraphTest.java
```diff
@@ -1904,6 +1910,185 @@ public void shouldThrowExceptionOnExecuteJobWithANullUser() throws OperationExce
         }
     }
```
d47853: Is it worth adding a test for when the "addExtraGroups" field is set?

Issue #2031, opened 2019-01-23
Investigate Graph and Store


m55624:
Currently there are two ways to execute, graph.execute and store.execute.  These two methods do slightly different things.  

For example if you execute on the Graph there are methods such as updateOperationChainView which are run, however if store.execute is called these methods are not and cannot be run (because they rely on things like the GraphConfig).

This issue becomes a problem if you run secondary Operations from within Handlers, where there is only access to the Store.

This issue should be addressed and resolved, with one point of entry and one way of execution.

m55624:
Branch relating to this is develop-2.

m55624:
This will be completed in Maestro.

PR #2099, opened 2019-02-25
gh-2031 - move logic out of Graph

m55624:
This also involves removing the reliance of OperationChains in the Graph and the Store.
Adding OperationValidation classes and Interfaces.
GraphHooks now use only GraphRequests, not passing OperationChains and Contexts.


m55624:
Closing for fixing.

Issue #2100, opened 2019-02-25
GetVariable and SetVariable operations don't work properly with FederatedStore


p013570:
The GetVariable and SetVariable operations don't work properly with FederatedStore. This is due to the federated store performing a shallow clone on the user context before sending it to each of the federated graphs. The logic to do the shallow clone does not copy the variables across to the clone.

This is the logic that needs to be updated: https://github.com/gchq/Gaffer/blob/master/core/store/src/main/java/uk/gov/gchq/gaffer/store/Context.java#L62

As a workaround you could execute your operation wrapped inside a FederatedOperationChain.

c015dariu:
Hi, looking into this issue, would this not be as simple as adjusting the constructor to copy over the variables along with the exporters?

I don't believe contexts are ever shallow cloned more than once (or however many sub-graphs in the federation) during their lifetime before they are thrown and new ones created for the next operation chain so I don't think there will be any size issues with this.

m55624:
Hi @c015dariu, yeah this is just as simple as copying the variables over on a shallowClone. Thanks.

c015dariu:
Oops just saw the comment, will work on it now. @m55624 also I can't seem to stop travis builds so it might be worth cancelling my ones.

m55624:
Just FYI this milestone has been update to 1.9 due to an urgent bugfix release for 1.8.1 required and a workaround for this issue makes it less critical.

d47853:
Merged into develop

PR #2107, opened 2019-03-05
gh-2100 - GetVariable and SetVariable operations don't work properly with FederatedStore

c015dariu:
#2100 

One line change within Context constructor that is called via shallowClone()


m55624:
Sorry @c015dariu, just added a comment to the ticket.  Shouldn't be long.

c015dariu:
@m55624 added the test, had to change the context builder because despite allowing the addition of variables, .build() did not actually apply them to the created context. The constructor used in .build() is deprecated however I've left it as it is however with the addition of a small check to see whether to add variables or not.

m55624:
Ah good spot.  Even deprecated it should have been added to the builder correctly.  Cheers.

c015dariu:
@m55624 Just noticed another bug within context regarding addVariables() where it is accidentally referencing the passed in variable for null rather than whether a map currently exists within the object which means it could crash if a variable is added without using set first. Just working on fixing that one and making sure it hasn't broken anything now.

m55624:
Ah yeah, it's missing the "this".  Cheers for that.

c015dariu:
Okay, so I decided to add a couple of tests to cover the variables side of Context and came across an issue where Collections.singletonMap was being used to instantiate variables where none existed, this meant context could only hold one variable in the map and would throw an exception if any others were added. I've now added two tests to cover setVariables and addVariables (previously unused) and fixed the two methods. Uploading once checkstyle and tests have passed on my end.

core/store/src/test/java/uk/gov/gchq/gaffer/store/ContextTest.java
```diff
@@ -141,4 +143,65 @@ public void shouldSetAndGetOriginalOpChain() {
...
+        final Context context = new Context.Builder()
+                .user(user)
+                .config(testConf, "testConfVal")
+                .build();
+        // When
+        context.setVariable("testVar", "testVarVal");
+        context.setVariable("testVar2", "testVarVal2");
+
+        // Then
+        assertTrue(!context.getVariables().isEmpty());
```
m55624: can we change this to assertFalse and remove the not operator.  Just can be slightly confusing to skip read if you miss the not.

core/store/src/test/java/uk/gov/gchq/gaffer/store/ContextTest.java
```diff
@@ -141,4 +143,65 @@ public void shouldSetAndGetOriginalOpChain() {
...
+        final Context context = new Context.Builder()
+                .user(user)
+                .config(testConf, "testConfVal")
+                .build();
+        // When
+        context.addVariables(Collections.singletonMap("testVar", "testVarVal"));
+        context.addVariables(Collections.singletonMap("testVar2", "testVarVal2"));
+
+        // Then
+        assertTrue(!context.getVariables().isEmpty());
```
m55624: Same here

c015dariu:
Requested changes resolved.

PR #2129, opened 2019-03-19
Gh 2072 parameter detail options

g609bmsma:


core/operation/src/test/java/uk/gov/gchq/gaffer/named/operation/AddNamedOperationTest.java
```diff
@@ -52,13 +43,19 @@
 
     @Override
     public void shouldJsonSerialiseAndDeserialise() {
```
d47853: Thanks for updating this test. Could we add another one to check backwards compatibility? ie serialising with no options

core/type/pom.xml
```diff
@@ -64,5 +64,17 @@
             <version>${project.parent.version}</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
```
d47853: Why are these dependencies added?

core/operation/src/test/java/uk/gov/gchq/gaffer/named/operation/AddNamedOperationTest.java
g609bmsma: Done, I've set the options to null and removed them from the expected result.

core/type/pom.xml
g609bmsma: No idea, this seems to have happened automatically? Should I remove the dependency at line 67, or leave it since it seems to not break anything?

g609bmsma: Removed both instances of this dependency

core/operation/src/test/java/uk/gov/gchq/gaffer/named/operation/AddNamedOperationTest.java
```diff
@@ -52,13 +43,56 @@
...
+                " \"overwriteFlag\" : true,%n" +
+                " \"parameters\" : {\"testOption\": {\"description\" :\"Description\", \"defaultValue\": \"On\", \"valueClass\": \"java.lang.String\", \"required\": false, \"options\": [\"option1\", \"option2\", \"option3\"]}},%n" +
+                " \"readAccessRoles\" : [ \"User\" ],%n" +
+                " \"writeAccessRoles\" : [ \"User\" ]%n" +
+                "}"), new String(json));
+        assertNotNull(deserialisedObj);
+    }
+
+    @Test
+    public void shouldJsonSerialiseAndDeserialiseOldVersion() {
```
d47853:
```suggestion
    public void shouldJsonSerialiseAndDeserialiseWithNoOptions() {
```

core/operation/src/main/java/uk/gov/gchq/gaffer/named/operation/ParameterDetail.java
```diff
@@ -37,8 +38,9 @@
     private Object defaultValue;
     private Class valueClass;
     private boolean required;
+    private List<String> options;
```
d47853: Just thinking whether it would be worth leaving this as an unparameterised list. If say a parameter was an integer but could only be 0 - 5, having the options as strings would not be compatible.

g609bmsma: Done, generified types input on list and added a new test in ParameterDetailTest to check it works

Issue #2154, opened 2019-05-09
Scheduled Jobs are unnecessarily hard to create.


d47853:
Scheduled Jobs are overly complicated to create as it requires the user to provide the operation chain as a string.

Instead the user should be allowed to provide an operation in it's java form and it should not be restricted to operation chains. 

This is technically a breaking change, however as the operation is only recently released, it seems justified to change it.

d47853:
Merged into develop

d47853:
## Migration steps
For any REST calls that were previously making use of the scheduled jobs will now use the new format.
The strings they used to represent operation chains can be replaced by json operations
For example:
```json
{
    "class": "Job",
    "opChain": "{ \"operations\": [ { \"class\": \"GetSchema\"}]}"
}
```
Can now be replaced with:
```json
{
    "class": "Job",
    "operation": {
        "class": "GetSchema"
    }
}
```
Anything that was using the java methods will have to change also. There is no method for getting the operation as a string or operation chain. If you want the operation to be guaranteed to be an operation chain, you should use `OperationChain.wrap()`

So anything that was previously:
```java
job.getOpChainAsOperationChain();
```
will need to change to:
```java
job.getOperation();
// or
OperationChain.wrap(job.getOperation());
```

PR #2155, opened 2019-05-09
gh-2154 simplified Job object

d47853:


core/operation/src/main/java/uk/gov/gchq/gaffer/jobtracker/Job.java
```diff
@@ -61,32 +53,48 @@ public void setRepeat(final Repeat repeat) {
...
+        if (null == obj || getClass() != obj.getClass()) {
+            return false;
+        }
+        final Job job = (Job) obj;
+
         try {
-            this.opChain = JSONSerialiser.deserialise(opChain,
-                    OperationChainDAO.class);
+            return new EqualsBuilder()
+                    .append(JSONSerialiser.serialise(operation), JSONSerialiser.serialise(job.operation))
```
GCHQDev404: Is this being used because not all operations implement equals method?

core/operation/src/main/java/uk/gov/gchq/gaffer/jobtracker/Job.java
```diff
@@ -61,32 +53,48 @@ public void setRepeat(final Repeat repeat) {
...
+                    .append(repeat, job.repeat)
+                    .isEquals();
         } catch (final SerialisationException e) {
-            throw new IllegalArgumentException("Unable to deserialise Job OperationChain ", e);
+            throw new RuntimeException("Unable to compare operations as one is not JSON serialisable", e);
```
GCHQDev404: Do we have GafferRuntimeException?

core/operation/src/main/java/uk/gov/gchq/gaffer/jobtracker/Job.java
GCHQDev404: If so see gh-2157

d47853: In short, yes. There are quite a few operations which don't have Equals/Hashcode method so in order to compare them, I used the JSON serialised form since all operations should be json serialisable.

core/operation/src/main/java/uk/gov/gchq/gaffer/jobtracker/Job.java
d47853: Yes. I'll make this change.

Issue #2153, opened 2019-05-09
Scheduled Jobs do not validate operations


d47853:
Scheduled Jobs do not run the same way as other Jobs. In other Jobs If you submit a simple GetAllElements query to an accumulo store eg:
```json
{
    "class": "GetAllElements"
}
```
It works perfectly well. However doing the same thing in as ScheduledJob causes an null pointer exception. This is because there is no view present when the operation enters the handler.

The graph is usually responsible for applying a view but this doesn't seem to happen with a scheduled Job.

d47853:
This issue should either be done together with #2154 or afterwards

d47853:
Merged into develop

PR #2156, opened 2019-05-09
gh-2153 updated operation chain in job using graph

d47853:


d47853:
The todo is only there for review purposes. I will remove it when I merge the other change in

Issue #2209, opened 2019-11-13
Improve error message for unsupported operations in the federated store


d47853:
At present, when handling an unsupported operation, the federated store returns an internal server error with status code 500. There is no descriptive error message. To correct this, the following code could be removed from the FederatedStore class:
```java
    @Override
    protected Object doUnhandledOperation(final Operation operation,
                                          final Context context) {
        throw new UnsupportedOperationException();
    }
```
A test would need to be added to confirm the new behaviour.

saksham93:
I wish to work on this issue.
@d47853 - Should we remove this code or add a meaningful message in the exception? Please suggest.

d47853:
I think as long as the standard error message is informative, you can remove the method.

saksham93:
@d47853 - I agree with you. It is already implemented correctly in super class i.e. Store.java. I have raised the PR for the same.
Thank you.

d47853:
Thanks for your contribution @saksham93 - it looks great. I've added a suggestion to the PR which adds a unit test to confirm the new behaviour. You should be able to commit the suggestion by navigating the the pull request and clicking "commit suggestion" on the comment. Once that's done, we can merge it in 👍

d47853:
Merged into develop

PR #2212, opened 2019-11-19
gh-2209: Improve error message for unsupported operations in the federated store

saksham93:
Removed unnecessary override of method
It is already implemented correctly in Super Class of FederatedStore.java i.e. Store.java

### DIFF
--- a/store-implementation/federated-store/src/test/java/uk/gov/gchq/gaffer/federatedstore/FederatedStoreTest.java
+++ b/store-implementation/federated-store/src/test/java/uk/gov/gchq/gaffer/federatedstore/FederatedStoreTest.java
@@ -271,7 +271,6 @@ public class FederatedStoreTest {
             assertEquals("Operation class uk.gov.gchq.gaffer.operation.impl.OperationImpl is not supported by the FederatedStore.", e.getMessage());
         }
     }
-    
     @Test
     public void shouldAlwaysReturnSupportedTraits() throws Exception {
         // Given
--- a/store-implementation/federated-store/src/test/java/uk/gov/gchq/gaffer/federatedstore/FederatedStoreTest.java
+++ b/store-implementation/federated-store/src/test/java/uk/gov/gchq/gaffer/federatedstore/FederatedStoreTest.java
@@ -255,7 +255,22 @@ public class FederatedStoreTest {
             assertContains(e, "GraphId: ", ACC_ID_2);
         }
     }
-
+    @Test
+    public void shouldThrowAppropriateExceptionWhenHandlingAnUnsupportedOperation() {
+        // Given
+        Operation op = new OperationImpl();
+        // When
+        try {
+            store.handleOperation(op, new Context());
+            fail("Exception expected");
+        } catch (final OperationException e) {
+            fail("Expected an UnsupportedOperationException rather than an OperationException");
+        } catch (final UnsupportedOperationException e) {
+            // Then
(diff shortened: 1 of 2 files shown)


store-implementation/federated-store/src/test/java/uk/gov/gchq/gaffer/federatedstore/FederatedStoreTest.java
```diff
@@ -256,11 +256,6 @@ public void shouldNotAllowOverwritingOfGraphWithinFederatedScope() throws Except
         }
     }
```
d47853: In order to replace the outgoing test, it would be good to add something like:
```suggestion
    @Test
    public void shouldThrowAppropriateExceptionWhenHandlingAnUnsupportedOperation() {
        // Given
        Operation op = new OperationImpl();
        // When
        try {
            store.handleOperation(op, new Context());
            fail("Exception expected");
        } catch (final OperationException e) {
            fail("Expected an UnsupportedOperationException rather than an OperationException");
        } catch (final UnsupportedOperationException e) {
            // Then
            assertEquals("Operation class uk.gov.gchq.gaffer.operation.impl.OperationImpl is not supported by the FederatedStore.", e.getMessage());
        }
    }
```

store-implementation/federated-store/src/test/java/uk/gov/gchq/gaffer/federatedstore/FederatedStoreTest.java
```diff
@@ -255,12 +255,22 @@ public void shouldNotAllowOverwritingOfGraphWithinFederatedScope() throws Except
...
-
-    @Test(expected = UnsupportedOperationException.class)
-    public void shouldDoUnhandledOperation() throws Exception {
-        store.doUnhandledOperation(null, null);
+    @Test
+    public void shouldThrowAppropriateExceptionWhenHandlingAnUnsupportedOperation() {
+        // Given
+        Operation op = new OperationImpl();
```
d47853: Ah - just spotted that you'll need to add some imports to get this to work

d47853: Add 
`import uk.gov.gchq.gaffer.operation.impl.OperationImpl;` between the OperationException and AddElements imports to get this to compile

saksham93: Should be fine now

store-implementation/federated-store/src/test/java/uk/gov/gchq/gaffer/federatedstore/FederatedStoreTest.java
```diff
@@ -255,12 +256,22 @@ public void shouldNotAllowOverwritingOfGraphWithinFederatedScope() throws Except
...
+            fail("Exception expected");
+        } catch (final OperationException e) {
+            fail("Expected an UnsupportedOperationException rather than an OperationException");
+        } catch (final UnsupportedOperationException e) {
+            // Then
+            assertEquals("Operation class uk.gov.gchq.gaffer.operation.impl.OperationImpl is not supported by the FederatedStore.", e.getMessage());
+        }
     }
-
+
```
d47853: Thanks for adding that. It compiles perfectly. Only reason checkstyle is failing is because this line has trailing spaces. Removing this should fix it
```suggestion

```

Issue #2214, opened 2019-11-27
FederatedAggregateHandler should Aggregate Duplicate Elements From Diff Stores 


GCHQDev404:

d47853:
Merged into develop

PR #2215, opened 2019-11-28
gh-2214 FederatedAggregateHandler should Aggregate Duplicate Elements From Diff Stores

GCHQDev404:


store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/operation/handler/FederatedAggregateHandler.java
```diff
@@ -40,6 +42,10 @@ public FederatedAggregateHandler(final AggregateHandler handler) {
                                                    final Context context,
                                                    final Store store)
             throws OperationException {
-        return handler.doOperation(operation, ((FederatedStore) store).getSchema(operation, context));
+        try {
+            return handler.doOperation(operation, ((FederatedStore) store).getSchema(operation, context));
+        } catch (SchemaException e) {
```
d47853:
```suggestion
        } catch (final SchemaException e) {
```

PR #2217, opened 2019-12-11
Gh 2216 transform nullpointer bug

GCHQDev404:


core/store/src/main/java/uk/gov/gchq/gaffer/store/operation/util/TransformStreamSupplier.java
```diff
@@ -56,8 +60,22 @@ public void close() throws IOException {
...
+            final Element rtn;
+            if (e instanceof Edge && nonNull(edgesTransform) && nonNull(group)) {
+                rtn = edgesTransform.get(group).apply(e);
+            } else if (e instanceof Entity && nonNull(entitiesTransform) && nonNull(group)) {
+                rtn = entitiesTransform.get(group).apply(e);
+            } else if (e instanceof Edge || e instanceof Entity) {
+                //either the group or transformer was null for the Edge/Entity.
+                rtn = e;
+            } else {
+                throw new UnsupportedOperationException("Transform is only compatible with Edge or Entity elements, found: " + e.getClass());
```
t11947: Might it be worth documenting this exception in the javadoc?

core/store/src/main/java/uk/gov/gchq/gaffer/store/operation/util/TransformStreamSupplier.java
```diff
@@ -54,10 +58,24 @@ public void close() throws IOException {
...
+
+            if (e instanceof Edge && nonNull(edgesTransform) && nonNull(group)) {
+                rtn = edgesTransform.get(group).apply(e);
+            } else if (e instanceof Entity && nonNull(entitiesTransform) && nonNull(group)) {
+                rtn = entitiesTransform.get(group).apply(e);
+            } else if (e instanceof Edge || e instanceof Entity) {
+                //either the group or transformer was null for the Edge/Entity.
+                rtn = e;
+            } else {
+                throw new UnsupportedOperationException("Transform is only compatible with Edge or Entity elements, found: " + e.getClass());
```
d47853: This is probably the wrong exception type. UnsupportedOperationExceptions tend to only be thrown when the user tries to execute an operation which is not supported, as opposed to when the input for an operation is the wrong type. An IllegalArgumentException may be more appropriate.

PR #2221, opened 2020-01-09
Gh 2220 add function hook

d47853:


core/graph/src/main/java/uk/gov/gchq/gaffer/graph/hook/FunctionAuthoriser.java
```diff
@@ -0,0 +1,209 @@
...
+                .map(Pattern::asPredicate)
+                .collect(Collectors.toList());
+    }
+
+    private void checkAllFunctionsUsedAppearInPatterns(final String chainString) {
+        boolean noWhitelistPatterns = (authorisedFunctionPatterns == null || authorisedFunctionPatterns.isEmpty());
+        boolean noBlacklistPatterns = (unauthorisedFunctionPatterns == null || unauthorisedFunctionPatterns.isEmpty());
+
+        Predicate<String> isAuthorised = createAuthorisationFunction(noBlacklistPatterns, noWhitelistPatterns);
+
```
ss130257: Matching of the "class":" string feels a bit fragile to me. You may want to consider creating a TreeModel from the Json string and traversing the tree of JsonNode objects checking for "class" fields.

core/graph/src/main/java/uk/gov/gchq/gaffer/graph/Graph.java
```diff
@@ -995,6 +1000,11 @@ private void updateGraphHooks(final GraphConfig config) {
                     config.getHooks().add(0, new NamedOperationResolver());
                 }
             }
+            if (!hasFunctionAuthoriserHook && !config.isSkipDefaultSecurityHooks()) {
+                config.getHooks().add(new FunctionAuthoriser.Builder()
+                        .unauthorisedFunctions(Lists.newArrayList(CreateObject.class))
```
p013570: Perhaps create a default black list and store that in another class, otherwise this list here could get rather big and messy.

core/graph/src/main/java/uk/gov/gchq/gaffer/graph/GraphConfig.java
```diff
@@ -64,6 +65,9 @@
     private String description;
     private List<GraphHook> hooks = new ArrayList<>();
 
+    @JsonInclude(JsonInclude.Include.NON_DEFAULT)
+    private boolean skipDefaultSecurityHooks = false;
```
p013570: I'm not sure about this.

core/graph/src/main/java/uk/gov/gchq/gaffer/graph/Graph.java
```diff
@@ -995,6 +1000,11 @@ private void updateGraphHooks(final GraphConfig config) {
                     config.getHooks().add(0, new NamedOperationResolver());
                 }
             }
+            if (!hasFunctionAuthoriserHook && !config.isSkipDefaultSecurityHooks()) {
```
p013570: If we are adding this FunctionAuthoriser by default for every Gaffer system, we should make sure it is efficient and doesn't error.

core/graph/src/main/java/uk/gov/gchq/gaffer/graph/hook/FunctionAuthoriser.java
```diff
@@ -0,0 +1,216 @@
...
+ * efficient than using any of the patterns.
+ */
+@JsonPropertyOrder(value = {"unauthorisedFunctions", "unauthorisedFunctionPatterns", "authorisedFunctionPatterns"})
+@JsonInclude(JsonInclude.Include.NON_DEFAULT)
+public class FunctionAuthoriser implements GraphHook {
+
+    private static final String ERROR_MESSAGE_PREFIX = "Operation chain contained an unauthorised function: ";
+    private static final Logger LOGGER = LoggerFactory.getLogger(FunctionAuthoriser.class);
+    private List<Class<? extends Function>> unauthorisedFunctions = new ArrayList<>();
+    private List<Pattern> authorisedFunctionPatterns = new ArrayList<>();
```
p013570: Just wondering if it might be simpler to just call these white and black lists? As you refer to white and black lists in other parts of the code.

core/graph/src/main/java/uk/gov/gchq/gaffer/graph/hook/FunctionAuthoriser.java
p013570: Rather than going to String then to JsonNode, you might be able to use the valueToTree method on ObjectMapper. Just had a quick look at it and it should work. You would need to expose this in the Gaffer JSONSerialiser class.

core/graph/src/main/java/uk/gov/gchq/gaffer/graph/hook/FunctionAuthoriser.java
```diff
@@ -0,0 +1,216 @@
...
+        try {
+            node = JSONSerialiser.getJsonNodeFromString(chainString);
+        } catch (final SerialisationException e) {
+            // This should never happen as the string is derived from a
+            // serialised object. Therefore throw an exception here.
+            throw new RuntimeException("Failed to convert serialised operation" +
+                    " chain into a JsonNode", e);
+        }
+
+        List<String> classNames = node.findValuesAsText("class");
```
p013570: The downside of using this mechanism, is you are assuming that all Functions will be serialised with a 'class' field. They should be, but if they are wrapped in some custom object and the JsonSubType annotation changes it, then it could be serialised differently.

core/graph/src/main/java/uk/gov/gchq/gaffer/graph/GraphConfig.java
d47853: This makes sure that any existing graph config which doesn't have the field set will have the value set to false. Is there something I overlooked?

core/graph/src/main/java/uk/gov/gchq/gaffer/graph/hook/FunctionAuthoriser.java
d47853: I agree it's probably better to be consistent, although when creating this in JSON, it makes it more understandable in my opinion - especially given that there are two whitelists.

core/graph/src/main/java/uk/gov/gchq/gaffer/graph/hook/FunctionAuthoriser.java
d47853: Would this get around the issue of being reliant on things being serialised as "class" fields?

core/graph/src/main/java/uk/gov/gchq/gaffer/graph/hook/FunctionAuthoriser.java
d47853: Not sure how you could do this without doing expensive serialisation

core/graph/src/main/java/uk/gov/gchq/gaffer/graph/hook/FunctionAuthoriser.java
d47853: From what I can tell, this would only provide a small performance benefit if you don't make use of the more efficient blacklisting capability (Since the blacklisting capability requires it to be a string anyway).

core/graph/src/main/java/uk/gov/gchq/gaffer/graph/GraphConfig.java
p013570: I am just not sure whether having a boolean to toggle security hooks on/off is the best way of doing it. Although I can't quickly think of a better way. For example, OperationAuthoriser and OperationChainLimiter are also security related hooks.

Issue #2228, opened 2020-01-22
update copyright headers


d47853:

d47853:
Merged into develop

PR #2229, opened 2020-01-22
gh-2228 updated copyright

d47853:


core/common-util/src/test/java/uk/gov/gchq/gaffer/commonutil/StreamUtilTest.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2017-2019 Crown Copyright
+ * Copyright 2017-2020 Crown Copyright
```
GCHQDev404: Is this correct ?

PR #2238, opened 2020-01-24
Federated admin changes

GCHQDev404:


store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/operation/ChangeGraphAccess.java
```diff
@@ -0,0 +1,169 @@
...
+
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Set;
+
+import static uk.gov.gchq.gaffer.federatedstore.FederatedStoreConstants.KEY_OPERATION_OPTIONS_GRAPH_IDS;
+
+
+@JsonPropertyOrder(value = {"class", "graphId"}, alphabetic = true)
```
d47853: Do you need the other fields here?

GCHQDev404: Not necessary as all will be displayed alphabetically, but I've added some more in.

PR #2260, opened 2020-03-16
Gh 2259 FederatedStore mitigate infinite loop

GCHQDev404:


store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/FederatedStore.java
```diff
@@ -334,10 +350,23 @@ public Schema getSchema(final Map<String, String> config, final User user) {
      *
      * @param user        the users scope to get graphs for.
      * @param graphIdsCsv the csv of graphIds to get, null returns all graphs.
+     * @param operation   the requesting operation.
      * @return the graph collection.
      */
-    public Collection<Graph> getGraphs(final User user, final String graphIdsCsv) {
```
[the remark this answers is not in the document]
GCHQDev404: I don't think it would or should be used anywhere.
The graphIdsCsv is optional value from operations during federation, ignoring this will ignore the users request to scope the operation.
Removing the the Operation parameter will remove the attempt to mitigate infinite loops, making this ticket change potentially useless.

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/FederatedStore.java
```diff
@@ -102,9 +111,21 @@
...
     private Boolean isPublicAccessAllowed = Boolean.valueOf(IS_PUBLIC_ACCESS_ALLOWED_DEFAULT);
+    private static final List<Integer> ALL_IDS = new ArrayList<>();
+    private final int id;
+
+    public FederatedStore() {
+        Integer i = null;
+        while (isNull(i) || ALL_IDS.contains(i)) {
+            i = new Random().nextInt();
+        }
+        ALL_IDS.add(id = i);
```
d47853: Does this mean that two FederatedStores in seperate JVMs could have the same id?

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/FederatedStore.java
```diff
@@ -328,16 +349,45 @@ public Schema getSchema(final Map<String, String> config, final User user) {
...
      */
-    public Collection<Graph> getGraphs(final User user, final String graphIdsCsv) {
-        return graphStorage.get(user, getCleanStrings(graphIdsCsv));
+    public Collection<Graph> getGraphs(final User user, final String graphIdsCsv, final Operation operation) {
+        Collection<Graph> rtn = new ArrayList<>();
+        if (nonNull(operation)) {
+            String optionKey = FEDERATED_STORE_PROCESSED + id;
+            boolean isIdFound = !operation.getOption(optionKey, "").isEmpty();
+            if (!isIdFound) {
+                HashMap<String, String> updatedOptions = isNull(operation.getOptions()) ? new HashMap<>() : new HashMap<>(operation.getOptions());
```
d47853: Does operation.getOpertions() need wrapping here?

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/FederatedStore.java
```diff
@@ -328,16 +349,45 @@ public Schema getSchema(final Map<String, String> config, final User user) {
...
+                updatedOptions.put(optionKey, getGraphId());
+                operation.setOptions(updatedOptions);
```
d47853: Why not just use operation.addOption()?

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/FederatedStore.java
```diff
@@ -102,9 +111,21 @@
...
     private FederatedGraphStorage graphStorage = new FederatedGraphStorage();
     private Set<String> customPropertiesAuths;
     private Boolean isPublicAccessAllowed = Boolean.valueOf(IS_PUBLIC_ACCESS_ALLOWED_DEFAULT);
+    private static final List<Integer> ALL_IDS = new ArrayList<>();
+    private final int id;
+
+    public FederatedStore() {
+        Integer i = null;
+        while (isNull(i) || ALL_IDS.contains(i)) {
+            i = new Random().nextInt();
```
t11947: Is it worth hoisting the Random object to stop it being created every loop, and instantiating i to a random int off the bat to prevent the need for a null check?

GCHQDev404: 99% of the time the loop will happen once.

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/FederatedStore.java
GCHQDev404: Yes, but collision is chance is very low. This is to mitigate against infinite loops not guarantee it doesn't ever happen. Next time one of the JVM is turned on and off again it should be okay.

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/FederatedStore.java
```diff
@@ -328,16 +349,45 @@ public Schema getSchema(final Map<String, String> config, final User user) {
...
      * </p>
      *
      * @param user        the users scope to get graphs for.
      * @param graphIdsCsv the csv of graphIds to get, null returns all graphs.
+     * @param operation   the requesting operation, graphs are returned only once per operation.
      * @return the graph collection.
      */
-    public Collection<Graph> getGraphs(final User user, final String graphIdsCsv) {
-        return graphStorage.get(user, getCleanStrings(graphIdsCsv));
+    public Collection<Graph> getGraphs(final User user, final String graphIdsCsv, final Operation operation) {
```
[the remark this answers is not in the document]
GCHQDev404: If that is done, then all infinate loop mitigation can be completely bypassed.
You will be completely reliant on at least 1 FederatedStore in the loop being standard with standard handlers.

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/FederatedStore.java
GCHQDev404: I don't follow?

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/FederatedStore.java
GCHQDev404: Coz I was looking for a `put()` method which wasn't there since I was dealing with properties map.

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/FederatedStore.java
```diff
@@ -102,9 +111,21 @@
  * @see Graph
  */
 public class FederatedStore extends Store {
+    private static final Logger LOGGER = LoggerFactory.getLogger(Store.class);
+    private static final String FEDERATED_STORE_PROCESSED = "FederatedStore.processed.";
     private FederatedGraphStorage graphStorage = new FederatedGraphStorage();
     private Set<String> customPropertiesAuths;
     private Boolean isPublicAccessAllowed = Boolean.valueOf(IS_PUBLIC_ACCESS_ALLOWED_DEFAULT);
+    private static final List<Integer> ALL_IDS = new ArrayList<>();
```
t11947: Consider not using constant case for mutable lists.

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/FederatedStore.java
t11947: Is there a reason you're not assigning the value right away rather than assigning null and then checking for null?

GCHQDev404: Because is there is collision I will need to rewrite the logic to assign the value again. duplicate line of code.

```
variable = assign value logic
check(variable)
variable = re-assign logic again // duplicate code.
```

t11947: If you're worried about code duplication, consider using a private method. Is there a reason you're not assigning to the field directly from the get go?

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/FederatedStore.java
d47853: Sorry my question wasn't very clear. I was more referring to the final part of the ternary where you wrap the options in a hashmap. I don't think it needs the extra wrap as you've already don a null check.

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/FederatedStore.java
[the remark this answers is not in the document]
d47853: Ah okay. Fair enough. To be safe though, it would be a good idea to re-add it and deprecate it with a comment on why it shouldn't be used.

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/FederatedStore.java
GCHQDev404: regarding `new HashMap<>(operation.getOptions())`?
This is because the operations interface makes each operation responsible for creating and returning the options map. It might be immutable.

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/FederatedStore.java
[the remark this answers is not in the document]
GCHQDev404: Any handers already written will continue to use the deprecated by passed method.
And will not be changed until the deprecation is removed.
I prefer to force the change to make them update. Are you happy with mitigation reliant on 1 FederatedStore?

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/FederatedStore.java
GCHQDev404:
```
variable = methodAssingment()
check(variable) {
variable = methodAssingment() // still a duplicate line

methodAssingment(){ /*now there permenant  */ }
}
```

GCHQDev404:
```
variable = methodAssingment()
check(variable) {
variable = methodAssingment() // still a duplicate line

methodAssingment(){ 
    //now an extra method with a one liner sitting in the class used only once 99.99% of instantiation.
}
```

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/FederatedStore.java
GCHQDev404: This is global list held and modified by all classes, to understand and know which ID number have been used by other classes.

What would you consider using?

t11947: Camel case for mutable class members.

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/FederatedStore.java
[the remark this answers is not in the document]
d47853: I guess as the Federated Store is labeled experimental and this mitigates a serious bug, as you say, it's probably better to force the change.

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/FederatedStore.java
GCHQDev404: OOooooooh completely misunderstood  you.

Issue #2262, opened 2020-03-25
Schema.fromJson() does not support schema directories containing subdirectories


ctas582:
I've been trying to run the [RoadTrafficDataLoader](https://github.com/gchq/Gaffer/blob/master/example/road-traffic/road-traffic-generators/src/main/java/uk/gov/gchq/gaffer/traffic/generator/RoadTrafficDataLoader.java), loading the schema from a directory which contains subdirectories. This results in an exception:
```
Exception in thread "main" uk.gov.gchq.gaffer.data.elementdefinition.exception.SchemaException: Failed to load element definitions from bytes
	at uk.gov.gchq.gaffer.data.elementdefinition.ElementDefinitions$BaseBuilder.json(ElementDefinitions.java:323)
	at uk.gov.gchq.gaffer.data.elementdefinition.ElementDefinitions$BaseBuilder.json(ElementDefinitions.java:285)
	at uk.gov.gchq.gaffer.store.schema.Schema$BaseBuilder.json(Schema.java:531)
	at uk.gov.gchq.gaffer.store.schema.Schema.fromJson(Schema.java:116)
	at uk.gov.gchq.gaffer.traffic.generator.RoadTrafficDataLoader.main(RoadTrafficDataLoader.java:106)
Caused by: java.io.IOException: Is a directory
	at sun.nio.ch.FileDispatcherImpl.read0(Native Method)
	at sun.nio.ch.FileDispatcherImpl.read(FileDispatcherImpl.java:46)
	at sun.nio.ch.IOUtil.readIntoNativeBuffer(IOUtil.java:223)
	at sun.nio.ch.IOUtil.read(IOUtil.java:197)
	at sun.nio.ch.FileChannelImpl.read(FileChannelImpl.java:159)
	at sun.nio.ch.ChannelInputStream.read(ChannelInputStream.java:65)
	at sun.nio.ch.ChannelInputStream.read(ChannelInputStream.java:109)
	at sun.nio.ch.ChannelInputStream.read(ChannelInputStream.java:103)
	at java.nio.file.Files.read(Files.java:3105)
	at java.nio.file.Files.readAllBytes(Files.java:3158)
	at uk.gov.gchq.gaffer.data.elementdefinition.ElementDefinitions$BaseBuilder.json(ElementDefinitions.java:314)
	... 4 more
```

[...]

PR #2263, opened 2020-03-26
gh-2262 Schema.fromJson should support traversing nested directories …

m29827:
…when processing a path.


core/store/src/test/java/uk/gov/gchq/gaffer/store/schema/SchemaTest.java
```diff
@@ -451,23 +452,62 @@ public void testSchemaConstructedFromInputStream() throws IOException {
...
+    }
+
+    @Test
+    public void testSchemaConstructedByRecursiveDirectoryProcessingOfPath() throws Exception {
+
+        final Schema schemaNested = Schema.fromJson(Paths.get(SchemaTest.class.getResource("/schema-nested").toURI()));
+
+        assertExpectedSchemaElementsContent(schemaNested);
+        assertExpectedSchemaSerialisationContent(schema);
+        assertExpectedSchemaTypesContent(schema);
```
d47853: Should these tests be using schemaNested instead of schema?

Issue #2257, opened 2020-03-10
Create an operation to set split points on an Accumulo store from an RDD or JavaRDD of Elements


gaffer01:
It's possible to import an `RDD` of `Element`s to an Accumulo store, but there's no easy way to set the split points on a store given an RDD. We should create an operation for adding the split points to a store given an `RDD<Element>` or `JavaRDD<Element>` and write a handler for this operation for the Accumulo store.

The code will be something like:

    JavaRDD<Element> elements = ...
    byte[] schemaAsJson = schema.toCompactJson();
    FlatMapFunction<Iterator<Element>, Pair<Key,Key>> generateKeys = new FlatMapFunction<>() {
        AccumuloElementConverter converter;
        @Override
        public Iterator<Pair<Key,Key>> call(Iterator<Element> elementIterator) throws Exception {
             converter = Class.forName(converterClassName).asSubclass(AccumuloElementConverter.class)
                   .getConstructor(Schema.class)
                   .newInstance(Schema.fromJson(schemaAsjson);
             return new Iterator<Pair<Key,Key>> {
                  @Override
                  public boolean hasNext() { return elementIterator.hasNext(); }
                  public Pair<Key,Key> next() { return converter.getKeysFromElement(elementIterator.next());
        }
    };
    JavaRDD<Text> rows = elements.mapPartitions(generateKeys)
        .flatMap(pair -> {
              if (null == pair.getSecond()) {return Arrays.asList(pair.getFirst()).iterator();}
              else {return Arrays.asList(pair.getFirst(), pair.getSecond()).iterator();}
        })
        .map(key -> key.getRow());

This RDD can now be sampled using `.sample()` and then collected to a local list. This list can then be used to set the split points on the table. It may be necessary to count the number of rows to determine an appropriate sampling proportion so that approximately the right number of split points can be sampled.

m29827:
I started writing some code based on the suggested approach: extract keys from an RDD of elements, flatten and then sample to collect a list of keys.  Then I wondered how best to convert that list to split points and came across the SampleElementsForSplitPoints operation which made me wonder whether sampling the RDD of elements directly and using the existing operation would be a good option?  Alternatively I could stick with the proposed approach and pull out a new operation that generates split points from a list of keys, reusing the existing code in SampleElementsForSplitPoints.

d47853:
Merged into develop

PR #2264, opened 2020-04-01
gh-2257 Set Accumulo store split points based on sampled elements from RDD's

m29827:


m29827:
Closing due to broken tests and check-style failures

PR #2267, opened 2020-04-14
Gh 791 add named operation group label (part of tools Issue #791)

macenturalxl1:
Hi - here is the back end change for Issue #791 (in gaffer-tools) - it's ready to go in to develop anytime by itself.

This backend feature in this PR is just enabling NamedOperationDetail to include a label and it is safe to merge anytime - as I'm still working on the gaffer-tools UI to send the request down.

It is compatible with the current request that gaffer-tools and Swagger sends (i.e. it sends AddNamedOperation request with "label" missing from body - the backend will read this as null).

Let me know what you think, thanks

### DIFF
--- a/core/operation/src/main/java/uk/gov/gchq/gaffer/named/operation/AddNamedOperation.java
+++ b/core/operation/src/main/java/uk/gov/gchq/gaffer/named/operation/AddNamedOperation.java
@@ -58,7 +58,7 @@ public class AddNamedOperation implements Operation, Operations<Operation> {
     @Required
     private String operations;
     private String operationName;
-    private String label;
+    private List<String> labels;
     private String description;
     private List<String> readAccessRoles = new ArrayList<>();
     private List<String> writeAccessRoles = new ArrayList<>();
@@ -124,12 +124,12 @@ public class AddNamedOperation implements Operation, Operations<Operation> {
         this.operationName = operationName;
     }
 
-    public String getLabel() {
-        return label;
+    public List<String> getLabels() {
+        return labels;
     }
 
-    public void setLabel(final String label) {
-        this.label = label;
+    public void setLabels(final List<String> labels) {
+        this.labels = labels;
     }
 
     public List<String> getReadAccessRoles() {
@@ -169,7 +169,7 @@ public class AddNamedOperation implements Operation, Operations<Operation> {
         return new AddNamedOperation.Builder()
                 .operationChain(operations)
                 .name(operationName)
-                .label(label)
+                .labels(labels)
                 .description(description)
                 .readAccessRoles(readAccessRoles.toArray(new String[readAccessRoles.size()]))
                 .writeAccessRoles(writeAccessRoles.toArray(new String[writeAccessRoles.size()]))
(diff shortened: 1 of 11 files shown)


core/operation/src/main/java/uk/gov/gchq/gaffer/named/operation/AddNamedOperation.java
```diff
@@ -54,9 +54,11 @@
 @Since("1.0.0")
 @Summary("Adds a new named operation")
 public class AddNamedOperation implements Operation, Operations<Operation> {
+
     @Required
     private String operations;
     private String operationName;
+    private String label;
```
d47853, line 61: It might be better to have a list of labels so that named operations can have more than one.

core/operation/src/main/java/uk/gov/gchq/gaffer/named/operation/NamedOperationDetail.java
```diff
@@ -58,15 +59,15 @@
     public NamedOperationDetail() {
     }
 
-    public NamedOperationDetail(final String operationName, final String description, final String userId,
+    public NamedOperationDetail(final String operationName, final String label, final String description, final String userId,
```
d47853, line 62: For backwards compatability, these constructors should remain the same and instead call a new constructor which has the label as the final parameter.

So instead of adding a label in all the constructors:
```java
public Example(String example1, String Label, String example2) {
    this(example1, label, example2, null);
}

public Example(String example1, String Label, String example2, Integer example3) {
    // do stuff
}
```
Leave the constructors as they are so any existing code still call them and they maintain the same behaviour:
```java
public Example(String example1, String example2) {
    this(example1, example2, null, null);
}

public Example(String example1, String example2, Integer example3) {
    this(example1, example2, example3, null);
}

// new constructor
public Example(String example1, String example2, Integer example3, String label) {
    // do stuff
}
```

core/operation/src/main/java/uk/gov/gchq/gaffer/named/operation/AddNamedOperation.java
macenturalxl1: Ah yes it did cross my mind. Having an Operation linked to many groups would be a better feature - and a lot more flexible - I'll make the change thanks

core/operation/src/main/java/uk/gov/gchq/gaffer/named/operation/NamedOperationDetail.java
macenturalxl1: Agreed, thanks for the recommended solution

core/operation/src/main/java/uk/gov/gchq/gaffer/named/operation/AddNamedOperation.java
macenturalxl1: Done

core/operation/src/main/java/uk/gov/gchq/gaffer/named/operation/NamedOperationDetail.java
macenturalxl1: Done

macenturalxl1:
Done the requested change so NamedOperations can handle a list of labels - just waiting on the queued Travis pipeline to finish

Issue #2216, opened 2019-12-11
Bug: No Message for java.lang.NullPointerException when not implementing transforms for all groups in query


c95560:
If you perform a transform as part of a view or operation, then you will encounter a java.lang.NullPointerException error if there is a partial mapping from the groups of incoming elements to transformFunctions. The globalElements, globalEntities, globalEdges properties are not affected by this issue.

If the user does not know what element groups are going to present at query runtime before a transform, then they are forced to specify a tranformFunction for all groups in the schema.

See lack of null checks around line 59: https://github.com/gchq/Gaffer/blob/06ea720b482f8be0d4a5cf4def59a95906615cee/core/store/src/main/java/uk/gov/gchq/gaffer/store/operation/util/TransformStreamSupplier.java

The following JSON will reproduce the error for road-traffic-demo:
https://github.com/gchq/Gaffer/blob/06ea720b482f8be0d4a5cf4def59a95906615cee/example/road-traffic/scripts/start.sh

`
{
	"class": "OperationChain",
	"operations": [{
			"class": "uk.gov.gchq.gaffer.operation.impl.get.GetAllElements"
		},
		{
			"class": "Transform",
			"entities": {
				"Cardinality": {
					"functions": [{
						"selection": [
							"count"
						],
						"function": {
							"class": "ToString"
						},
						"projection": [
							"countStringB"
						]
					}]
				}
			}
		},
		{
			"class": "Filter",
			"entities": {
				"Cardinality": {
					"predicates": [{
						"selection": [
							"count"
						],
						"predictate": {
							"class": "IsMoreThan",
							"orEqualTo": false,
							"value": {
								"Long": 580
							}
						}
					}]
				}
			}
		}
	]
}
`

d47853:
Merged into develop

PR #2269, opened 2020-04-20
 gh-2216 Scheduled Jobs will not reschedule following restart.

m29827:


core/operation/src/main/java/uk/gov/gchq/gaffer/jobtracker/JobDetail.java
```diff
@@ -60,36 +66,37 @@ public JobDetail(final JobDetail oldJobDetail, final JobDetail newJobDetail) {
         }
     }
 
-    public JobDetail(final String jobId, final String userId, final OperationChain<?> opChain, final JobStatus jobStatus, final String description) {
-        this(jobId, null, userId, opChain, jobStatus, description);
+    public JobDetail(final String jobId, final User user, final OperationChain<?> opChain, final JobStatus jobStatus, final String description) {
```
d47853, line 69: This will change the public API which would mean a breaking change. We've committed not to introduce breaking changes in version 1.x. Would it be possible to revert these and add some new constructors.
```suggestion
    @Deprecated
    public JobDetail(final String jobId, final String userId, final OperationChain<?> opChain, final JobStatus jobStatus, final String description) {
        this(jobId, null, new User(userId), opChain, jobStatus, description);
    }
    // other constructors following same pattern
    public JobDetail(final String jobId, final String parentJobId, final User user, final OperationChain<?> opChain, final JobStatus jobStatus, final String description) {   
         // do stuff
     }
```

core/operation/src/main/java/uk/gov/gchq/gaffer/user/User.java
```diff
@@ -34,7 +35,8 @@
  * uses the values of the dataAuths and opAuths objects associated with the user
  * to determine if they are allowed to perform the operation or not.
  */
-public class User {
+public class User implements Serializable {
+
```
d47853, line 39: Add a serialVersionUID

m29827: Added serialVersionUID

core/operation/src/main/java/uk/gov/gchq/gaffer/jobtracker/JobDetail.java
m29827: Re-instated deleted code and marked @Deprecated

PR #2275, opened 2020-04-30
Gh 2270 JUnit5 upgrade (modules: cache, common-utils, data, exception)

macenturalxl1:
Hi - here is the first iteration of the JUnit5 upgrade ticket.

I'll break the root level modules in to manageable batches of work so that it hopefully makes PRs easier to review (as oppose to one large piece). I'll make a note of this on the ticket too.

Here is the first 4x modules under core - I'll raise another PR for the second x4 (graph, operation, serialisation, store) as a part 2.

### DIFF
--- a/core/common-util/src/test/java/uk/gov/gchq/gaffer/commonutil/JsonAssert.java
+++ b/core/common-util/src/test/java/uk/gov/gchq/gaffer/commonutil/JsonAssert.java
@@ -17,12 +17,14 @@
 package uk.gov.gchq.gaffer.commonutil;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
-import org.junit.Assert;
 
 import java.io.IOException;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotEquals;
 
 public final class JsonAssert {
     private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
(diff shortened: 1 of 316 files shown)


library/flink-library/src/test/java/uk/gov/gchq/gaffer/flink/operation/FlinkTest.java
```diff
@@ -122,7 +122,7 @@ public static Store createStore() {
 
     public static void verifyElements(final Graph graph) throws OperationException, InterruptedException {
         // Wait for the elements to be ingested.
-        Thread.sleep(2000);
+        Thread.sleep(3000);
```
d47853, line 125: Is this failing at the moment?

core/common-util/src/test/java/uk/gov/gchq/gaffer/commonutil/JsonAssert.java
```diff
@@ -17,25 +17,27 @@
...
+
 public final class JsonAssert {
     private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
 
     private JsonAssert() {
         //Private to prevent instantiation
     }
 
-    public static void assertEquals(final String expectedJson, final String actualJson) {
+    public static void assertJsonEquals(final String expectedJson, final String actualJson) {
```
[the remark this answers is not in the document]
macenturalxl1: Agreed I've reverted this back to original method name

core/data/src/test/java/uk/gov/gchq/gaffer/data/element/function/FunctionTest.java
```diff
@@ -0,0 +1,113 @@
+/*
```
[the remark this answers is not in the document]
macenturalxl1: I've added documentation to this class with a @see annotation to reference the koryphe dependency and why (in regards to JUnit5)

library/flink-library/src/test/java/uk/gov/gchq/gaffer/flink/operation/FlinkTest.java
macenturalxl1: I think it originally failed once on the pipeline and intermittently on my personal machine (maybe it is slightly slower in some test runs) so the assertion runs too early and fails the test or breaks pipeline

I can change it back to 2000?

d47853: No it's fine just might be worth noting if it's unreliable.

Issue #2266, opened 2020-04-14
Create Command line utility to run an operation


d47853:
Create a command line utility which creates a graph from a store properties file and graph config, and runs an operation as a provided user. You could use json serialisation for the operation and user.

PR #2276, opened 2020-05-12
gh-2266 Command line Operation runner.

m29827:


operation-runner/pom.xml
```diff
@@ -0,0 +1,120 @@
...
+            <artifactId>sketches-library</artifactId>
+            <version>${project.parent.version}</version>
+            <classifier>deploy</classifier>
+        </dependency>
+        <dependency>
+            <groupId>uk.gov.gchq.gaffer</groupId>
+            <artifactId>time-library</artifactId>
+            <version>${project.parent.version}</version>
+            <classifier>deploy</classifier>
+        </dependency>
```
d47853, line 39: As this is designed to be a command line tool, it might be better to leave these out and the user can include them if they need to. Also when running with Accumulo the dependencies are included automatically as iterators. By doing this we can shrink the size of the actual Jar

operation-runner/pom.xml
```diff
@@ -0,0 +1,120 @@
...
+            <groupId>org.apache.hadoop</groupId>
+            <artifactId>hadoop-client</artifactId>
+            <version>${hadoop.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.hadoop</groupId>
+            <artifactId>hadoop-mapreduce-client-common</artifactId>
+            <version>${hadoop.version}</version>
+            <scope>compile</scope>
+        </dependency>
```
d47853, line 67: Including Hadoop as a dependency might cause issues with deploying onto different versions. It might be better to rely on the user using the Hadoop classpath on their cluster. Again with Accumulo, the hadoop client jars are included in the iterators. With Hbase, you can do a similar thing with `hbase classpath`. Removing these should reduce the size of the build artefact without sacrificing functionality, provided the correct classpath is provided by the user.

operation-runner/pom.xml
```diff
@@ -0,0 +1,120 @@
...
+            <artifactId>accumulo-store</artifactId>
+            <version>${project.parent.version}</version>
+            <classifier>iterators</classifier>
+        </dependency>
+        <dependency>
+            <groupId>uk.gov.gchq.gaffer</groupId>
+            <artifactId>hbase-store</artifactId>
+            <version>${project.parent.version}</version>
+            <classifier>deploy</classifier>
+        </dependency>
```
d47853, line 80: This should be store agnostic and so should not require store dependencies. These should be provided by the user at runtime.

core/store/src/main/java/uk/gov/gchq/gaffer/store/Context.java
```diff
@@ -34,6 +34,7 @@
  * as the user who executed the operation chain and a map of {@link Exporter}s.
  */
 public class Context {
+    public static final String COMMAND_LINE_ARGS_CONFIG_KEY = "config.command.line.args";
```
d47853, line 37: It might be worth extending this class in the operation-runner module rather than adding this code here.

PR #2281, opened 2020-06-02
Gh 1836 mock accumulo replacement

phosphorbronze50:
Changes for the replacement of MockAccumuloStore with new MiniAccumuloStore in tests.


store-implementation/accumulo-store/src/main/java/uk/gov/gchq/gaffer/accumulostore/MiniAccumuloStore.java
```diff
@@ -0,0 +1,274 @@
...
+
+    @Override
+    public void preInitialise(final String graphId, final Schema schema, final StoreProperties properties) throws StoreException {
+        setProperties(properties);
+        // This MiniAccumuloStore object may be being reused so we need to clean up some of the attributes
+        clearAllOperationalHandlers();
+        super.preInitialise(graphId, schema, getProperties());
+    }
+
+    private void clearAllOperationalHandlers() {
```
d47853, line 184: Would this work if operations were added using OperationDeclarations?

store-implementation/accumulo-store/src/main/java/uk/gov/gchq/gaffer/accumulostore/MiniAccumuloStore.java
```diff
@@ -0,0 +1,274 @@
...
+
+    public Connector getMiniConnector() {
+        return miniConnector;
+    }
+
+    @Override
+    public void preInitialise(final String graphId, final Schema schema, final StoreProperties properties) throws StoreException {
+        setProperties(properties);
+        // This MiniAccumuloStore object may be being reused so we need to clean up some of the attributes
+        clearAllOperationalHandlers();
```
d47853, line 180: I feel like if the store is being re-used then it shouldn't be using the PreInitialise() method. But perhaps my assumption is wrong.

store-implementation/accumulo-store/src/main/java/uk/gov/gchq/gaffer/accumulostore/MiniAccumuloStore.java
```diff
@@ -0,0 +1,274 @@
...
+            miniAccumuloCluster = new MiniAccumuloCluster(miniAccumuloConfig);
+            miniAccumuloCluster.start();
+        } catch (final IOException | InterruptedException e) {
+            throw new StoreException(e.getMessage(), e);
+        }
+
+        // Create the user specified in the properties (if not root)
+        // together with the specified password and give them all authorisations
+        try {
+            if (!getProperties().getUser().equalsIgnoreCase("root")) {
```
d47853, line 141: I don't think Gaffer needs to do this. Certainly it's not done as part of the traditional AccumuloStore implementation. In our [docs](https://gchq.github.io/gaffer-doc/stores/accumulo-store.html#trouble-shooting) we have covered this in the BAD_AUTHORIZATIONS error.

store-implementation/accumulo-store/src/main/java/uk/gov/gchq/gaffer/accumulostore/MiniAccumuloStore.java
```diff
@@ -0,0 +1,274 @@
...
+    private MiniAccumuloCluster miniAccumuloCluster = null;
+    private MiniAccumuloConfig miniAccumuloConfig = null;
+    private Connector miniConnector;
+
+    public StoreProperties setUpTestDB(final StoreProperties properties) throws StoreException {
+        setProperties(properties);
+
+        File targetDir = new File("target");
+        File baseDir;
+        if (targetDir.exists() && targetDir.isDirectory()) {
```
d47853, line 122: If this logic was moved to the pre-initialise method (in combination with the suggestion below), this method would not be needed and could be removed from the Store class also

store-implementation/accumulo-store/src/main/java/uk/gov/gchq/gaffer/accumulostore/MiniAccumuloStore.java
```diff
@@ -0,0 +1,274 @@
...
+    }
+
+    private void closeMiniAccumuloStore() {
+        if (null == miniAccumuloCluster) {
+            return;
+        }
+        try {
+            miniAccumuloCluster.stop();
+        } catch (final IOException | InterruptedException e) {
+            // Ignore any errors here;
```
d47853, line 266: It would be better if these errors were logged in case they hide any underlying bugs.

store-implementation/accumulo-store/src/main/java/uk/gov/gchq/gaffer/accumulostore/MiniAccumuloStore.java
```diff
@@ -0,0 +1,274 @@
...
+        // Create the new properties object to pass back, including connection items
+        AccumuloProperties accumuloProperties = (AccumuloProperties) properties.clone();
+        accumuloProperties.setInstance(miniAccumuloCluster.getInstanceName());
+        accumuloProperties.setZookeepers(miniAccumuloCluster.getZooKeepers());
+
+        return accumuloProperties;
+    }
+
+    public void tearDownTestDB() {
+        this.closeMiniAccumuloStore();
```
d47853, line 165: What would be the impact if the store was not closed down? I only ask because if we could remove this method, we wouldn't have to pollute the Store class with code only designed to be used with tests.

store-implementation/accumulo-store/src/main/java/uk/gov/gchq/gaffer/accumulostore/MiniAccumuloStore.java
```diff
@@ -0,0 +1,274 @@
...
+    private static final String BASE_DIRECTORY = "miniAccumuloStoreTest-";
+    private static final String ROOTPW = "rootPW";
+    private MiniAccumuloCluster miniAccumuloCluster = null;
+    private MiniAccumuloConfig miniAccumuloConfig = null;
+    private Connector miniConnector;
+
+    public StoreProperties setUpTestDB(final StoreProperties properties) throws StoreException {
+        setProperties(properties);
+
+        File targetDir = new File("target");
```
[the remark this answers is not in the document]
phosphorbronze50: I can put it into the store properties via AccumuloProperties if you like.
I would avoid putting it into a subclass (e.g. MiniAccumuloProperties) as that will cause issues where tests are just supplied with properties and set up the (SingleUse)MiniAccumuloStore when it is required. This is because such tests currently lend themselves to being used with alternative properties including a 'real' and persistent 'AccumuloStore' rather than the Mini one for local test purposes.
I am wondering what sort of problem specifying the sub-directory name prefix as a property resolves, as the directory gets differentiated where more than one is required through the use of a UUID.randomUUID() suffix for each instance.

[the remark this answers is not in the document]
d47853: In combination with the other suggestions, basically the benefit would be that you don't need this method which currently pollutes the store class.

store-implementation/accumulo-store/src/main/java/uk/gov/gchq/gaffer/accumulostore/MiniAccumuloStore.java
phosphorbronze50: The store's re-use it at the heart of the changes being implemented.
'MockAccumuloStore's are/were always created afresh every time for every test in a test suite.
To do this with 'MiniAccumuloStore's would be a massive overhead so it was requested that tests will set up a mini Accumulo instance at the start of a test suite, use it for all the tests in the suite, and then tear it down at the end of the suite.
It was found that when some tests were migrated to this approach some tests were failing as pre-conditions set for one test would, naturally enough, persist in the subsequent tests. Given that JUnit will not guarantee any specific sequence for tests in a suite it is necessary to ensure the Accumulo instance is returned to its 'default' condition before the  instance is used for the next test.

store-implementation/accumulo-store/src/main/java/uk/gov/gchq/gaffer/accumulostore/MiniAccumuloStore.java
phosphorbronze50: That situation has tests and they are still passed when the MiniAccumuloStore is used for them.
This was added because tests were failing for operation handlers. It creates a consistent environment for each test in a suite.

store-implementation/accumulo-store/src/main/java/uk/gov/gchq/gaffer/accumulostore/MiniAccumuloStore.java
phosphorbronze50: This was added because with certain sets of properties we were getting BAD_AUTHORIZATIONS exceptions when they weren't expected.
MockAccumulo has no effective understanding or handling of users.
This code has been added because MiniAccumulo does more properly mimic Accumulo user behaviour.
Root is excluded as it is unnecessary as root automatically always has all permissions and when it is not excluded odd behaviour was encountered on setting the authorisations.

store-implementation/accumulo-store/src/main/java/uk/gov/gchq/gaffer/accumulostore/MiniAccumuloStore.java
phosphorbronze50: MockAccumulo instances are memory based and are instantiated within the JVM. When they go out of scope they cease to exist. They need no lifecycle handling.
Full Accumulo installations are fully persistent and can be connected to when needed and if the connection is dropped then they can happily continue to exist. They need no lifecycle handling.
MiniAccumulo instances have to be created with details of a directory where the data can reside, and after creation they have to be explicitly started. The filesystem / environment should be (as far as possible) returned to the original state as before the tests the Mini Accumulo instance was created for. This requires that the MiniAccumulo Instance is gracefully closed and the directory structure deleted.
The MiniAccumuloStore thus needs to be explicitly created and started (setUpTestDB).
It also needs to create a set of AccumuloProperties based on the one supplied to create the Accumulo instance, but suitably updated to include the connection details for the zookeeper of the MiniAccumuloStore. It is this set of properties that is returned by setUpTestDB() to be used in all the subsequent tests in the suite.
Additionally we need to create the MiniAccumulo instance in such a way that we can maintain a reference to it so we can call the corresponding tearDownTestDB() to be able to target the correct instance to close it gracefully and then remove the directory structure that the Mini instance used.
The setUpTestDB() is called from the @BeforeAll / @BeforeClass in the test suite in which it is required, and the tearDownTestDB() is called in the @AfterAll / @AfterClass of the same test suite. This ensures that, assuming JUnit is functioning correctly, the MiniAccumulo instance is tidily handled.

If it was moved to the pre-initialise method there would be no handling of the closure and disposal of the instance, and the tests would have no means of accessing the zookeeper details to access the MiniAccumulo instance.

store-implementation/accumulo-store/src/main/java/uk/gov/gchq/gaffer/accumulostore/MiniAccumuloStore.java
phosphorbronze50: OK.
Where and at what level(s) would you like them to be logged or would simply not catching them allow them to trigger any further required actions in the builds and pipeline?

store-implementation/accumulo-store/src/main/java/uk/gov/gchq/gaffer/accumulostore/MiniAccumuloStore.java
phosphorbronze50: If there was no handling for disposal of the Mini Accumulo instance then the processes may continue to run, and the filesystem space used by the instance will not be recovered, gradually building up over time in the target directory.
Adding the 2 concrete methods to the store class was not done lightly, was asked about early on and was accepted as a necessary evil.
Adding them to the store class means that the mechanism for a store that requires creation and destruction of the store for test purposes can be used for all other store types. Indeed, it is looking like federated-store may need to make use of the mechanism.
They were made as compact and efficient as possible.
Making them concrete methods that are then overridden where needed rather than abstract methods means the other Store subclasses do not need to be changed to have what will simply be a copy of what has been put in the Store class.
As regards polluting the main code with functionality only used in tests I agree it is not ideal, but whether it ended up in Store, AccumuloStore or MiniAccumuloStore (it being derived from the deprecated MockAccumuloStore) makes no difference from that point of view - all are in the main code base.

library/spark/spark-accumulo-library/src/test/java/uk/gov/gchq/gaffer/sparkaccumulo/operation/handler/javardd/GetJavaRDDOfElementsHandlerTest.java
```diff
@@ -42,13 +46,42 @@
...
+                    .newInstance();
+        } catch (final InstantiationException | IllegalAccessException | ClassNotFoundException e) {
+            throw new IllegalArgumentException("Could not create store of type: " + storeClass, e);
+        }
+        // Set up the data store and set the properties to suit.
+        storeProperties = (StoreProperties) store.setUpTestDB(suppliedProperties);
+    }
+
+    @AfterAll
+    public static void tearDownAfterClass() throws Exception {
```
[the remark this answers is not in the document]
phosphorbronze50: I will look into it.
I had considered a util class but much of the context sensitive code then no longer worked.
I will look at an abstract superclass.

store-implementation/accumulo-store/src/main/java/uk/gov/gchq/gaffer/accumulostore/MiniAccumuloStore.java
d47853: We tend to use slf4j logger factories: 
```java
private static final Logger LOGGER = LoggerFactory.getLogger(MiniAccumuloStore.class);
// ...
LOGGER.error(myerror)
```

For something like this, you could retry it a couple of times and if it continues to fail, throw the exception.

store-implementation/accumulo-store/src/main/java/uk/gov/gchq/gaffer/accumulostore/MiniAccumuloStore.java
d47853: Thanks for your responses. To avoid copy and pasting the same thing to each of the comments, as they address similar issues, I'll try and articulate a reply here. 

I feel like the problem lies in the fact that MiniAccumulo is trying to imitate MockAccumulo exactly (by creating the store and then connecting to it). As you've highlighted it would be too inefficient to do this for every single test.

Perhaps it would be better if it was closer aligned to the AccumuloStore. By that, I mean that it merely connects to the store using the Accumulo properties. The setting up and tearing down of the actual Mini Accumulo instance could be done by the tests. This feels like a better solution as you could then create a new store object for each test as you'd only be creating the connection objects, rather than writing to disk.

phosphorbronze50: I think you may be assuming that because all the Mini Accumulo Cluster set up and take down code is co-located in the MiniAccumuloStore class together with other Store overridden methods etc. this means that it all the code runs at the same time; it does not.

The issue:
AccumuloStore can simply be instantiated and, provided it has valid zookeeper details in its supplied properties it can work, connecting to the remote Accumulo Server which will already be running and will continue to run after we have finished with it.
MockAccumuloStore can simply be instantiated and will spin up an in-memory Accumulo MockInstance, use it, and can then forget about it. When the references go out of scope the garbage collector will clear it away.
MiniAccumuloStore requires a more complex lifecycle because Mini Accumulo Cluster requires local resources (filestore) to be able to run, and needs explicit starting and stopping.

The solution used:
To run a JUnit test class/suite using MiniAccumuloStore as the Store we need to set up the Mini Accumulo Cluster that the tests will connect to.
This is achieved at the start of the test suite in the @BeforeClass / @BeforeAll annotated method by instantiating a MiniAccumuloStore and calling the MiniAccumuloStore.setUpTestDB() class supplying the StoreProperties already provided in the test. This method:
- creates a Mini Accumulo Cluster with its files under the project's target directory
- starts the cluster
- interrogates the cluster to discover its zookeeper socket details
- clones the originally supplied StoreProperties and updates the zookeeper details in it
- returns the cloned and updated StoreProperties
The returned StoreProperties and the MiniAccumuloStore objects are kept in private attributes in this JUnit instance.
In the subsequent test methods of the JUnit suite the returned StoreProperties are the ones used. All the Graph builds etc will use the normal AccumuloStore mechanisms to access the Mini Accumulo Cluster, instantiating MiniAccumuloStore objects as required, each of which will only actually be a holder of the StoreProperties / AccumuloProperties containing the Mini Accumulo Cluster's zookeeper address details.
When all the test methods have been run then JUnit will run the @AfterClass / @AfterAll annotated method. This will use the MiniAccumuloStore instance kept in the JUnit's private attributes to call its tearDownTestDB() method. This particular MiniAccumuloStore instance contains the MiniAccumuloCluster object and the directory name used for the creation of the Mini Accumulo Cluster so it can use them to stop the cluster and then delete the filestore directory it was using.

In the light of the above outline I hope you can see that the MiniAccumuloStore is operating in a manner far closer to the AccumuloStore than the MockAccumuloStore. In the test methods any Graph build "merely connects to the store using the Accumulo properties" as you would like.

Is it that some of your concerns revolve around the inclusion of the 2 methods (setUpTestDB and tearDownTestDB) in the main Store superclass? I will explain the design thinking behind these.
There are a number of tests that previously used but have no explicit dependency on MockAccumuloStore e.g. in the uk.gov.gchq.gaffer.sparkaccumulo test classes. They only have a set of StoreProperties provided which specified the SingleUseMockAccumuloStore as the store class to be used.
This works OK for AccumuloStore and MockAccumuloStore but as we have seen above MiniAccumuloStore requires the Mini Accumulo Cluster to be set up and taken down again.
The Store setup code added to these test suites' @BeforeClass/@BeforeAll was deliberately written with only uk.gov.gchq.gaffer.store superclass dependencies used.
The test setup reads the StoreProperties, instantiates the declared Store class, and calls the setUpTestDB() method with the supplied StoreParameters against it.
In all implementations other than MiniAccumuloStore this simply uses the Store superclass implementation and returns the properties it was called with. The override of the setUpTestDB() in MiniAccumuloStore sets up the Mini Accumulo Cluster instance under the target directory, starts the instance and returns a new set of properties with the updated zookeeper details.
The returned properties, in either case, are then used in the subsequent test methods in the normal way for Graph building etc.
Once all the test methods have been run the tearDownTestDB() method is called against the store. In the 'default' Store superclass implementation this simply returns. For the MiniAccumuloStore implementation this stops the Mini Accumulo Cluster and deletes the directory the instance was using.
As a result of this approach these tests can now simply have their properties changed to support different stores (AccumuloStore, MockAccumuloStore, MiniAccumuloStore or any of the other store implementations) exactly as they could before. The methods are there for use by other store implementations in the future should the need arise. Federated-store tests may have need of this facility.
In hindsight I do admit that the use of the word 'Test' in the new stub Store method names was ill-advised. Maybe setUpStore and tearDownStore would be more appropriate names?

library/spark/spark-accumulo-library/src/test/java/uk/gov/gchq/gaffer/sparkaccumulo/operation/handler/javardd/GetJavaRDDOfElementsHandlerTest.java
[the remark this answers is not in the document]
phosphorbronze50: Moved to an abstract superclass as requested and classes changed to use it where appropriate.

store-implementation/accumulo-store/src/main/java/uk/gov/gchq/gaffer/accumulostore/MiniAccumuloStore.java
phosphorbronze50: Changed as requested.

store-implementation/accumulo-store/src/main/java/uk/gov/gchq/gaffer/accumulostore/MiniAccumuloStore.java
d47853: In that case, it's probably better that we use the standard accumulo store for tests that really need to use an accumulo store (Such as the AccumuloStoreITs). For tests that just need an in-memory store, they can just use the MapStore.

Here's an example of how I'd do it: https://github.com/gchq/Gaffer/compare/gh-1836-demo?expand=1

I've only made changes to the AccumuloStoreITs in the demo and the rest of the accumulo store test's need refactoring. While doing it, I found that I didn't really need another Store class as the AccumuloStore works well with the MiniAccumuloCluster.

PR #2283, opened 2020-06-09
Gh 2270 JUnit5 upgrade (modules: example)

macenturalxl1:
To merge with `develop` after pull request: https://github.com/gchq/Gaffer/pull/2277 is merged


macenturalxl1:
Hi - the code coverage for JUnit4 tests aren't being picked up for modules in library and store-implementation (outside of the modules I've upgraded in this pull request) - I'll think of a solution to unblock the coverage check - might need to combine this with upgraded tests in those modules so the coverage is picked up

Issue #1836, opened 2018-07-23
Remove Instances of Mock Accumulo 


james010101101:
In Accumulo 2.0.0 the mock accumulo store will be removed, in order to prepare for a later 2.0.0 migration we should remove all instances of the Mock Accumulo Store and replace them with Mini Accumulo Clusters

james010101101:
The overheads of running Mini Accumulo raise the time taken for tests significantly. 
Integration tests in particular are a problem, refactoring the IT's to (where possible) use a single store/graph would be best.

ctas582:
This hopefully shouldn't be that difficult. We are already running the integration tests against a single store when we run them on AWS. All we change is the store class, see: https://github.com/gchq/gaffer-tools/blob/master/deployment/aws/tests/integration-tests/emr-step-scripts/run-gaffer-integration-tests.sh#L118

phosphorbronze50:
I have started to look into this issue.
Creating a MiniAccumuloStore and using this as the gaffer.store.class property has integrated moderately well into the testing regime but needs to be further enhanced to avoid resource issues of the Mini Accumulo Clusters not being gracefully closed after tests are complete. This will be tied into the refactoring of tests for performance issues already noted above.

phosphorbronze50:
I have moved ahead with getting the MiniAccumuloCluster to start in new JUnit @BeforeClass / @BeforeAll methods and stop and tidy up in a new @AfterClass / @AfterAll methods.
The tests that simply specify the properties and supply them to the graph generation have been amended to read the store.properties:gaffer.store.class and instantiate the store if required.
The code that I am adding to the tests is non-Accumulo specific, using Store and StoreProperties classes, generating the Accumulo specifics.
The new Accumulo functionality has been added to the MiniAccumuloStore class itself.
In the MiniAccumuloStore many of the methods which get overridden in MockAccumuloStore now use the methods in the AccumuloStore superclass.
 
I would like to add a couple of additional methods to the Store Class :

public StoreProperties setUpTestDB(final StoreProperties properties) throws StoreException {
    return properties;
}

public void tearDownTestDB() {
    return;
}

These are concrete methods. 
They would be overridden in MiniAccumuloStore to call the necessary setup/start and stop/delete of the MiniAccumuloCluster. These overrides are then able to be called from the test classes' @BeforeClass / @BeforeAll and @AfterClass / @AfterAll methods.
[...]

PR #2286, opened 2020-06-22
Response to the gh-1836-demo example.

phosphorbronze50:
I have looked at the demo and taken that and other comments and reorganised the approach to the Mock to Mini Accumulo changes.
This pull request branch has only a few changes outlining how I will go about progressing gh-1836.

It was noted for the gh-1836-demo that as the Mini Accumulo Cluster should behave exactly as full Accumulo we shouldn't need any MiniAccumuloStore sub classes. This does indeed work but given the normal mechanism used by Gaffer to identify a store type is the class name specified in the properties files. This means the MiniAccumuloStore and SingleUseMiniAccumuloStore classes are needed for identification purposes, unless I have misunderstood something in this area.
The new MiniAccumuloStore class extends AccumuloStore with no changes to functionality.
The new SingleUseMiniAccumuloStore extends MiniAccumuloStore in exactly the same manner as SingleUseAccumuloStore extends AccumuloStore.
They are extended in this way rather than SingleUseMiniAccumuloStore extending SingleUseAccumuloStore to replicate the AccumuloStore object hierarchy.

Control of the decision whether to start a MiniAccumuloCluster, and the starting and stopping it, is handled by the new AccumuloTestClusterManager class. This takes an AccumuloProperties object in its constructor and reads it to find the store class. If it is a Mini Accumulo one it sets up and starts a MiniAccumuloCluster and updates the properties with the zookeeper details. If not it just returns.
Creation of the object is called at the beginning of a test suite, often through the @BeforeClass annotated method.
At the end of a test suite (@AfterClass) the close method is called to close and tidy away the Mini Accumulo Cluster if one has been started. The test suite will have kept the reference to the AccumuloTestClusterManager object for this purpose.

There are three tests converted from using Mock to Mini:
AccumuloStoreITs from the demo is suitably modified to use the new AccumuloTestClusterManager. This is slightly odd looking as for this test the constructors receive the properties. This prevents use of the @BeforeClass as that has to be a static method so cannot use instance attributes.
GetDataFrameOfElementsHandlerTest is a spark-accumulo test that was constructed to set up its store only from properties. This was among the tests that previously had a lot of duplicated code. They will now have an abstract superclass, AbstractPropertiesDrivenTest, to handle most of the repeating setup.
AccumuloStoreTest is an example of the most common types of unit tests. This had explicit Mock handling as part of the tests, in some cases with SingleUseMockAccumuloStore being instantiated by individual tests. This has been converted to reuse of a pair of mini accumulo clusters for all the tests.

In all the test examples above I have avoided declaration of the MiniAccumuloStore and SingleUseMiniAccumuloStore classes as far as possible, instead having any necessary Store declarations being for AccumuloStore.

Is this closer to what was expected?


d47853:
Thanks @phosphorbronze50. That sounds a lot more in-tune with my thinking at least. Have you made these changes in your other pull request? And if so is it ready for a second review?

phosphorbronze50:
Because of the nature of the changes and the need to revisit everything in the last pull request branch I am going to abandon the previous one and move ahead with this one, bringing the changes from the other through to this one as required.

d47853:
Seems sensible. Do you mind if I close both pull requests so you can re-open this one when you're ready for it to be reviewed?

phosphorbronze50:
Closing both pull requests is fine with me.

PR #2299, opened 2020-08-04
Gh 2290 view flags all edges all entities not cloned

r32575:
The tests are now working and have been converted from junit 4 to 5 apart from the integration tests as changing them was more complex and could do with another task. Some of the unit tests have been disabled after failing with a lot of effort effort put it to correcting the problems. Another task will have to be raised to look into these further.

The original task to correct the View has also been fixed. It was when running maven after making the changes that problem above arose.


core/data/src/main/java/uk/gov/gchq/gaffer/data/elementdefinition/view/View.java
```diff
@@ -533,6 +542,9 @@ public CHILD_CLASS merge(final View view) {
                 if (null != view.config) {
                     getThisView().config.putAll(view.config);
                 }
+
+                getThisView().setAllEntities(view.allEntities);
```
d47853, line 546: Maybe it would be better to decouple the clone from the fromJson / toJson methods and have a separate tests for this?

core/data/src/main/java/uk/gov/gchq/gaffer/data/elementdefinition/view/View.java
```diff
@@ -433,6 +440,7 @@ public CHILD_CLASS globalElements(final GlobalViewElementDefinition... globalEle
             return self();
         }
 
+        @JsonSetter("globalEntities")
```
d47853, line 443: Are these annotations needed? Usually jackson uses the public setter methods. I may be wrong.

core/data/src/main/java/uk/gov/gchq/gaffer/data/elementdefinition/view/View.java
r32575: After discussion decided to leave the clone method as is

core/data/src/main/java/uk/gov/gchq/gaffer/data/elementdefinition/view/View.java
r32575: Removed all of the @JsonSetter methods as it was not required.

Issue #2295, opened 2020-07-22
Harmonise access mechanisms for Gaffer resources


d47853:
For resources such as NamedOperations and Graphs in the Federated Store, it would be really neat to harmonise and improve the way that access control is implemented. It would need to be flexible so that it can be applied to different resources. Therefore it would make sense to have a predicate that is applied to a user accessing a given resource which can be configured by more advanced users. 

For this to be accepted in Gaffer 1.x, it would need to be backwards compatible with what's already there.

The resources I'd expect this to work for are:
* NamedOperations
* NamedViews
* Graphs (Federated Store)

Doing this would make what's access control more flexible and easier to implement for any new resources.

d47853:
Merged into develop

PR #2301, opened 2020-08-06
gh-2295 Harmonise resource access controls for Named Operations Named Views and Federated Store Graphs.

m29827:


core/access/src/main/java/uk/gov/gchq/gaffer/access/predicate/AccessPredicate.java
```diff
@@ -0,0 +1,30 @@
...
+import com.fasterxml.jackson.annotation.JsonTypeInfo;
+
+import uk.gov.gchq.gaffer.user.User;
+
+import java.io.Serializable;
+import java.util.List;
+import java.util.function.BiPredicate;
+
+@JsonTypeInfo(use = JsonTypeInfo.Id.CLASS, property = "class")
+public abstract class AccessPredicate implements BiPredicate<User, String>, Serializable {
```
d47853, line 28: Instead of making this a BiPredicate, could we make it a Predicate<User> which would allow use to make use of the existing Koryphe Functions and Predicates?

The second adminUser argument could be a class field of the predicate as it's unlikely to change.

d47853 (changes requested):
Just thinking of a way we could make this more flexible

d47853 (changes requested):

d47853 (approved):

PR #2317, opened 2020-09-30
Update version.properties

d47853:


d47853:
PR no longer required

PR #2329, opened 2020-10-23
Gh 2302 create spring boot app

d47853:
Common code shared between core-rest and the new spring-rest project has been moved into a new common-rest module.


d47853:
closed while failing tests are investigated

Issue #1284, opened 2017-09-13
Using the MiniAccumuloStore in Travis CI causes the jobs to be killed


p013570:
We have had to disable 2 spark-accumulo-library tests from running in Travis CI due to the jobs always being killed - probably due to using too much memory. They rely on the MiniAccumuloStore, so it is probably that, which is causing the problems.

We need to get those 2 tests running in Travis CI again. They can be enabled again by removing this block from the parent pom:

```xml
<exclude.tests>**/sparkaccumulo/integration/**/*IT.java</exclude.tests>
```

PR #2330, opened 2020-10-23
gh-1284 Re-enabled spark-accumulo tests

d47853:


d47853:
#1284 can also be resolved by merging #2332

d47853:
No longer required as of #2332

PR #2336, opened 2020-10-27
Gh 2302 create spring boot app

d47853:


t11947 (approved):
Looks good. The only thing I could think to suggest is that one or two of the classes seem a little heavy and it could be worth making them a little more compositional. AbstractExamplesFactory for example. I can also see never getting the benefit of a more compositional implementation, which is why I'm happy to approve if you feel like this is the case also.

d47853:
We can always refactor if we need to break them apart in future. For now I think it's as simple as I can get it while re-using code shared between the two modules

PR #2337, opened 2020-10-27
link-PR-to-issue

p3430233:


d47853 (approved):

d47853:
To avoid confusion, it may be better to change the name of this workflow to something like link-issue.yaml

PR #2239, opened 2020-01-28
Gh 2211 functions

p013570:

Issue #2325, opened 2020-10-20
Improve Kafka flink tests


d47853:
The Flink integration tests use Thread.sleep() lines to ensure the data is loaded. This is both unreliable and inefficient. Refactor to make them reliably pass without using Thread.sleep().

d47853:
Merged into develop

PR #2343, opened 2020-11-17
bugfix for gh-2239:  Merging views with global elements

### LINKED ISSUE #2293: Global Elements/Edges/Entities definitions only applied when there is an explicit edge/element definition
Combinations of globalEdges/globalEntities/globalElements and properties/groupBy/predicates do not work as intended when no explicit element definition is included.

For example:
```
{
    "class": "uk.gov.gchq.gaffer.operation.impl.get.GetElements",
    "options": {
        "gaffer.federatedstore.operation.graphIds": "Combined_Graph"
    },
    "view": {
        "globalEdges": [
            {
                "groupBy": []
            }
        ]
    },
    "input": [...]
}
```
Returns all edges and entities without aggregation (adding allEdges/allEntities params makes no difference).

However, inheritance from global definitions does appear to work. If an explicit edge or entity is defined in the view then it will inherit from the global.

For example:
```
{
    "class": "uk.gov.gchq.gaffer.operation.impl.get.GetElements",
    "options": {
        "gaffer.federatedstore.operation.graphIds": "Combined_Graph"
    },
    "view": {
        "edges": {
            "<example_edge>": {
                "properties": [
                    "count"
                ]
            }
        },
        "globalEdges": [
            {
                "groupBy": []
            }
        ]
    },
    "input": [...]
}
```
Returns <example_edge> edges with only the "count" property (as expected from the explicit edge definition) but those edges are aggregated in line with the globalEdge groubBy.

Tested on a Federated Graph setup.

Could be linked to https://github.com/gchq/Gaffer/issues/2290

sw96411:
Resolves #2293 

Only expand Global* elements in a View if the schema contains an entry with that element type in.  This supports stores such as the FederatedStore where the first Store a query gets run against is in some way a proxy of some other store(s) and doesn't have the schema information to do the expansion.

Also, update the logic in updateOperationChain such that it no longer expands  a View with no entities and edges but, e.g., a globalEntities element to include all of the edges.

When reviewing, please note that I've chosen to change updateOperationChain to default scope for testing purposes and because the code didn't feel like it was perfectly encapsulated by the logic of executing a query.  Alternatively, I could change the tests to run against the public execute method (which felt a bit broad) or use reflection to test the method as a private (but I'm not sure how you feel about that 'round these parts).  Lastly, where there's identical logic for Edges and Entities, I've only included a unit for the edges part, but let me know if you want a repeated unit test for the Entities logic

### DIFF
--- a/core/graph/src/main/java/uk/gov/gchq/gaffer/graph/Graph.java
+++ b/core/graph/src/main/java/uk/gov/gchq/gaffer/graph/Graph.java
@@ -72,19 +72,20 @@ import static org.apache.commons.collections.CollectionUtils.isEmpty;
 /**
  * <p>
  * The Graph separates the user from the {@link Store}. It holds an instance of
- * the {@link Store} and acts as a proxy for the store, delegating
- * {@link Operation}s to the store.
+ * the {@link Store} and
+ * acts as a proxy for the store, delegating {@link Operation}s to the store.
  * </p>
  * <p>
  * The Graph provides users with a single point of entry for executing
- * operations on a store. This allows the underlying store to be swapped and the
- * same operations can still be applied.
+ * operations on a store.
+ * This allows the underlying store to be swapped and the same operations can
+ * still be applied.
  * </p>
  * <p>
  * Graphs also provides a view of the data with a instance of {@link View}. The
- * view filters out unwanted information and can transform
- * {@link uk.gov.gchq.gaffer.data.element.Properties} into transient properties
- * such as averages.
+ * view filters out unwanted information
+ * and can transform {@link uk.gov.gchq.gaffer.data.element.Properties} into
+ * transient properties such as averages.
  * </p>
  * <p>
  * When executing operations on a graph, an operation view would override the
@@ -104,11 +105,12 @@ public final class Graph {
     private GraphConfig config;
 
     /**
-     * Constructs a {@code Graph} with the given
-     * {@link uk.gov.gchq.gaffer.store.Store} and
+     * Constructs a {@code Graph} with the given {@link uk.gov.gchq.gaffer.store.Store}
+     * and
      * {@link uk.gov.gchq.gaffer.data.elementdefinition.view.View}.
      *
-     * @param config a {@link GraphConfig} used to store the configuration for a
+     * @param config a {@link GraphConfig} used to store the configuration for
+     *               a
      *               Graph.
      * @param store  a {@link Store} used to store the elements and handle
      *               operations.
@@ -119,8 +121,8 @@ public final class Graph {
     }
 
     /**
-     * Performs the given operation on the store. If the operation does not have a
-     * view then the graph view is used.
+     * Performs the given operation on the store.
+     * If the operation does not have a view then the graph view is used.
      *
      * @param operation the operation to be executed.
      * @param user      the user executing the operation.
@@ -131,9 +133,9 @@ public final class Graph {
     }
 
     /**
-     * Performs the given operation on the store. If the operation does not have a
-     * view then the graph view is used. The context will be cloned and a new jobId
-     * will be created.
+     * Performs the given operation on the store.
+     * If the operation does not have a view then the graph view is used.
+     * The context will be cloned and a new jobId will be created.
      *
      * @param operation the operation to be executed.
      * @param context   the user context for the execution of the operation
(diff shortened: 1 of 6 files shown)


sw96411:
It looks like a CI build failed because travis couldn't download an apache commons library from maven - I don't think that's anything to do with this change; is it possible for someone re-run the checks?

core/graph/src/main/java/uk/gov/gchq/gaffer/graph/Graph.java
```diff
@@ -350,22 +354,32 @@ public JobDetail executeJob(final Job job, final Context context) throws Operati
         return new GraphResult<>(result, clonedContext);
     }
 
-    private void updateOperationChainView(final Operations<?> operations) {
+    void updateOperationChainView(final Operations<?> operations) {
```
d47853, line 357: Is this change needed? If it's just for the test then it would be better to indirectly call it from one of the public methods

core/graph/src/test/java/uk/gov/gchq/gaffer/graph/GraphTest.java
```diff
@@ -166,73 +167,41 @@ public void shouldConstructGraphFromSchemaModules() {
...
-                        .clazz(String.class)
-                        .build())
-                .edge(TestGroups.EDGE, new SchemaEdgeDefinition.Builder()
-                        .property(TestPropertyNames.PROP_1, TestTypes.PROP_STRING)
-                        .aggregate(false)
-                        .source("vertex")
-                        .destination("vertex")
-                        .directed(DIRECTED_EITHER)
-                        .build())
+                .type(TestTypes.PROP_STRING, new TypeDefinition.Builder().clazz(String.class).build())
```
d47853, line 170: There are a lot of changes to this file, most are whitespace from what I can tell. Please could you revert some of the whitespace changes so it's easier to review?

core/graph/src/test/java/uk/gov/gchq/gaffer/graph/GraphTest.java
```diff
@@ -2418,20 +1970,94 @@ public void shouldNotAddDefaultSecurityHooksIfAlreadyAdded() {
...
+
+        // Then
+        assertEquals(Arrays.asList(NamedOperationResolver.class, NamedViewResolver.class, FunctionAuthoriser.class),
+                graph.getGraphHooks());
+        assertEquals(Identity.class,
+                ((FunctionAuthoriser) graph.getConfig().getHooks().get(2)).getUnauthorisedFunctions().get(0));
+    }
+
+    @Test
+    public void shouldExpandGlobalEdges() {
```
d47853, line 1987: Might be worth adding some integration tests into the federated store to check the behaviour

d47853 (changes requested):

core/graph/src/main/java/uk/gov/gchq/gaffer/graph/Graph.java
sw96411: Sure thing, this is what I was trying to get at in the PR description - I'll refactor the test to work on the execute method and revert the visibility of the method.

core/graph/src/test/java/uk/gov/gchq/gaffer/graph/GraphTest.java
sw96411: Sorry about that - was trying to setup my save actions to work with the Gaffer checkstyle and it looks like it's started merging lines (this also seems to have caused the measured test coverage % to drop) - I'll sort that out

sw96411:
@d47853 :  I think I've addressed the points you discussed.  The automated PR checks show the unit tests are failing, but I note they're failing the same way with other open PRs right now, and in a way that doesn't seem to be to do with this change.  Is there something up with the CI build at the moment?  If so, happy to wait for that to get resolved and then re-run the checks here.

d47853:
Yeah it looked like the Flink tests failed which we're looking at as part of #2325. I've re-run the CI

sw96411:
Looks like we've got a little further, and I've added some tests to improve coverage, but it looks like the checks are failing again - I note for the same reasons as #2344 .  Do you think this is something I need to resolve in this PR, @d47853 , or is this something up with the build?  I've run the mvn build executed by the REST CI build locally and it seems fine, and the coverage of my change is higher than the project overall, and I've tried to verify locally with Emma, although I appreciate different coverage tools may use different metrics

d47853:
No I think that's something that should be looked at as part of a separate issue. I've re-run the tests as the failure seems to be intermittent

d47853:
As for coverage, we won't get a full picture until all the tests pass. You can imagine that some modules have slightly different coverages so when one or more reports aren't uploaded, we don't get a full picture

d47853 (approved):

d47853:
That failure now appears to happen pretty consistently. I've raised a new issue for it: #2345. Once a fix for it is added, we can re-run the tests and merge if successful.

Issue #2335, opened 2020-10-26
Automatically update Koryphe version on new koryphe release


d47853:
Add an action which updates the koryphe version upon a new Koryphe release. It can use the ./cd/updateKorypheVersion.sh script if necessary.

This will save developer time.

d47853:
Merged into develop

PR #2344, opened 2020-11-19
gh-2335 Adding github workflow to automate koryphe version update.

m29827:


m29827:
@d47853 This change is still using the updateKorypheVersion.sh script to perform the update. Happy to migrate the logic into the workflow but it might be useful to keep the script intact so it can be run outside of github.

.github/workflows/update-koryphe-version.yaml
```diff
@@ -0,0 +1,33 @@
...
+        run: |
+          git config user.name Gaffer
+          git config user.email github-actions@github.com
+
+      - name: Set up JDK 1.8
+        uses: actions/setup-java@v1
+        with:
+          java-version: 1.8
+
+      - name: Update Korpyhe Version
```
d47853, line 32: Could make the pull request afterwards?

d47853 (commented):
Looks good. Although it might be worth automatically submitting a pull request if the version is upgraded.

.github/workflows/update-koryphe-version.yaml
m29827: Now creating a PR.

.github/workflows/update-koryphe-version.yaml
```diff
@@ -0,0 +1,42 @@
...
+        run: ./cd/updateKorypheVersion.sh ${{ github.event.inputs.version }}
+
+      - name: Submit PR
+        if: ${{ success() }}
+        uses: repo-sync/pull-request@v2
+        with:
+          source_branch: ${{ env.VERSION_UPDATE_BRANCH }}
+          destination_branch: develop
+          github_token: ${{ secrets.ADMIN_GITHUB_TOKEN }}
```
d47853, line 42: Don't think admin token would be required for this. You could use the default one:
```suggestion
          github_token: ${{ secrets.GITHUB_TOKEN }}
```

d47853 (changes requested):

d47853 (approved):

PR #2347, opened 2020-11-30
Gh 2325 improve flink tests using end of elements indicator

m29827:
Sharing this approach for stabilising the flink unit-tests, not ideal as it relies on adding code to make the tests predictable.


m29827:
Implemented using alternative approach

PR #2349, opened 2020-12-01
gh-2325 Improve flink integration tests

m29827:


library/flink-library/src/main/java/uk/gov/gchq/gaffer/flink/operation/handler/AddElementsFromFileHandler.java
```diff
@@ -40,6 +41,18 @@
  * </p>
  */
 public class AddElementsFromFileHandler implements OperationHandler<AddElementsFromFile> {
+    private static final RichOutputFormat<Element> CREATE_GAFFER_OUTPUT_ON_DO_OPERATION = null;
+
+    private final RichOutputFormat<Element> outputFormat;
+
+    public AddElementsFromFileHandler() {
+        this(CREATE_GAFFER_OUTPUT_ON_DO_OPERATION);
```
d47853, line 49: Here and in other places, I'd remove the static nulls and just put null. I think the code is pretty clear without the need for static nulls

d47853 (changes requested):
Happy with it overall. I'd just suggest removing the static nulls

d47853 (approved):

PR #2350, opened 2020-12-01
Gh 2008 remove view validation

### LINKED ISSUE #2008: Remove view Edge Entity Validation for operations
Following the bug raised and fixed in gh-1992.
Should the views been validated at all?
Can the view validation be removed completely and opChains simply return null if the view specified does not exist?

GCHQDev404:
# Related Issue

- Resolve #2008


d47853 (changes requested):
I'm happy with the proposed changes from a code point of view. Would it be okay adding a test confirming the new behaviour (permitting view's that don't match schemas and propagating exceptions from delegate stores)

core/store/src/main/java/uk/gov/gchq/gaffer/store/SchemaOperationChainValidator.java
```diff
@@ -28,12 +27,7 @@
 
     Schema schema;
 
-    public SchemaOperationChainValidator(final ViewValidator viewValidator) {
```
d47853, line 31: For the sake of backwards compatibility, can you deprecate the old constructors rather than removing them?

core/store/src/main/java/uk/gov/gchq/gaffer/store/schema/ViewValidator.java
```diff
@@ -1,305 +0,0 @@
...
-import java.util.Set;
-
-/**
- * An {@code ViewValidator} validates a view against a {@link Schema}
- * {@link uk.gov.gchq.gaffer.data.elementdefinition.view.ViewElementDefinition}.
- * Checks all function input and output types are compatible with the
- * properties and identifiers in the Schema and the transient properties in the
- * View.
- */
-public class ViewValidator {
```
d47853, line 50: Again, for the sake of backwards compatibility, can we maintain this class in the codebase. We don't have to use it but if someone has extended the class to use, they would be pretty annoyed if it disappeared in a non-major release. We can put a note on it saying it'll be removed in V2 but for now, keep it in place.

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/operation/FederatedOperationChainValidator.java
```diff
@@ -1,126 +0,0 @@
-/*
```
d47853, line 1: FederatedStore is marked as experimental so I'm happy for this to be removed.

d47853 (changes requested):
A couple of comments mainly around backwards compatibility. The CI is reporting a failing test: 
FederatedStoreTest.shouldFailGetAllElementsFromSelectedGraphsWithViewOfMissingEntityGroupWhileHasConflictingSchemasDueToDiffVertexSerialiser:1388 exception expected

d47853:
closing in favour of #2376

PR #2393, opened 2021-01-28
gh-2008 bugfix

### LINKED ISSUE #2008: Remove view Edge Entity Validation for operations

GCHQDev404:
# Related Issue

- Resolve #2008


d47853:
closing in favour of #2394

PR #2403, opened 2021-02-16
Gh 2402 integration test support for multiple rest services

m29827:


core/cache/pom.xml
```diff
@@ -3,7 +3,7 @@
     <parent>
         <artifactId>core</artifactId>
         <groupId>uk.gov.gchq.gaffer</groupId>
-        <version>1.13.5-SNAPSHOT</version>
+        <version>1.15.1-SNAPSHOT</version>
```
GCHQDev404, line 6: Using maven <dependencyManagement> there should be no reference to versions.

m29827: I think this is OK, it's specifying the parent module version. As an aside the branches containing v2 work have drifted behind develop which is now at 1.16.+ so there is some work to pull in the latest changes and bump the version on these branches eventually to 2.0.0-SNAPSHOT.

GCHQDev404: OHHHHHH god! that's embarrassing.
Yes, this is correct. I thought I was looking at different part of the pom.

r32575:
The conversion of the `rest-api` modules to use `spring-rest` rather than creating a `WAR`  has been put on hold.

Issue #2400, opened 2021-02-03
Fix scripts in demos


d47853:
Demo scripts in develop-v2 are out of date with the changes to the codebase. Update these so they work as before.

javadev001001:
Branch pushed 6th Apr, updated to use spring-boot maven plugin

n3101:
Include this in a later alpha that reworks the demos.

n3101:
This was addressed in unlinked  PR #2406 . However that has been closed as OBE. The develop-v2  branch is now obsolete, but the underlying problem here has been addressed by #2608 and this issue can now be closed.

PR #2406, opened 2021-04-06
gh-2400 update example scripts to use spring boot maven plugin

javadev001001:
Update example scripts to use Maven Spring-Boot plugin


GCHQDev404 (changes requested):
retracted

t92549:
Having problems testing this as the branch is out of date

n3101:
Leave this until after dependency up versioning which will include Spring.

GCHQDeveloper314:
Was the intention of this to change the demos to run using Spring Boot instead of using Tomcat? In #2608 they have been improved to use Tomcat 9 instead of 7. The demos seem to work ok, what is the benefit of this PR? Can it be closed now?

n3101:
Overtaken by events

PR #2408, opened 2021-04-09
gh-2364 Fixed typos from assumeTrue to assertTrue

### LINKED ISSUE #2364: Change AssumeTrue typo to AssertTrue
Why has AssumeTrue been used in OperationTest?
It is the last line in a test, there is then no further test to abort/skip if found false.
When found false the test does not fail, it just aborts.

If this is incorrect then the use of assumeTrue need to be reviewed in `OperationTest`, `JSONSerialisationTest` `HazelcastCacheServiceTest`

t92549:
Resolves #2364


t92549:
It would seem that fixing these tests uncovered a failing test that was previously being skipped: https://github.com/gchq/Gaffer/pull/2408/checks?check_run_id=2306916330#step:6:319

GCHQDev404 (approved):

PR #2418, opened 2021-04-28
Gh 2205 jsonserialiser

hj325488:
Added special handling for String values which cannot be parsed using the Jackson deserialiser - only in cases where the specific output type is not known (i.e. is an Object).


GCHQDev404 (commented):
Question/Sense Check.

Until now this type of serialisation wasn't possible.
So this change should cause ZERO (near enough) issues with deserialising existing data out in the wild? correct?

hj325488:
> Question/Sense Check.
> 
> Until now this type of serialisation wasn't possible.
> So this change should cause ZERO (near enough) issues with deserialising existing data out in the wild? correct?
> 
> Otherwise I'm happy with this PR.

Yes, that's right - anything that's formatted as a JSON object (or a raw, valid JSON type object) will continue to be parsed as usual, and this change will permit an input type of ```String``` as well. I can add some more test cases to the ```DefaultResponseDeserialiser``` class if that would be useful?

GCHQDev404 (approved):

Issue #2425, opened 2021-05-17
Remove duplicate updateOperationInput


GCHQDev404:
`OperationHandlerUtil.updateOperationInput`
`NamedOperationResolver.updateOperationInput`
`OperationChainHandler.updateOperationInput`

n3101:
@GCHQDev404 Is this a breaking change?

PR #2426, opened 2021-05-17
gh-2425 remove duplicate updateOperationInput

GCHQDev404:


n3101:
@GCHQDev404 Is this a breaking change?

GCHQDev404:
@n3101 

I wouldn't consider it a breaking change, you need someone in the wild extending the class and overriding the fundamental inputUpdate.

would you prefer me to add back in protected method and a junk call out to util?

```  
 protected void updateOperationInput(final Operation op, final Object result) {
        OperationHandlerUtil.updateOperationInput(op, result);
    }
```

n3101:
@GCHQDev404 I refer your question to  @d21211122 to answer

GCHQDev404:
> @n3101
> 
> I wouldn't consider it a breaking change, you need someone in the wild extending the class and overriding the fundamental protected inputUpdate. You might as well implement your own Handler or override the doOperation.
> 
> Would you prefer me to add back in protected method and a junk call out to util?
> 
> ```
>  protected void updateOperationInput(final Operation op, final Object result) {
>         OperationHandlerUtil.updateOperationInput(op, result);
>     }
> ```

I'm going to self answer no. I because it would be a dead hanging method. because I would call the util method. If someone wants to override how to update the input of operations in a chain then can self implement the method.

PR #2470, opened 2021-07-08
Added hotfix check that skips merge develop

### LINKED ISSUE #2468: Change release action to allow for hotfixes
https://github.com/gchq/Gaffer/issues/2463 needs to be fixed in a 1.17.1 hotfix release. This could be done manually, but it would be faster and less error prone to do this through GitHub Actions like our normal release. 
A normal release cannot be used though because we don't want to merge all the changes from develop into the 1.17.1 release, they should instead be put into the next 1.18.0 release. Therefore, the release action should be changed so that if 'hotfix' is in the description of the milestone, develop isn't merged into master, it is instead assumes that the hotfix has already been merged into master.

t92549:
Also removed use of `.m2/repository` cache as it is unused in a release

# Related Issue

- Resolve #2468

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -32,6 +32,7 @@ jobs:
         git config user.email github-actions@github.com
 
     - name: Merge develop into master
+      if: ${{ ! contains(github.event.milestone.description, 'hotfix') }}
       run: |
         git remote set-branches --add origin develop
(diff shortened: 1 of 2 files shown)


GCHQDev404 (approved):
Reviewed with Challenge & Response with the author.

PR #2479, opened 2021-07-16
gh-2457 FederatedGraphStoreage double-caching issue

### LINKED ISSUE #2457: FederatedGraphStore double-caching causes desync issues in replicated deployment
FederatedGraphStore maintains both a uk.gov.gchq.gaffer.cache.Cache of the various Graphs managed by a federated store, and it's own private Map<FederatedAccess, Set<Graph>> named 'storage' (the latter storing deserialised objects).

There is no mechanism to ensure that the Map reflects any changes in the Cache.  This could be caused, for instance, by a deployment scenario where the customer is running multiple instances of Gaffer and trying to use a shared CacheLoader implementation to share data between them.  This is demonstrated (mostly.  Sorta) by the unit test in https://github.com/sw96411/Gaffer/commit/15a2cdd60f8aa59bdb16674e2b963d3fe62864f8

The trivial solution is simply to abandon the use of the private Map and always deserialise the objects from the Cache on demand.  I'm reluctant to suggest this, though, as it would greatly increase the cost of calling any of the methods on FederatedStore.

Other options available include refreshing the Map from the Cache at a maximum frequency (even once per second would in theory save a lot of deserialisation), writing a serial number to the cache on modification and then testing this serial number on read, or creating a specialisation of Cache which can record the last time the Cache (or part thereof) was modified and then delegating the handling of this issue off to that somehow.

I'd be happy to PR in a solution based on any of the above suggestions, but before I do, I wanted to check with the collective:

1. Is this scenario (multiple "hot spare" instances of Gaffer collaborating around a shared cache) even meant to be supported?
2. How does this issue interact with https://github.com/gchq/Gaffer/issues/2435?  Is this effectively a duplicate?

Tagging in @GCHQDev404 as I see they have made commits to related issues and @m29827
as they were the last editor on FederatedGraphStorage.  

Grateful for thoughts from any of the community, of course!

GCHQDev404:
# Related Issue

- Resolve #2457


core/graph/src/main/java/uk/gov/gchq/gaffer/graph/GraphSerialisable.java
```diff
@@ -93,7 +91,11 @@ public Graph getGraph() {
      */
     @JsonIgnore
     public Graph getGraph(final GraphLibrary library) {
-        if (null == graph) {
+        if (isNull(graph)) {
+
+            //TODO FS use GraphDelegate to create and verify just like if a User was Adding.
```
GCHQDev404, line 96: Write a new ticket

GCHQDev404:
This PR got damaged when '-merge-alpha1 was merged into the branch. it dragged this PR with in. It would need to be unpicked to get back to Gaffer1 standard.

PR #2486, opened 2021-08-13
gh-2483: Added StringSerialiser changes and amended tests

### LINKED ISSUE #2483: Ensure StringSerialiser can handle both empty strings and null values
Currently, the [`StringSerialiser`](https://github.com/gchq/Gaffer/blob/eec0c3f1e9851e1310b7345b1fe97d812c497608/core/serialisation/src/main/java/uk/gov/gchq/gaffer/serialisation/implementation/StringSerialiser.java) deserialises both empty string values and `null` values to the empty string `""`:
https://github.com/gchq/Gaffer/blob/eec0c3f1e9851e1310b7345b1fe97d812c497608/core/serialisation/src/main/java/uk/gov/gchq/gaffer/serialisation/implementation/StringSerialiser.java#L48-L51
This can create strange cases where expected properties are not added but are aggregated. For example, when missing properties are aggregated with `StringConcat`, they are aggregated as comma separated empty strings: `,`.

This is because gh-485 changed the `StringSerialiser` to allow for encoding/decoding of empty strings instead of decoding them as `null`, but now `null` is decoded as an empty string.

As this could potentially change expected behaviour we may pile this change in with Gaffer 2.

t92549:
Potentially breaking changes that will need to be tested more and go in Gaffer 2.

# Related Issue

- Resolve #2483


store-implementation/accumulo-store/src/test/java/uk/gov/gchq/gaffer/accumulostore/key/core/impl/bytedEntity/ByteEntityAccumuloElementConverterTest.java
```diff
@@ -72,14 +72,15 @@ public void shouldSerialiseWithHistoricColumnQualifier() throws Exception {
                 put(AccumuloPropertyNames.COLUMN_QUALIFIER_4, Integer.MIN_VALUE);
             }
         };
-        byte[] historicColumnQualifierBytes = {4, 1, 0, 0, 0, 4, -1, -1, -1, 127, 4, 3, 0, 0, 0, 4, 0, 0, 0, -128};
+        byte[] historicColumnQualifierBytes = {4, 1, 0, 0, 0, 4, -1, -1, -1, 127, 4, 3, 0, 0, 0, 4, 0, 0, 0, -128, 0};
```
GCHQDev404, line 75: Breaking change

t92549: The only reason a 0 is added to the end of the byte array is because I changed the test to use "schemaWithVisibilities", and therefore, the 0 is used to serialise that extra property. Can confirm that this works with prior versions of Gaffer. To guarantee this though, perhaps this test should be duplicated, once with and once without visibilities in current Gaffer, then this branch can just prove that it passes both?

core/serialisation/src/test/java/uk/gov/gchq/gaffer/serialisation/ToBytesSerialisationTest.java
```diff
@@ -27,16 +27,18 @@
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertNotSame;
 import static org.junit.jupiter.api.Assertions.assertNull;
+import static uk.gov.gchq.gaffer.serialisation.ToBytesSerialiser.EMPTY_BYTES;
 
 public abstract class ToBytesSerialisationTest<T> extends SerialisationTest<T, byte[]> {
 
+    @Test
     @Override
     public void shouldSerialiseNull() throws SerialisationException {
```
GCHQDev404, line 36: Test is called "should serialise Null" but your using "EmptyBytes" ?

store-implementation/accumulo-store/src/main/java/uk/gov/gchq/gaffer/accumulostore/key/core/AbstractCoreKeyAccumuloElementConverter.java
```diff
@@ -220,8 +220,6 @@ public String getGroupFromColumnFamily(final byte[] columnFamily) {
                     } catch (final SerialisationException e) {
                         throw new AccumuloElementConversionException(e.getMessage(), e);
                     }
-                } else {
-                    rtn = serialiser.serialiseNull();
```
GCHQDev404, line 224: So if there is no property, it uses empty byte ?

store-implementation/accumulo-store/src/main/java/uk/gov/gchq/gaffer/accumulostore/key/core/AbstractCoreKeyAccumuloElementConverter.java
```diff
@@ -286,7 +284,17 @@ protected void serialiseSizeAndPropertyValue(final String propertyName, final Sc
             } else {
                 Object value = properties.get(propertyName);
                 //serialiseNull could be different to AccumuloStoreConstants.EMPTY_BYTES
-                bytes = (null == value) ? serialiser.serialiseNull() : serialiser.serialise(value);
+                if (null != value) {
+                    bytes = serialiser.serialise(value);
+                } else if (null != this.schema.getVisibilityProperty()) {
+                    if (this.schema.getVisibilityProperty().equals(propertyName)) {
+                        bytes = serialiser.serialise("");
```
GCHQDev404, line 291: can you change "" to be a constant please.

store-implementation/accumulo-store/src/test/java/uk/gov/gchq/gaffer/accumulostore/integration/AccumuloAggregationIT.java
```diff
@@ -239,10 +235,8 @@ public void shouldHandleAggregationWhenGroupByPropertiesAreNull() throws Operati
         final Entity expectedEntity = new Entity.Builder()
                 .vertex(VERTEX)
                 .group(TestGroups.ENTITY)
-                .property(AccumuloPropertyNames.COLUMN_QUALIFIER, ",") //String Aggregation is combining two empty strings -> "",""
```
GCHQDev404, line 242: What happens now in this scenario ? Does entity1 need to be modified for testing the comments?

core/serialisation/src/test/java/uk/gov/gchq/gaffer/serialisation/ToBytesSerialisationTest.java
t92549: In the default case (and how Gaffer has always worked), serialisers will return `EMPTY_BYTES` when you `serialiseNull()`. Slightly confusing naming but it is called that because it is an empty byte array.
This PR changes it so that specifically `StringSerialiser` will return `NULL_BYTES` when you `serialiseNull()`, with that meaning a byte array with just a 0 byte in it: `new byte[]{0}`.

store-implementation/accumulo-store/src/main/java/uk/gov/gchq/gaffer/accumulostore/key/core/AbstractCoreKeyAccumuloElementConverter.java
t92549: Will do

store-implementation/accumulo-store/src/main/java/uk/gov/gchq/gaffer/accumulostore/key/core/AbstractCoreKeyAccumuloElementConverter.java
t92549: Yes, this is a corner case where if visibility is missing (null) or empty string, it will still be deserialised as empty string.

store-implementation/accumulo-store/src/test/java/uk/gov/gchq/gaffer/accumulostore/integration/AccumuloAggregationIT.java
t92549: Because they were added as `null` it is the same as not adding them at all. I think this behaviour is okay.

store-implementation/accumulo-store/src/test/java/uk/gov/gchq/gaffer/accumulostore/key/core/impl/bytedEntity/ByteEntityAccumuloElementConverterTest.java
t92549: Me and @GCHQDev404 have decided that the best way forward is to leave this test alone and add a second test that uses the visibility schema and merge that into Gaffer first. Then this branch can just show that it passes both tests. See #2488

store-implementation/accumulo-store/src/main/java/uk/gov/gchq/gaffer/accumulostore/key/impl/AggregatorIterator.java
```diff
@@ -57,6 +57,9 @@
     public Value reduce(final Key key, final Iterator<Value> iter) {
```
t92549, line 57: Should the `AggregatorIterator` break when reducing an empty iterator? I fixed this as I thought it was strange but perhaps it is better leaving it?

store-implementation/accumulo-store/src/main/java/uk/gov/gchq/gaffer/accumulostore/key/impl/AggregatorIterator.java
```diff
@@ -57,6 +57,9 @@
...
         // Get first Value. If this is the only Value then return it straight
         // away;
+        if (!iter.hasNext()) {
+            return null;
```
GCHQDev404, line 61: can you re-arrange to have 1 return statement.

store-implementation/hbase-store/src/main/java/uk/gov/gchq/gaffer/hbasestore/serialisation/ElementSerialisation.java
```diff
@@ -83,13 +83,19 @@ public ElementSerialisation(final Schema schema) {
                 try {
                     if (null != serialiser) {
                         Object value = properties.get(propertyName);
+                        byte[] bytes;
                         if (null != value) {
-                            final byte[] bytes = serialiser.serialise(value);
-                            writeBytes(bytes, out);
+                            bytes = serialiser.serialise(value);
+                        } else if (null != this.schema.getVisibilityProperty()) {
```
GCHQDev404, line 89: nonNull()

store-implementation/hbase-store/src/main/java/uk/gov/gchq/gaffer/hbasestore/serialisation/ElementSerialisation.java
```diff
@@ -194,8 +200,6 @@ public Properties getProperties(final String group, final Cell cell) throws Seri
                     } catch (final SerialisationException e) {
                         throw new SerialisationException(e.getMessage(), e);
                     }
-                } else {
-                    return serialiser.serialiseNull();
```
GCHQDev404, line 198: What happens now ?

store-implementation/federated-store/src/test/java/uk/gov/gchq/gaffer/federatedstore/FederatedStoreSchemaTest.java
```diff
@@ -117,10 +115,7 @@ public void shouldBeAbleToAddGraphsWithSchemaCollisions() throws OperationExcept
         Collection<String> graphIds = fStore.getAllGraphIds(testUser);
 
         // Then
-        HashSet<String> expected = new HashSet<>();
-        expected.addAll(Arrays.asList("a", "b", "c"));
-
-        assertEquals(expected, graphIds);
+        assertThat(graphIds).containsExactlyInAnyOrder("a", "b", "c");
```
GCHQDev404, line 118: I think order did matter.

GCHQDev404 (changes requested):

core/serialisation/src/test/java/uk/gov/gchq/gaffer/serialisation/implementation/StringSerialiserTest.java
```diff
@@ -55,6 +60,15 @@ public void canSerialiseStringClass() {
         assertTrue(serialiser.canHandle(String.class));
     }
 
+    @Test
+    public void shouldSerialiseEmpty() throws SerialisationException {
+        // When
+        final byte[] bytes = serialiser.serialise("");
+
+        // Then
+        assertArrayEquals(EMPTY_BYTES, bytes);
```
GCHQDev404, line 69: can you have little look to see if other things might be getting empty concats too.

`,,,,,`

store-implementation/hbase-store/src/main/java/uk/gov/gchq/gaffer/hbasestore/serialisation/ElementSerialisation.java
t92549: The previous return is no longer hit so instead it goes down to `return HBaseStoreConstants.EMPTY_BYTES;`

store-implementation/federated-store/src/test/java/uk/gov/gchq/gaffer/federatedstore/FederatedStoreSchemaTest.java
t92549: HashSets are not ordered so order wasn't checked before and `graphIds` is returned in a different order every time as it is also made from a Set

core/serialisation/src/main/java/uk/gov/gchq/gaffer/serialisation/ToBytesViaStringDeserialiser.java
```diff
@@ -56,6 +57,9 @@ public final T deserialise(final byte[] bytes) throws SerialisationException {
     @Override
     public final T deserialise(final byte[] allBytes, final int offset, final int length) throws SerialisationException {
         try {
+            if (Arrays.equals(Arrays.copyOfRange(allBytes, offset, offset + length), NULL_BYTES)) {
+                return null;
```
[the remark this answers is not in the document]
t92549: This isn't really the kind of multiple return that makes the code harder to read, I actually think it is easier to read and [less verbose](https://refactoring.com/catalog/replaceNestedConditionalWithGuardClauses.html). Debugging would also be the same as the breakpoint would not be hit if inside an else block.

store-implementation/hbase-store/src/main/java/uk/gov/gchq/gaffer/hbasestore/serialisation/ElementSerialisation.java
t92549: This isn't uniform across the codebase and is preference, I personally find `null != x` easier to read.

core/serialisation/src/main/java/uk/gov/gchq/gaffer/serialisation/ToBytesViaStringDeserialiser.java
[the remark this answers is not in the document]
GCHQDev404: I disagree.
The example you provided is talking about using multiple `else` then `if` statements. 
You don't have that here. The first example is horrendous and verbose anything will look better than that ugly first example. You'd use `if else` on a single line and you still require the encapsulating braces for CheckStyle in our project.

```java 
function getPayAmount() {
 rtn ;
  if (isDead) { rtn = deadAmount(); }
  else if (isSeparated) { rtn =  separatedAmount(); }
  else if (isRetired) { rtn =  retiredAmount(); }
  else { rtn =  normalPayAmount(); }
  return rtn;
 //single return 
}
```


 This code being examined in our code is single `If` statement with no nesting. Multiple returns provide little to no benefit here.

4 parts & 2 returns which at times would need a debug break point on each return.
```java
// 1 if statement
    if (Arrays.equals(Arrays.copyOfRange(allBytes, offset, offset + length), NULL_BYTES)) {
// 2 return statement
         return null;
     }
// 3 assignment
             String valueString = new String(allBytes, offset, length, charset);
// 4 return statements
             return deserialiseString(valueString);
}             
```


vs 1 return for 1 `if` statement 
```java
T rtn = null;

// 1 if statement
    if (!Arrays.equals(Arrays.copyOfRange(allBytes, offset, offset + length), NULL_BYTES)) {
// 2 assignment
      rtn =  deserialiseString(new String(allBytes, offset, length, charset));
    }
// 3 return
return  rtn
```

This follows a Edsger Dijkstra’s approach of having 1 exit from a method/function.

if you really want to just smash in a ternary statement. can't get less verbose than that.
```java
return  (!Arrays.equals(Arrays.copyOfRange(allBytes, offset, offset + length), NULL_BYTES)) 
    ? deserialiseString(new String(allBytes, offset, length, charset))
    : null;
```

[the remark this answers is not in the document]
t92549: Single exit code style is [outdated](https://softwareengineering.stackexchange.com/questions/118703/where-did-the-notion-of-one-return-only-come-from) and I still think the example you provide for the `getPayAmount` function still looks worse and won't work nicely in a debugger.
Having said that, I do think that flipping the if statement like you did makes this function a lot cleaner so I will amend it. Cheers for the feedback!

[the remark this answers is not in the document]
GCHQDev404: > example you provide for the getPayAmount function still looks worse and won't work nicely in a debugger.

It has more characters to allow the single return type yes.

But with the example given when you scanning through code to find where an unusual value is being created
you would require 4 break points and you'd need to step/in/over to view the value. The debug stops on line ` if (isSeparated) { return separatedAmount(); }` but you don't know the value of this line, you gotta manipulate the debug step in/out to get the value of `separatedAmount()`

```Java
function getPayAmount() { 
  if (isDead) { return deadAmount(); } //break 1
  if (isSeparated) { return separatedAmount(); } //break 2
  if (isRetired) { return retiredAmount(); } //break 3
  return { normalPayAmount(); } //break 4
}
```

with the single return the breakpoint, it will have already resolved the value of this method `rtn` in a single hit at the breakpoint, without needing to tab in/tab out of methods to resolve on the breakpoint line. The `rtn` value is visible and ready
```java
function getPayAmount() {
 rtn ;
  if (isDead) { rtn = deadAmount(); }
  else if (isSeparated) { rtn =  separatedAmount(); }
  else if (isRetired) { rtn =  retiredAmount(); }
  else { rtn =  normalPayAmount(); }
  return rtn; //break 1 !!!!!!
}
```

^ this is why I like to write like this I think it looks cleaners and debugging has less steps and tabbing around.

yeah Edsger Dijkstra’s is super old school. 
I like it as an approach to writing as it is usually cleaner to have 1 return partially for what I perceive/describe above^

> "Structured Programming
> Some programmers follow Edsger Dijkstra’s rules of structured programming.14 Dijkstra said that every function, and every block within a function, should have one entry and one exit. Following these rules means that there should only be one return statement in a func- tion, no break or continue statements in a loop, and never, ever, any goto statements.
> While we are sympathetic to the goals and disciplines of structured programming, those rules serve little benefit 
> **when functions are very small. It is only in larger functions that such rules provide significant benefit.
> So if you keep your functions small, then the occasional multiple return, break, or continue statement does no harm and can sometimes even be more expressive than the sin- gle-entry, single-exit rule.** 
> " [Clean Code: A Handbook of Agile Software Craftsmanship (Robert C. Martin) (Robert C. Martin Series)]

I don't see a an advantage to having 4 returns here, when 1 will do.

This has been/become an exercise/discussion programming approaches. 

You can keep this code as you like. But I hope you understand the debug breakpoint reasoning I use/have.

When your in Debug Hell you want every little morsel on your side :D

store-implementation/accumulo-store/src/main/java/uk/gov/gchq/gaffer/accumulostore/key/impl/AggregatorIterator.java
GCHQDev404: *shrug*
if it needed/should/best to break then a comment should notify so.

store-implementation/hbase-store/src/main/java/uk/gov/gchq/gaffer/hbasestore/serialisation/ElementSerialisation.java
GCHQDev404: :D when one is seen it gets changed. This is best way to avoid a ticket to change them all. :D
I request this in all PR.
nonNull is 100% more human readable :D it has words in it for a start.

store-implementation/federated-store/src/test/java/uk/gov/gchq/gaffer/federatedstore/FederatedStoreSchemaTest.java
GCHQDev404: nice one.

store-implementation/hbase-store/src/main/java/uk/gov/gchq/gaffer/hbasestore/serialisation/ElementSerialisation.java
t92549: Is `!=` reinventing the wheel? I would argue `nonNull` is

core/serialisation/src/main/java/uk/gov/gchq/gaffer/serialisation/ToBytesViaStringDeserialiser.java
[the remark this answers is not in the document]
t92549: I see what you mean about debugging and in large functions like the quote says, I think this approach can be helpful. But in 4 line functions I think debugging is easy enough. You are right though it is preference at this point. I did modify it though to swap the if statement as it looks nicer that way.

store-implementation/hbase-store/src/main/java/uk/gov/gchq/gaffer/hbasestore/serialisation/ElementSerialisation.java
GCHQDev404: Your not implementing it and its part of core Java API. so you'd be"making use of core Java API"
but if you feel really strongly to not have it.

> "API Note: This method exists to be used as a Predicate, filter(Objects::isNull)"

t92549: Yeah, the method was made to make predicates nicer than `filter(x -> x == null)`. The implementation is just
```java
 public static boolean isNull(Object obj) {
     return obj == null;
 }
```
Personally it is not nicer to read so I think it is just a preference thing.

store-implementation/accumulo-store/src/main/java/uk/gov/gchq/gaffer/accumulostore/key/impl/AggregatorIterator.java
t92549: Still not sure if this fix should even be here. Should `AggregatorIterator.reduce` break or return null when asked to aggregate null? This iterator runs within Accumulo to perform reductions

core/serialisation/src/test/java/uk/gov/gchq/gaffer/serialisation/implementation/StringSerialiserTest.java
t92549: @GCHQDev404  So I did some digging in the serialisers that implement `ToBytesSerialiser` and found some interesting results. For most serialisers, like Long, Float and Boolean, this problem doesn't apply as those data types cannot be initialised empty.

But for quite a few others there is indeed an issue with serialising empty values and having them deserialised as `null`, like what used to happen with String before this PR. For example, if you serialise `new HashMap()` with the `MapSerialiser` it will be deserialised as `null`, not an empty map.

Some serialisers behave the opposite way too. For example, using a `SetSerialiser`, when `null` is serialised and then deserialsed the result is an empty `Set`.

However, these inconsistencies are not quite as bad as the old `StringSerialiser`, as `StringConcat` only applies to Strings. For other `BinaryOperators` that aggregate properties, such as `First`, this doesn't produce quite such a strange result, but it is still inconsistent.

Either way, I think this discussion on whether we should make all of the serialisers consistent is worthy of another issue and out of the scope of this PR which just changes the `StringSerialiser`. It might be the case though that we do choose to make all of the serialisers consistent and wait to merge this PR until then, perhaps in Gaffer v2.

GCHQDev404: The reason why we had `null` serialsiation is that storing "a lot" of empty maps takes up "a lot" of data.

t92549: That makes sense, I would be okay with leaving it like that but there are a few serialisers, like Set, that do the opposite. If you give it `null` it actually serialises to an empty Set.

t92549:
This PR has been put on hold until a decision has been made regarding whether to update all of the Serialisers to differentiate `null` and empty values. See https://github.com/gchq/Gaffer/pull/2486#discussion_r709438443. If this were to happen, a new issue will be made and it will probably be incorporated into Gaffer v2.

n3101:
@t92549 For planning purposes - does this depend on any other 2.0  work, or could it be done in parallel if effort becomes available?

t92549:
This work can be done in isolation from any 2.0 changes.

Development stalled though because we couldn't decide how best to implement these fixes. They could be implemented as a breaking change, or as additional serialisers.
We also weren't sure if the fixes that add more consistency were worth potentially increasing data volumes.

More discussion needed before more work is done.

n3101:
@t92549 "This work can be done in isolation from any 2.0 changes.

Development stalled though because we couldn't decide how best to implement these fixes. They could be implemented as a breaking change, or as additional serialisers.
We also weren't sure if the fixes that add more consistency were worth potentially increasing data volumes.

More discussion needed before more work is done."

Ok, for now I have labelled this as post-2.0, BUT if we don't do this in 2.0, then that rules out breaking changes. Lets discuss further.

t92549:
Closed because this branch should be adapted to provide an alternate serialiser, rather than breaking the existing one

t92549:
Note the branch will be kept as a reference implementation

PR #2489, opened 2021-08-18
gh-2488: Added ByteEntityAccumuloElementConverter visibility tests

### LINKED ISSUE #2488: Test AccumuloElementConverter with visibility schema
The current tests use a schema without the visibility property.

https://github.com/gchq/Gaffer/blob/eec0c3f1e9851e1310b7345b1fe97d812c497608/store-implementation/accumulo-store/src/test/java/uk/gov/gchq/gaffer/accumulostore/key/AbstractAccumuloElementConverterTest.java#L64

This is the [elements.json](https://github.com/gchq/Gaffer/blob/5170aff18215639e83ca84e2228b8926df1e276b/store-implementation/accumulo-store/src/test/resources/schema/elements.json) used.

The visibility property can be prone to breaking with serialisation changes, so it is probably good to also run these tests against the schema that has a visibility property. This would help in testing the compatibility of #2486

t92549:
Added a new test class that is a subclass of ByteEntityAccumuloElementConverterTest, with the only changes being that it uses the visibility schema and so expects an extra serialised property. 

There is also a new test which creates a serialised Properties from the full column qualifier, rather than just the first two properties.

# Related Issue

- Resolve #2488

### DIFF
--- a/store-implementation/accumulo-store/src/test/java/uk/gov/gchq/gaffer/accumulostore/key/core/impl/bytedEntity/ByteEntityAccumuloElementConverterVisibilityTest.java
+++ b/store-implementation/accumulo-store/src/test/java/uk/gov/gchq/gaffer/accumulostore/key/core/impl/bytedEntity/ByteEntityAccumuloElementConverterVisibilityTest.java
@@ -4,6 +4,7 @@ import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 
 import uk.gov.gchq.gaffer.accumulostore.utils.AccumuloPropertyNames;
+import uk.gov.gchq.gaffer.accumulostore.utils.BytesAndRange;
 import uk.gov.gchq.gaffer.commonutil.StreamUtil;
 import uk.gov.gchq.gaffer.commonutil.TestGroups;
 import uk.gov.gchq.gaffer.data.element.Properties;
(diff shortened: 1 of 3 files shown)


store-implementation/accumulo-store/src/test/resources/schemaWithVisibilities/types.json
```diff
@@ -31,6 +31,9 @@
       "class": "java.lang.String",
       "aggregateFunction": {
         "class": "uk.gov.gchq.koryphe.impl.binaryoperator.StringConcat"
+      },
+      "serialiser": {
+        "class": "uk.gov.gchq.gaffer.serialisation.implementation.StringSerialiser"
```
GCHQDev404, line 36: Do you need two different Json files? is this being used by the mentioned parents test?

store-implementation/accumulo-store/src/test/java/uk/gov/gchq/gaffer/accumulostore/key/core/impl/bytedEntity/ByteEntityAccumuloElementConverterVisibilityTest.java
```diff
@@ -0,0 +1,95 @@
...
+        // Given
+        final Properties properties = new Properties() {
+            {
+                put(AccumuloPropertyNames.COLUMN_QUALIFIER, 1);
+                put(AccumuloPropertyNames.COLUMN_QUALIFIER_2, 2);
+                put(AccumuloPropertyNames.COLUMN_QUALIFIER_3, 3);
+                put(AccumuloPropertyNames.COLUMN_QUALIFIER_4, 4);
+            }
+        };
+        // An extra 0 at the end of the byte array compared to the parent method accounts for the fact that
```
GCHQDev404, line 55: 👍

store-implementation/accumulo-store/src/test/resources/schemaWithVisibilities/types.json
t92549: The parent test uses `store-implementation/accumulo-store/src/test/resources/schema/`, this is a different schema but it was throwing an error as the visibility type didn't have a Serialiser assigned. The only other test that I can see that uses this schema is `InputFormatTest` which doesn't actually add any elements, hence why this was probably missed.

GCHQDev404 (approved):

PR #2508, opened 2021-09-17
gh-2421: Change MapStore GetElements to handle Elements

### LINKED ISSUE #2421: Some stores don't allow Elements as GetElements input
**_(Originally titled: Inconsistent behaviour between Accumulostore & Mapstore)_**

The GetElements operation in the AccumuloStore (and MockAccumuloStore) seemingly accept input consisting of Edges, but MapStore (and SingleUseMapStore) don't; they insist on them being wrapped as EdgeSeeds. Admittedly the examples in the docs only show EntitySeed/EdgeSeed.

An exploration of this issue also revealed a couple of nullpointerexceptions. See https://github.com/gchq/Gaffer/tree/gh-2421-BugHunt

Further exploration found that this error happened with the Parquet store too

t92549:
This should let the input to the `MapStore`'s `GetElements` be an `Element`, as well as well as an `ElementSeed`.
Marked as draft as some test for this might be a good idea.

# Related Issue

- Resolve #2421

### DIFF
--- a/store-implementation/map-store/src/main/java/uk/gov/gchq/gaffer/mapstore/impl/GetElementsUtil.java
+++ b/store-implementation/map-store/src/main/java/uk/gov/gchq/gaffer/mapstore/impl/GetElementsUtil.java
@@ -99,12 +99,13 @@ public final class GetElementsUtil {
         } else {
             relevantElements = new HashSet<>();
 
-            final EdgeId edgeId = (EdgeSeed) elementId;
+            final EdgeId edgeId = (EdgeId) elementId;
+
             if (DirectedType.isEither(edgeId.getDirectedType())) {
                 relevantElements.addAll(mapImpl.lookup(new EdgeSeed(edgeId.getSource(), edgeId.getDestination(), false)));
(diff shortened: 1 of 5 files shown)


store-implementation/map-store/src/main/java/uk/gov/gchq/gaffer/mapstore/impl/GetElementsUtil.java
```diff
@@ -99,12 +99,13 @@ private GetElementsUtil() {
         } else {
             relevantElements = new HashSet<>();
 
-            final EdgeId edgeId = (EdgeSeed) elementId;
```
t92549, line 102: This is the cast that was causing errors. Now we create a new `EdgeSeed` below, whether from an `EdgeSeed` or an `Edge`.

t92549:
This PR adds integration tests to ensure that **all** stores support `GetElements`' inputs being both `ElementSeed`s and `Element`s. This makes sense considering that `input` is of type `ElementId`, of which `ElementSeed` and `Element` are sub-classes of.
In order for these to pass, not only was `MapStore` fixed, but so was `ParquetStore` as it failed the tests.

p3430233 (approved):

PR #2516, opened 2021-09-27
gh-2515: Seperate httpcore and httpclient versions

t92549:


t92549:
I believe this should not be a breaking change and can make it into Gaffer 1

t11947 (approved):

t92549:
Put into draft while decision is made on providing newer library versions that are used by other dependencies. 
It seems that https://github.com/gchq/Gaffer/pull/498 cleaned up which dependencies are needed by supplying specific versions. The risk of updating these beyond what is known to work with old versions of other dependencies might outweigh the risk in just using the current versions. And undoing https://github.com/gchq/Gaffer/pull/498 will lead to a messy dependency tree with still outdated dependencies. Perhaps then it is best to leave Gaffer 1 with outdated dependencies and upgrade them all at once, eliminating a lot of vulnerabilities, with Gaffer 2.

Issue #2519, opened 2021-10-01
MapStore GetElements iterator sometimes breaks when called twice


t92549:
This is quite a niche bug, but when it happens it breaks MapStore's GetElements operation. One example where this does happen is if you try to use a MapStore as a Gaffer Results Cache.

To recreate it, you must create a MapStore where there are some elements in the schema which have aggregation set to false. As well as this, the schema needs a visibility property. Then you need to insert data with an empty visibility field. 

When this is done and you call GetElements, if you call `.iterator().hasNext()` on the results, the first time it will be true, but the second time you do it, it will be false. This is because the act of calling `.iterator()` changes the data in the MapStore which is a bug.

This happens because when the element that is fetched from the map has an empty visibility field, this function adds a default empty string:
https://github.com/gchq/Gaffer/blob/897253adf6c4b8a0beaf4ee21d3cb9f1d300c6ac/store-implementation/map-store/src/main/java/uk/gov/gchq/gaffer/mapstore/impl/GetElementsUtil.java#L156-L159
This is good because it replicates the Serialiser behaviour when using an AccumuloStore. The bug is that this changes the actual data in the MapStore though, rather than a copy of the element fetched due to this code returning a reference to the element: 
https://github.com/gchq/Gaffer/blob/897253adf6c4b8a0beaf4ee21d3cb9f1d300c6ac/store-implementation/map-store/src/main/java/uk/gov/gchq/gaffer/mapstore/impl/MapImpl.java#L158-L164
The reason this doesn't effect most MapStore graphs is because the equivalent code for aggregated elements does return a clone: https://github.com/gchq/Gaffer/blob/897253adf6c4b8a0beaf4ee21d3cb9f1d300c6ac/store-implementation/map-store/src/main/java/uk/gov/gchq/gaffer/mapstore/impl/MapImpl.java#L166-L171

The reason that this iterator breaks when called twice is that the first time the data is fetched, the data is left changed. Then when you do another lookup by calling the iterator again, a result is not found as the data was changed in the first operation.

A fix would be to make the `getNonAggElements` function return a clone as well.

PR #2520, opened 2021-10-01
gh-2519: Fixed MapImpl changing data when getting

t92549:


GCHQDev404 (approved):
Challenge & Response Review.

PR #2521, opened 2021-10-15
Gh 1867 csv element generator 2c

### LINKED ISSUE #1867: Create CSV Element Generator Functions
Create a CsvElementGenerator that takes a CSV and generate Edges and Entities to allow CSVs to be easily ingested into Gaffer.

This could also be exposed as an Operation so it could be used within the Federated Store.

p3430233:
# Related Issue

- Resolve #1867

### DIFF
--- a/library/time-library/src/main/java/uk/gov/gchq/gaffer/time/function/DateToTimeBucketEnd.java
+++ b/library/time-library/src/main/java/uk/gov/gchq/gaffer/time/function/DateToTimeBucketEnd.java
@@ -17,6 +17,7 @@
 package uk.gov.gchq.gaffer.time.function;
 
 import uk.gov.gchq.gaffer.time.CommonTimeUtil;
+import uk.gov.gchq.gaffer.time.CommonTimeUtil.TimeBucket;
 import uk.gov.gchq.koryphe.Since;
 import uk.gov.gchq.koryphe.Summary;
 import uk.gov.gchq.koryphe.function.KorypheFunction;
@@ -27,17 +28,17 @@ import static java.util.Objects.isNull;
 
 /**
  * Converts a Date into the end of a timestamp bucket, based on a provided
- * {@link CommonTimeUtil.TimeBucket}.
+ * {@link TimeBucket}.
  */
 @Since("1.21.0")
 @Summary("Converts a Date into the end of a timestamp bucket, based on a provided TimeBucket")
 public class DateToTimeBucketEnd extends KorypheFunction<Date, Date> {
-    private CommonTimeUtil.TimeBucket bucket;
+    private TimeBucket bucket;
 
     public DateToTimeBucketEnd() {
     }
 
-    public DateToTimeBucketEnd(final CommonTimeUtil.TimeBucket bucket) {
+    public DateToTimeBucketEnd(final TimeBucket bucket) {
         this.bucket = bucket;
     }
 
@@ -49,11 +50,11 @@ public class DateToTimeBucketEnd extends KorypheFunction<Date, Date> {
         return new Date(CommonTimeUtil.timeToBucketEnd(date.getTime(), bucket));
     }
 
-    public CommonTimeUtil.TimeBucket getBucket() {
+    public TimeBucket getBucket() {
         return bucket;
     }
 
-    public void setBucket(final CommonTimeUtil.TimeBucket bucket) {
+    public void setBucket(final TimeBucket bucket) {
         this.bucket = bucket;
     }
 }
--- a/example/road-traffic/road-traffic-generators/src/test/java/uk/gov/gchq/gaffer/traffic/generator/RoadTrafficCsvElementGenerator2Test.java
+++ b/example/road-traffic/road-traffic-generators/src/test/java/uk/gov/gchq/gaffer/traffic/generator/RoadTrafficCsvElementGenerator2Test.java
@@ -28,7 +28,7 @@ import uk.gov.gchq.gaffer.data.element.function.TuplesToElements;
 import uk.gov.gchq.gaffer.data.util.ElementUtil;
 import uk.gov.gchq.gaffer.jsonserialisation.JSONSerialiser;
 import uk.gov.gchq.gaffer.sketches.clearspring.cardinality.HyperLogLogPlusEntityGenerator;
-import uk.gov.gchq.gaffer.time.CommonTimeUtil;
+import uk.gov.gchq.gaffer.time.CommonTimeUtil.TimeBucket;
 import uk.gov.gchq.gaffer.time.function.DateToTimeBucketEnd;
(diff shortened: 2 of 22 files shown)


t92549 (changes requested):
Some Util functions and Koryphe functions need to be tested: https://app.codecov.io/gh/gchq/Gaffer/compare/develop...gh-1867-csv-element-generator-2c/diff

library/time-library/src/main/java/uk/gov/gchq/gaffer/time/BoundedTimestampSet.java
```diff
@@ -25,9 +25,8 @@
 import org.apache.commons.lang3.builder.EqualsBuilder;
 import org.apache.commons.lang3.builder.HashCodeBuilder;
 
-import uk.gov.gchq.gaffer.commonutil.CommonTimeUtil;
```
t92549, line 28: A lot of files changes are because `CommonTimeUtil` was moved from the common-util package to the time package

core/data/src/test/java/uk/gov/gchq/gaffer/data/element/function/ToElementTupleTest.java
```diff
@@ -31,7 +31,7 @@
 import static org.junit.jupiter.api.Assertions.assertNotNull;
 import static org.junit.jupiter.api.Assertions.assertNull;
 
-public class ToElementTupleTest extends FunctionTest {
+public class ToElementTupleTest extends FunctionTest<ToElementTuple> {
```
t92549, line 34: All of the `FunctionTest`s were changed in this PR to be correctly parameterised

core/data/src/test/java/uk/gov/gchq/gaffer/data/element/function/ReduceRelatedElementsTest.java
```diff
@@ -378,7 +378,7 @@ protected ReduceRelatedElements getInstance() {
     }
 
     @Override
-    protected Iterable getDifferentInstancesOrNull() {
+    protected Iterable<ReduceRelatedElements> getDifferentInstancesOrNull() {
```
t92549, line 381: As part of parameterising all of the `FunctionTest`s, function signatures like this were corrected

core/data/src/main/java/uk/gov/gchq/gaffer/data/element/function/ElementTransformer.java
```diff
@@ -31,6 +34,8 @@
  * An {@code ElementTransformer} is a {@link Function} which applies a series of
  * transformations to an {@link Element}.
  */
+@Since("0.3.0")
+@Summary("A Function which applies a series of transformations to a Properties object")
```
t511203, line 38: This summary doesn't match this class.

core/type/src/main/java/uk/gov/gchq/gaffer/types/function/ToTypeValue.java
```diff
@@ -22,14 +22,32 @@
...
 @Since("1.8.0")
 @Summary("Converts a value into a TypeValue")
 public class ToTypeValue extends KorypheFunction<Object, TypeValue> {
+    private String type;
+
+    public ToTypeValue(final String type) {
+        setType(type);
+    }
+
+    public ToTypeValue() {
```
t511203, line 37: Is this redundant? If the default no-arg constructor is used, then the type String is null anyway?

library/time-library/src/main/java/uk/gov/gchq/gaffer/time/function/DateToTimeBucketEnd.java
```diff
@@ -0,0 +1,59 @@
...
+ */
+@Since("1.21.0")
+@Summary("Converts a Date into the end of a timestamp bucket, based on a provided TimeBucket")
+public class DateToTimeBucketEnd extends KorypheFunction<Date, Date> {
+    private CommonTimeUtil.TimeBucket bucket;
+
+    public DateToTimeBucketEnd() {
+    }
+
+    public DateToTimeBucketEnd(final CommonTimeUtil.TimeBucket bucket) {
```
t511203, line 40: It might be a bit easier to read if these references were made less explicit (via import)

t92549: I agree it looks ugly but there are other places already in the codebase where it is done like this: https://github.com/gchq/Gaffer/blob/fc9fe8e52ab64667ca5fbec80deef0a5ac36a27a/library/time-library/src/main/java/uk/gov/gchq/gaffer/time/serialisation/BoundedTimestampSetSerialiser.java#L82

Should I change it everywhere or just the files being added in this PR?

core/data/src/main/java/uk/gov/gchq/gaffer/data/element/function/ElementTupleDefinition.java
```diff
@@ -0,0 +1,88 @@
...
+    public ElementTupleDefinition(final String group) {
+        put(GROUP, group);
+    }
+
+    public String getGroup() {
+        return (String) get(GROUP);
+    }
+
+    @Override
+    public Object get(final Object key) {
```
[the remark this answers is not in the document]
t92549: I agree but strangely enough when I change the type to String I get an error as it is overriding `HashMap` and the signature is `public V get(java.lang.Object arg0)`

core/data/src/main/java/uk/gov/gchq/gaffer/data/element/function/PropertiesFilter.java
```diff
@@ -0,0 +1,167 @@
...
+        if (readOnly) {
+            return Collections.unmodifiableList(super.getComponents());
+        }
+
+        return super.getComponents();
+    }
+
+    public void lock() {
+        readOnly = true;
+    }
```
GCHQDeveloper314, line 95: Very generally comment, but I would expect to see a matching unlock. It seems strange that once locked there is no way to reverse this.

GCHQDeveloper314 (commented):
i've had a look through and nothing really stands out. I'm not qualified to approve so this is just a general comment.

core/data/src/main/java/uk/gov/gchq/gaffer/data/element/function/PropertiesFilter.java
t92549: Hmm you're right, seems it is the same as how `ElementFilter` currently works: 

https://github.com/gchq/Gaffer/blob/b600adedaebc9d7eb8d2dd44da1b05b98fd7bc7f/core/data/src/main/java/uk/gov/gchq/gaffer/data/element/function/ElementFilter.java#L90-L93

I wonder if there is a reason for this

t92549 (approved):

PR #2534, opened 2021-11-26
gh-2500: Added release branches work

### LINKED ISSUE #2500: Update release pipeline to use release branches
Currently, the Gaffer release pipeline first merges develop into master then performs subsequent release steps. This should be changed because:
- This approach does not properly follow the Gitflow branching model, which uses release branches (see below)
- Release branches provide an opportunity to monitor the health of a build and add bugfixes.
- Having an intermediate step in between starting a release and it finishing reduces the risk of erroneous releases: #2467
- With the current setup, hotfixes are hard and cause merge issues: #2475

Therefore, before the release pipeline merges the code into master and performs the full release, a release branch should be made with updated versions. Then, a PR should be made to merge this branch into master. When merged, this will trigger the rest of the pipeline.

### Gitflow branching model:
Git-branching-model

t92549:
# Related Issue

- Resolve #2500


.github/workflows/create-alpha-release-branch.yaml
```diff
@@ -0,0 +1,80 @@
...
+
+on:
+  workflow_dispatch:
+    inputs:
+      version:
+        description: 'Alpha Release Branch Version'
+        required: false
+
+env:
+  artifactId: gaffer2
```
r32575, line 11: As it is an environment variable should it be `ARTIFACT_ID`

.github/workflows/create-alpha-release-branch.yaml
```diff
@@ -0,0 +1,80 @@
...
+      if: ${{ !github.event.inputs.version }}
+      run: echo "RELEASE_VERSION=$(echo ${{ steps.bump-semver.outputs.new_version }} )" >> $GITHUB_ENV
+
+    - name: Verify version regex
+      uses: actions-ecosystem/action-regex-match@v2
+      id: regex-match
+      with:
+        text: ${{ env.RELEASE_VERSION }}
+        # Ensure version starts in 2
+        regex: '^(2)\.(0|[1-9]\d*)\.(0|[1-9]\d*)(?:-((?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\+([0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?$'
```
r32575, line 57: A bit messy and lets you enter in strange version numbers such as `2.0.0-a0-+0-` (see [https://regex101.com/](url)).

Breaking it up into smaller reusable parts would help if that is possible

.github/workflows/create-hotfix-branch.yaml
```diff
@@ -0,0 +1,80 @@
...
+
+on:
+  workflow_dispatch:
+    inputs:
+      version:
+        description: 'Hotfix Branch Version'
+        required: false
+
+env:
+  artifactId: gaffer2
```
r32575, line 11: As it is an environment variable should it be `ARTIFACT_ID`

.github/workflows/create-release-branch.yaml
```diff
@@ -0,0 +1,80 @@
...
+
+on:
+  workflow_dispatch:
+    inputs:
+      version:
+        description: 'Release Branch Version'
+        required: false
+
+env:
+  artifactId: gaffer2
```
r32575, line 11: As it is an environment variable should it be `ARTIFACT_ID`

.github/workflows/release.yaml
```diff
@@ -2,8 +2,8 @@ name: Full Release
 
 on:
   milestone:
-   types:
-   - closed
+    types:
+    - closed
 
 env:
   artifactId: gaffer2
```
r32575, line 9: As it is an environment variable should it be `ARTIFACT_ID`

r32575 (commented):
Had a quick first look and found a couple of things. Not familiar with actions so will have to read up on it

.github/workflows/create-alpha-release-branch.yaml
t92549: `artifactId` is used because it is used in maven release process for naming the artifact: http://maven.apache.org/guides/mini/guide-naming-conventions.html

.github/workflows/create-alpha-release-branch.yaml
t92549: I agree it is weird but it is the official suggested regex on semver.org: https://semver.org/#is-there-a-suggested-regular-expression-regex-to-check-a-semver-string

.github/workflows/create-alpha-release-branch.yaml
r32575: So is it used externally in a maven process somewhere? If not and it is just used by actions then shouldn't you use the actions env precedence which is capitals with underscores?

.github/workflows/create-alpha-release-branch.yaml
GCHQDeveloper314: I also think this is weird and overly complex. Most of that complexity seems to be to prevent versions with leading zeros, like `01.023.03`. It also has capturing groups which we don't need. We could use something like `^\d+\.\d+\.\d+(?:-\w+)?$` if we just want a simple check. Maybe leave as it and add that link to explain why it's weird?

.github/workflows/create-alpha-release-branch.yaml
t92549: When there is a maven release (in the release workflow) it is picked up by maven to give the artifact name. Therefore it just seems right to use that across all workflows rather than change it from `ARTIFACT_ID` in create-release-branch to `artifactId` in release

.github/workflows/create-alpha-release-branch.yaml
t92549: Yeah perhaps a link. If there is a suggested would be a shame not to use it.

.github/workflows/create-alpha-release-branch.yaml
```diff
@@ -0,0 +1,80 @@
...
+    - name: Update versions
+      run: |
+        mvn versions:set -DnewVersion=$RELEASE_VERSION -DgenerateBackupPoms=false
+        sed -i'' -e "s/^gaffer.version=.*/gaffer.version=$RELEASE_VERSION/" rest-api/common-rest/src/main/resources/version.properties
+
+    - name: Push to release branch
+      run: |
+        git checkout -b $BRANCH_NAME
+        git commit -a -m "prepare release $artifactId-$RELEASE_VERSION"
+        git push --set-upstream origin $BRANCH_NAME
```
r32575, line 80: Seems that a lot of the code is duplicated. If storing env variables in the in the central cache `GITHUB_ENV` keeps them for the entirety of the action (with multiple workflows) then you could have an alpha, hot-fix etc workflows which store the values such as `BRANCH_NAME` and `ARTIFACT_ID` in the cache using:
```echo "BRANCH=..." >> $GITHUB_ENV```
Then call a generic workflow where the values are retrieved from the cache.

File: .github/workflows/create-alpha-release-branch.yaml (hunk 1 above)
Comment (r32575) [reply .github/workflows/create-alpha-release-branch.yaml:11]: The name is reserved in maven as it actually is used to indicate the artifact in the pom. This is an totally different language with different rules

Comment (t92549) [reply .github/workflows/create-alpha-release-branch.yaml:11]: When `mvn deploy` is run in the release it grabs `artifactId` from an environment variable and uses it to set the name of the artifact in the pom. https://github.com/gchq/Gaffer/blob/0117e3faf758d87e339bcb2513028b485f03fe1a/.github/workflows/release.yaml#L227

The line in question sets that variable as an environment variable. The reason it is used in all 4 workflows is consistency.

File: .github/workflows/create-alpha-release-branch.yaml (hunk 6 above)
Comment (t92549) [reply .github/workflows/create-alpha-release-branch.yaml:80]: When you define an environment variable within a job, then that variable can only be used within that job. https://docs.github.com/en/actions/learn-github-actions/environment-variables

This is because each job is run on a different machine. The only variables that can be used across all jobs in a workflow are ones that are predefined at the top of the workflow.

To export a variable to dependent jobs you have to use outputs like this: https://github.com/gchq/Gaffer/blob/0117e3faf758d87e339bcb2513028b485f03fe1a/.github/workflows/release.yaml#L41-L46

And use them like this: https://github.com/gchq/Gaffer/blob/0117e3faf758d87e339bcb2513028b485f03fe1a/.github/workflows/release.yaml#L210

File: .github/workflows/create-alpha-release-branch.yaml (hunk 1 above)
Comment (r32575) [reply .github/workflows/create-alpha-release-branch.yaml:11]: From what I can tell `artifactId` is an actions env variable that is injected:
```git commit -a -m "prepare release $artifactId-$RELEASE_VERSION"```
If it already existed before then it should have been flagged up at a previous code review.

File: .github/workflows/create-alpha-release-branch.yaml (hunk 6 above)
Comment (r32575) [reply .github/workflows/create-alpha-release-branch.yaml:80]: Thats a shame...

File: .github/workflows/create-alpha-release-branch.yaml (hunk 1 above)
Comment (t92549) [reply .github/workflows/create-alpha-release-branch.yaml:11]: It did already exist, GitHub actions adds it as an environment variable to the machine that is running the job, so that git commit line correctly uses it, but `mvn deploy` also uses it as it looks for the `artifactId` environment variable.

File: .github/workflows/create-alpha-release-branch.yaml (hunk 6 above)
Comment (t92549) [reply .github/workflows/create-alpha-release-branch.yaml:80]: It may seem like duplication but state and conditionals aren't really a thing you can do easily with these workflows, otherwise it would just be a simple case of `if X: ..., else if Y: ..., else if Z: ...`
Instead my approach was to split the "conditionals" into 3 workflows and then the release workflow handles the common elements.

File: .github/workflows/release.yaml [hunk 7]
```diff
@@ -76,18 +143,21 @@ jobs:
...
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
       with:
         tag_name: ${{ env.artifactId }}-${{ env.RELEASE_VERSION }}
-        release_name: Gaffer ${{ env.RELEASE_VERSION }}
+        name: Gaffer ${{ env.RELEASE_VERSION }}
         body_path: release_file.md
+        prerelease: ${{ startsWith( needs.create-release-tag.outputs.branch_name , 'alpha') }}
 
+  # Skip if alpha release
   generate-javadoc:
```
GCHQDeveloper314, line 157: For an alpha, this job clones master, sets some credentials, installs the JDK, sets version and then does nothing.
Instead of doing the prerequisites and skipping only the build & upload of the Javadoc, for alphas the whole job can be skipped by moving the `if` to cover the whole job and not just one step. [Described here.](https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#jobsjob_idif)

t92549: Okay nice, I didn't know that, will amend now on all 4 PRs! I originally did it as I didn't want to keep skipping steps but didn't know you can just skip jobs!

.github/workflows/create-alpha-release-branch.yaml
r32575: Sorry, still not seeing where `artifactId` is actually being used other than with a `$` which substitutes the env variable...

.github/workflows/create-alpha-release-branch.yaml
r32575: Seems you can have reusable workflows with inputs:
[https://docs.github.com/en/actions/learn-github-actions/reusing-workflows](url)

t92549: This can be done but with the create release branch action there would still be 4 conditionals with 3 distinct choices each. This means you have two choices: one action riddled with lots of ugly versions of if, else if, and else within GitHub actions syntax, or one action where the value from these conditionals are set by the user, but it seems easier for the user to just choose the correct workflow from 3 options rather than have to define the branch to merge from, name prefix of the release branch, correct regex and semver release increment every release.

r32575: There are just a handful of variables you pass in... The files are almost identical other than the regex, branch type and one conditional

t92549: There are 4 variables to pass in (branch to merge from, name prefix of the release branch, correct regex and semver release increment ). I think it makes the action harder to use if someone has to define those every time as options.
The other option like I said is you just input one of the 3 types (standard, hotfix or alpha) but then there has to be 4 conditionals at the top of the action that programmatically set the 4 variables to one of the 3 options which is very ugly. For example, this is how I set one variable with 2 options: https://github.com/gchq/Gaffer/blob/dd084ad45339d5b49aa1b71dd0de67a4281a724b/.github/workflows/release.yaml#L52-L58

I just think it would be ugly to have 12 of these at the start of the action and would rather have 3 different actions.

.github/workflows/create-alpha-release-branch.yaml
t92549: Perhaps you are right, I presumed because it was like this before, `artifactId` was used by maven in the deploy stage but I can't find any evidence that it is. I can change this to `ARTIFACT_ID` and if something goes weird and breaks on the first release run then change them all back I suppose.

t92549: Even though it is correct for environment variables to be all caps, `artifactId` still looks better to me as it mirrors the pom.

.github/workflows/create-alpha-release-branch.yaml
r32575: I just had:
```
on:
  workflow_call:
    inputs:
      regex:
        required: true
        type: string
      branch:
        required: true
        type: string
        # release: "release"
        # alpha: "alpha-release"
        # hotfix: "hotfix"
      branch_type:
        required: true
        type: string
        # release: "develop"
        # alpha: "alpha-v2"
        # hotfix: "master"
      conditional_regex:
        required: false
        type: string
        # alpha: "alpha"
      version_type:
        required: true
        type: string
        # release: "minor"
        # hotfix: "patch"
        # alpha: "prerelease"
    secrets:
      token:
        required: true
        # ADMIN_GITHUB_TOKEN
```
And included extra step for the `conditional_regex`:
```
- name: Fail if doesn't contain conditional regex
      run: |
        if [[ ! -z ${{ input.conditional_regex }} && ! ${{ env.RELEASE_VERSION }} =~ ${{ input.conditional_regex }} ]] 
        then
            exit 1
        fi
```
But it's only a suggestion. Just saves having 3 near identical `create-.*.yaml` files.

t92549: The lines that are commented out don't do anything, what happens if I submit a job with branch set to "develop"? And if you plan to let a user have to input all of the correctly manually each time that is pretty unusable. Or another workflow can trigger this, but then that workflow will be an ugly list of those conditionals that I showed above. At the end of the day you just can't do this type of thing with GitHub actions and make it pretty sadly.

r32575: The comments were just what I would enter. In the `create-release-branch.yaml` file you would have:
```
on:
  workflow_dispatch:
    inputs:
      version:
        description: 'Release Branch Version'
        required: false

env:
  artifactId: gaffer2
  MAVEN_OPTS: -Dmaven.wagon.http.retryHandler.count=3 -Dmaven.wagon.httpconnectionManager.ttlSeconds=25

jobs:
  create-release-branch:
    uses: .github/workflows/create-branch.yml@v1
    with:
      inputs:
        regex: '<regex>'
        branch: 'release'
        branch_type: 'develop'
        version_type: 'minor'
    secrets:
      token: ${{ secrets.ADMIN_GITHUB_TOKEN }}
```
And do the same for `alpha` and `hotfix`

t92549: So you think there should be 4 files instead of 3? I think that's adding more complexity for the sake of reducing number of lines which I don't see as an issue.

r32575: The `create-branch.yaml` has the same code as `create-alpha-release.yaml` but with the inputs as above and values substituted for for the input variables along with the the extra step to do the regex check if a value is provided (for the other types you don't include an `conditional_regex` so it just continues).

t92549: I suppose it would be easier to maintain but harder to understand

r32575: Whatever you think is best. Code reuse means you don't need to alter multiple files if you make a change to the create logic...

t92549: Yeah I think it will be easier to maintain for those that understand how it works but I worry that it is becoming increasingly complex and using more niche features of GitHub actions that are harder to understand. Maybe cleaner though, will have a think and ask some more people maybe

t92549: So I think a clean way to reduce code duplication might just be to put it all in one file with a single input like "branch type" which can either be left blank (standard), or set to hotfix or alpha. Then there can be one big ugly block setting all the variables, then it makes the branch. This reduces files from 4 to 1 and uses less complex features.

Having said that, I think for now we should just use what I have made because it will take a long time to rejig and retest all the branch types on all 4 repos. We will have to rejig the actions in the future anyway once Gaffer 2 is released to change the alpha branch option to a Gaffer 1 support branch. If we still wish to reduce it to 1 file and 1 workflow then, it can be done. What do you think? @r32575

p3430233 (approved):

PR #2572, opened 2022-01-17
gh-2571: Fixed matchedVertex behaviour and added new ITs

### LINKED ISSUE #2571: Improve and test Accumulo matched vertex behaviour
After removing `SeedMatching` in #2552, we decided that the `includeMatchedVertex` flag should be set depending on the operation. It should only be true when there are only EntityIds in the input. As well as this, there should be tests to document the fact that sometimes one Edge can be returned twice depending on how it matches with an Entity and Edge input.

https://github.com/gchq/Gaffer/blob/ab0b66c11f99d30d6c9b9f004e09f6166e206188/store-implementation/accumulo-store/src/main/java/uk/gov/gchq/gaffer/accumulostore/retriever/impl/AccumuloElementsRetriever.java#L31

t92549:
# Related Issue

- Resolve #2571

### DIFF
--- a/store-implementation/accumulo-store/src/test/java/uk/gov/gchq/gaffer/accumulostore/integration/AccumuloMatchedVertexIT.java
+++ b/store-implementation/accumulo-store/src/test/java/uk/gov/gchq/gaffer/accumulostore/integration/AccumuloMatchedVertexIT.java
@@ -257,7 +257,7 @@ public class AccumuloMatchedVertexIT extends StandaloneIT {
 
     @ParameterizedTest
     @MethodSource("getEdgeVariants")
-    public void shouldTest9(Edge edge) throws OperationException {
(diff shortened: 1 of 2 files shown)


store-implementation/accumulo-store/src/test/java/uk/gov/gchq/gaffer/accumulostore/integration/AccumuloMatchedVertexIT.java
```diff
@@ -0,0 +1,337 @@
...
+        final List<Edge> edgeResults = getEdgeResults(results);
+
+        // Then
+        assertThat(results).hasSize(1);
+        assertThat(edgeResults).hasSize(0);
+    }
+
+    @ParameterizedTest
+    @MethodSource("getEdgeVariants")
+    public void shouldTest9(Edge edge) throws OperationException {
```
GCHQDev404, line 260: name?

GCHQDev404 (approved):

PR #2577, opened 2022-01-25
gh-2560 GetTraitsHandler is initialised with store traits

### LINKED ISSUE #2560: Make GetTraits Operation configurable per store
This will mean it no longer relies on the deprecated getTraits function: https://github.com/gchq/Gaffer/blob/296c453ed415d38d7ccd57324089242a2c591a31/core/store/src/main/java/uk/gov/gchq/gaffer/store/Store.java#L359-L360

Instead the operation should be configured per store and setup so that each store's handler relates to the traits of that store.

p3430233:
# Related Issue

- Resolve #2560


t92549:
Closed as this needs to be merged into https://github.com/gchq/Gaffer/pull/2564

PR #2582, opened 2022-01-31
gh-2560-getTraitsOperation

GCHQDev404:
Removing redundant adding of GetTraitHandler from addAdditionalOperationHandlers() because it has its own method getGetTraitsHandler().


t92549 (approved):
Good catch! I think we did this but it got lost somewhere

PR #2583, opened 2022-02-08
Gh 2553 add closeable classes from koryphe and remove closeable Iterable, iterator and wrapper

r32575:
Closeable iterator and iterable have been removed from `Koryphe` for v2. This issue removes closeable iterator, iterable and wrapper from `Gaffer`. Other changes include:

- updating the testing framework to `AssertJ`
- fixing bugs discovered


store-implementation/accumulo-store/src/test/java/uk/gov/gchq/gaffer/accumulostore/operation/handler/GetElementsBetweenSetsHandlerTest.java
```diff
@@ -158,15 +158,15 @@ public void shouldReturnElementsNoSummarisationGaffer1Store() throws OperationEx
     private void shouldReturnElementsNoSummarisation(final AccumuloStore store) throws OperationException {
         final GetElementsBetweenSets op = new GetElementsBetweenSets.Builder().input(inputA).inputB(inputB).view(defaultView).build();
         final GetElementsBetweenSetsHandler handler = new GetElementsBetweenSetsHandler();
-        final CloseableIterable<? extends Element> elements = handler.doOperation(op, user, store);
+        final Iterable<? extends Element> elements = handler.doOperation(op, user, store);
```
r32575, line 161: Add a try-finally to close the `Iterable`

core/common-util/src/main/java/uk/gov/gchq/gaffer/commonutil/iterable/ChainedIterable.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016-2020 Crown Copyright
+ * Copyright 2022-2026 Crown Copyright
```
GCHQDeveloper314, line 2: Something went wrong when updating this!

core/common-util/src/main/java/uk/gov/gchq/gaffer/commonutil/iterable/ConsumableBlockingQueue.java
```diff
@@ -24,6 +24,7 @@
 
 import java.util.Iterator;
 import java.util.NoSuchElementException;
+import java.util.Objects;
```
GCHQDeveloper314, line 27: Could be static? Also seems like using this doesn't provide any benefit, I'm not sure we should be adding it if the code is otherwise ok, [this isn't it's intended use case](https://docs.oracle.com/javase/8/docs/api/java/util/Objects.html#isNull-java.lang.Object-).

GCHQDeveloper314 (commented):
The diff for this PR contains changes to lots of files which have already been merged (removal of hbase and parquet) I think this is because git still sees this as a change for some reason. Rebasing the branch will fix that.

core/common-util/src/main/java/uk/gov/gchq/gaffer/commonutil/iterable/ConsumableBlockingQueue.java
r32575: I personally find it easier to read than `null == variable` or `null != variable`. I find the ordering of the conditionals stating with `null` difficult to read as well.

core/common-util/src/main/java/uk/gov/gchq/gaffer/commonutil/iterable/ChainedIterable.java
r32575: Not sure I understand... The file has been altered so the new version has been updated with the new copyright period...

core/common-util/src/main/java/uk/gov/gchq/gaffer/commonutil/iterable/ConsumableBlockingQueue.java
t92549: I also prefer `null == variable` but it's just a preference thing and there is no right answer. However, the precedent in Gaffer is to use `null == variable` so perhaps we shouldn't have an endless back and forth and edit pre-existing uses, but instead just write new uses how we prefer. Will resolve this conversation though as it doesn't matter too much.

core/store/src/main/java/uk/gov/gchq/gaffer/store/Store.java
```diff
@@ -257,7 +260,7 @@ public static Store createStore(final String graphId, final Schema schema, final
                     .asSubclass(Store.class)
                     .newInstance();
         } catch (final InstantiationException | IllegalAccessException | ClassNotFoundException e) {
-            throw new IllegalArgumentException("Could not create store of type: " + storeClass, e);
+            throw new IllegalArgumentException(String.format("Could not create store of type: %s", storeClass), e);
```
t92549, line 263: I do like this change but it does make it harder to review the actual changes when there are so many unrelated formatting changes

core/store/src/test/java/uk/gov/gchq/gaffer/store/StoreTest.java
```diff
@@ -197,46 +203,36 @@ public void setup() {
...
-                .edge(TestGroups.EDGE, new SchemaEdgeDefinition.Builder()
-                        .source("string")
-                        .destination("string")
-                        .directed("true")
-                        .property(TestPropertyNames.PROP_1, "string")
-                        .property(TestPropertyNames.PROP_2, "string")
-                        .build())
-                .edge(TestGroups.EDGE_2, new SchemaEdgeDefinition.Builder()
-                        .source("string")
+        lenient().when(schemaOptimiser.optimise(any(Schema.class), any(Boolean.class))).then(returnsFirstArg());
```
t92549, line 207: Again, why lenient?

rest-api/core-rest/src/main/java/uk/gov/gchq/gaffer/rest/service/v1/OperationService.java
```diff
@@ -106,27 +106,27 @@ public Object execute(final Operation operation) {
     }
 
     @Override
-    public CloseableIterable<Object> generateObjects(final GenerateObjects<Object> operation) {
+    public Iterable<Object> generateObjects(final GenerateObjects<Object> operation) {
```
t92549, line 109: Why were these CloseableIterables before? Is there any type to replace this return type of a simple Closeable Iterable?

store-implementation/accumulo-store/src/main/java/uk/gov/gchq/gaffer/accumulostore/operation/handler/GetAllElementsHandler.java
```diff
@@ -29,14 +29,18 @@
 import uk.gov.gchq.gaffer.store.operation.handler.OutputOperationHandler;
 import uk.gov.gchq.gaffer.user.User;
 
-public class GetAllElementsHandler implements OutputOperationHandler<GetAllElements, CloseableIterable<? extends Element>> {
+public class GetAllElementsHandler implements OutputOperationHandler<GetAllElements, Iterable<? extends Element>> {
+
     @Override
-    public CloseableIterable<? extends Element> doOperation(final GetAllElements operation, final Context context, final Store store)
```
t92549, line 34: Same question, why was this returning CloseableIterable before and is there a return type that is Closeable? Or does this make no difference?

t92549 (commented):
Just a bunch of questions. My main comment is that, although I like most of the reformatting you have done (but not all of it), I think it makes the PR way harder to actually review when most of it is formatting, it almost hides the actual changes.

core/store/src/test/java/uk/gov/gchq/gaffer/store/StoreTest.java
r32575: With `when` it errors if the mock was declared but not used as was the case with this. If you add lenient it does not error if a particular mock was not used.

core/common-util/src/test/java/uk/gov/gchq/gaffer/commonutil/iterable/LimitedIterableTest.java
```diff
@@ -27,20 +27,22 @@
...
 
-        final CloseableIterable<Integer> limitedValues = new LimitedCloseableIterable<>(values, start, end);
-
-        assertThat(limitedValues).containsExactlyElementsOf(values.subList(start, end));
+        LimitedIterable<Integer> limitedValues = null;
+        try {
+            limitedValues = new LimitedIterable<>(values, start, end);
+            assertThat(limitedValues).containsExactlyElementsOf(values.subList(start, end));
+        } finally {
+            CloseableUtil.close(limitedValues);
```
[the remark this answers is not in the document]
r32575: Not sure whether it was required since the change or it was down to good/bad habit

core/common-util/src/main/java/uk/gov/gchq/gaffer/commonutil/iterable/EmptyIterable.java
```diff
@@ -19,23 +19,26 @@
...
- * {@link EmptyCloseableIterator}, and contains no objects.
+ * An {@code EmptyIterable} is an {@link java.io.Closeable} {@link java.lang.Iterable}
+ * which is backed by a {@link EmptyIterator}, and contains no objects.
  *
  * This is useful when a {@link java.lang.Iterable} is required, but there is no data present.
  *
  * @param <T> the type of items in the iterable.
  */
-public class EmptyClosableIterable<T> implements CloseableIterable<T> {
+public class EmptyIterable<T> implements Closeable, Iterable<T> {
```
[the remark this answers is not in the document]
r32575: Possibly... As you can't return both types and `Iterable` is preferred to `Closeable` I changed it

core/store/src/test/java/uk/gov/gchq/gaffer/store/StoreTest.java
t92549: I guess this is fine for migrating to new mockito but probably means the tests should be refactored

core/common-util/src/main/java/uk/gov/gchq/gaffer/commonutil/iterable/EmptyIterable.java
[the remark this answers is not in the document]
r32575: Looking at the `Iterable` `SuppliedIterable`, it implemented `CloseableIterable` but was still kept it's `Iterable` name. Which precedence do we want to follow?

store-implementation/accumulo-store/src/main/java/uk/gov/gchq/gaffer/accumulostore/operation/handler/GetAllElementsHandler.java
r32575: Not entirely sure what you mean. Originally returned `CloseableIterable` but this was removed. There is not such an interface in the Java libraries that combines `Iterable` and `Closeable` and as we want to return an iterable of results 'Itearble` was chosen as the return type rather than `Closeable`.

core/common-util/src/main/java/uk/gov/gchq/gaffer/commonutil/iterable/EmptyIterable.java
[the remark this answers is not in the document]
t92549: You're right it seems most don't have this, just wondering if it might make sense to keep the ones that were called Closeable before as Closeable. Will make migration easier but might also be worth adding breaking changes now?

store-implementation/accumulo-store/src/main/java/uk/gov/gchq/gaffer/accumulostore/operation/handler/GetAllElementsHandler.java
t92549: I am just wondering if this changes functionality as the result is no longer Closeable?

r32575: That's what I was trying to get across before. The client using Java would have to use the `try-finally` block with the `CloseableUtil` check than the `try` before the change. But returning `Closeable` would mean you would have to check if the `Object` was an `Iterable` or not.

t92549: So the only thing this did before was stop you having to use the `CloseableUtil` to close it? No functionality differences?

store-implementation/federated-store/src/test/java/uk/gov/gchq/gaffer/federatedstore/FederatedGraphStorageTraitsTest.java
```diff
@@ -0,0 +1,410 @@
...
+import static uk.gov.gchq.gaffer.user.StoreUser.AUTH_1;
+import static uk.gov.gchq.gaffer.user.StoreUser.AUTH_USER_ID;
+import static uk.gov.gchq.gaffer.user.StoreUser.TEST_USER_ID;
+import static uk.gov.gchq.gaffer.user.StoreUser.UNUSED_AUTH_STRING;
+import static uk.gov.gchq.gaffer.user.StoreUser.authUser;
+import static uk.gov.gchq.gaffer.user.StoreUser.blankUser;
+import static uk.gov.gchq.gaffer.user.StoreUser.nullUser;
+import static uk.gov.gchq.gaffer.user.StoreUser.testUser;
+
+public class FederatedGraphStorageTraitsTest {
```
GCHQDev404, line 57: This class was deleted.

GCHQDev404 (changes requested):
class deleted.

core/graph/src/test/java/uk/gov/gchq/gaffer/graph/hook/FunctionAuthoriserTest.java
```diff
@@ -21,7 +21,7 @@
 
 import uk.gov.gchq.gaffer.commonutil.JsonAssert;
 import uk.gov.gchq.gaffer.commonutil.exception.UnauthorisedException;
-import uk.gov.gchq.gaffer.commonutil.iterable.CloseableIterable;
+
```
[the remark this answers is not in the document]
r32575: Removed all occurences

core/graph/src/test/java/uk/gov/gchq/gaffer/integration/graph/SchemaHidingIT.java
```diff
@@ -23,7 +23,7 @@
 import uk.gov.gchq.gaffer.commonutil.StreamUtil;
 import uk.gov.gchq.gaffer.commonutil.TestGroups;
 import uk.gov.gchq.gaffer.commonutil.TestPropertyNames;
-import uk.gov.gchq.gaffer.commonutil.iterable.CloseableIterable;
+
```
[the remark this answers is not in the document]
r32575: Removed all occurrences

core/common-util/src/main/java/uk/gov/gchq/gaffer/commonutil/iterable/EmptyIterable.java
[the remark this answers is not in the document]
t92549: After a discussion, we think that renaming the classes here to keep them consistent is a good place to start, as it's an alpha.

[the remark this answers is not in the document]
t92549: Worse case, if they need to be changed back for compatibility, they can be in a later alpha.

rest-api/core-rest/src/main/java/uk/gov/gchq/gaffer/rest/service/v1/OperationService.java
t92549: Resolved (see comment below)

store-implementation/accumulo-store/src/main/java/uk/gov/gchq/gaffer/accumulostore/operation/handler/GetAllElementsHandler.java
r32575: Yeah, that is the only difference. The java user will have to include a `try-finally` block with the `CloseableUtil.close()` check if they don't want a memory leak.

core/common-util/src/test/java/uk/gov/gchq/gaffer/commonutil/iterable/LimitedIterableTest.java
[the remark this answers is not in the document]
r32575: Removed the `try-finally` blocks from the tests.

GCHQDev404 (commented):
class deleted.

core/common-util/src/main/java/uk/gov/gchq/gaffer/commonutil/iterable/ChainedIterable.java
```diff
@@ -16,97 +16,97 @@
...
  * @param <T> the type of items in the iterable.
  */
-public class ChainedIterable<T> implements CloseableIterable<T> {
-    private final Iterable<T>[] itrs;
-    private final int n;
+public class ChainedIterable<T> implements Closeable, Iterable<T> {
+
+    private final Iterable<? extends Iterable<? extends T>> iterables;
+
+    public ChainedIterable(final Iterable<? extends T>... iterables) {
```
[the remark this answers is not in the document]
t92549: Seems fine, it did this before right?

store-implementation/accumulo-store/src/main/java/uk/gov/gchq/gaffer/accumulostore/operation/handler/GetAllElementsHandler.java
t92549: But they should have done this anyway, so no change really other than name.

core/common-util/src/test/java/uk/gov/gchq/gaffer/commonutil/iterable/LimitedIterableTest.java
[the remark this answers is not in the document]
t92549: This makes sense to me now, tests don't need to finally call close because they are short lived and exceptions shouldn't happen.
In operational code, if an exception might happen that would potentially leave a Closeable un-closed, they should be closed in a finally block to prevent a memory leak.
This is the same as how it should have been done before.
The only difference is, now you don't sometimes get a `CloseableIterable` type returned (and therefore know you must finally close it), instead you should ensure you close the `Iterable` if it may have come from an AccumuloStore and be Closeable. To do this, you use the CloseableUtil in Gaffer.

PR #2584, opened 2022-02-09
Gh 2562 GetElementsWithinSet not allowing for the exclusion of properties - Gaffer v2

### LINKED ISSUE #2562: GetElementsWithinSet not allowing for the exclusion of properties
The GetElementsWithinSet operation is not handling properties as expected. The expected behaviour is that this operation would handle properties like the GetElements operation whereby
including an empty list when setting properties  should not return any properties 
using excludeProperties within the view should exclude the specific properties set.
 
Instead, GetElementsWithinSets returns all properties in the case of both a & b.

r32575:
GetElementsWithinSet was not filtering view properties. The bug has been identified and fixed. Further tests have been added to check that the fix is functioning correctly.

# Related Issue

- Resolve #2562


t92549:
Closing as https://github.com/gchq/Gaffer/pull/2597 can be cherry picked into v2-alpha (with new ticket and PR that doesn't conflict)

PR #2605, opened 2022-03-10
gh-2556: Add HBaseStore deprecated tag, javadocs and note to README

### LINKED ISSUE #2556: Deprecate HBase.
The unsupported HBase store appears to be unused. https://github.com/gchq/Gaffer/issues/2367 elicited one response which concluded that the user would switch to Accumulo if they continued to use Gaffer.

https://github.com/gchq/Gaffer/discussions/2555 has announced that HBase will not be in Gaffer 2.0, so follow through on the process and deprecate its use in Gaffer 1 so we can legitimately remove it.

t92549:
# Related Issue

- Resolve #2556

### DIFF
--- a/store-implementation/hbase-store/README.md
+++ b/store-implementation/hbase-store/README.md
@@ -14,6 +14,11 @@ limitations under the License.
 
 The master copy of this page is the README in the hbase-store module.
 
+Deprecated
+===================
+The HBase Store has been [deprecated](https://github.com/gchq/Gaffer/issues/2556). It will be deleted during the Gaffer 2 development process and Gaffer v2.0.0 will not have a HBase Store implementation included with it.  
+
+
 HBase Store
 ===================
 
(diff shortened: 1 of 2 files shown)


GCHQDeveloper314 (approved):
In the title of this PR/issue, what's the difference between adding deprecated tag and Javadocs change? Seems to be the same thing.

t92549:
@GCHQDeveloper314 I added the `@Deprecated` tag above the class (which is used in IDE I believe), as well as some `@deprecated` javadoc stuff:
https://github.com/gchq/Gaffer/blob/937d6df59953e7403adfd36be0b6d65f37042024/store-implementation/hbase-store/src/main/java/uk/gov/gchq/gaffer/hbasestore/HBaseStore.java#L95-L99

r32575 (approved):

PR #2606, opened 2022-03-10
gh-2604: Deprecates Parquet

### LINKED ISSUE #2604: Deprecate Parquet store
As described in https://github.com/gchq/Gaffer/discussions/2557, the Parquet store is unsupported and doesn't seem to be used. We will be removing it as part of Gaffer 2 so it should be marked as Deprecated.
In the future if there was demand, a parquet store could be implemented in Gaffer 2.

t92549:
# Related Issue

- Resolve #2604

### DIFF
--- a/store-implementation/parquet-store/README.md
+++ b/store-implementation/parquet-store/README.md
@@ -12,6 +12,9 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 
+# Deprecated
+The Parquet Store has been [deprecated](https://github.com/gchq/Gaffer/issues/2604). It will be deleted during the Gaffer 2 development process and Gaffer v2.0.0 will not have a Parquet Store implementation included with it.  
+
 # Parquet Store
 
 Contents:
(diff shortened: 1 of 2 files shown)


store-implementation/parquet-store/README.md
```diff
@@ -12,6 +12,9 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 
+# Deprecated
+The Parquet Store has been [deprecated](https://github.com/gchq/Gaffer/issues/2604). It will be deleted during the Gaffer 2 development process and Gaffer v2.0.0 will not have a Parquet Store implementation included with it.
```
GCHQDeveloper314, line 16: Technically it's already been deleted during the development process. Correct to say it will not be in Gaffer 2, as there is not yet a non development release of that.

GCHQDeveloper314 (approved):

store-implementation/parquet-store/README.md
t92549: Yeah it's weird, kinda a ret-con but perhaps it would be more correct to say "has been deleted"

r32575 (approved):

PR #2607, opened 2022-03-10
gh-2600: Update v2 README and docs links

### LINKED ISSUE #2600: Update Gaffer 2 README
The Gaffer Readme used for alpha 1 still documents gaffer-tools artefacts we have removed in alpha1 - jar-shader, slider & schema-builder. (but not the analytic ui).

t92549:
# Related Issue

- Resolve #2600

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-Copyright 2016-2020 Crown Copyright
+Copyright 2016-2022 Crown Copyright
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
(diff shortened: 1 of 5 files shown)


t92549:
I removed semi-colons at the end of bullet points because I thought it looked weird but that's just a stylistic decision that is subjective.

GCHQDeveloper314 (approved):
I was going to mention the change of capitalisation for the doc and tools links, then I realised this was because those repositories are not capitalised. All looks good to me, I agree with removing the semicolons.

t92549:
@GCHQDeveloper314 There are a lot of links to things like the getting-started docs which still point to v1docs, I think for these links we should just replace them as and when the v2 versions of these docs are added. Sound good to you?

GCHQDeveloper314:
Yes I agree with that. As part of creating the new docs we should review the READMEs with a view to moving the majority of their content over. This will make maintenance easier in future.

PR #2609, opened 2022-03-11
Gh-2602: Replace Log4j with Reload4j

### LINKED ISSUE #2602: Port Log4J replacement to Gaffer 1.*
As part of Gaffer 2.0 development a replacement for the current dependency on the insecure Log4J has been developed. Backfill this into Gaffer 1.

GCHQDeveloper314:
Log4j has been replaced with Reload4j. Maven enforcer plugin is also introduced to prevent future classpath problems if new or upgraded dependencies try to pull in Log4j or other related conflicting dependencies.

# Related Issue

- Resolve #2602

### DIFF
--- a/example/basic/basic-model/pom.xml
+++ b/example/basic/basic-model/pom.xml
@@ -67,6 +67,17 @@
             <artifactId>accumulo-minicluster</artifactId>
             <version>${accumulo.version}</version>
             <scope>test</scope>
+            <exclusions>
+                <!-- Conflicts with Reload4j -->
+                <exclusion>
+                    <groupId>org.slf4j</groupId>
+                    <artifactId>slf4j-log4j12</artifactId>
+                </exclusion>
+                <exclusion>
+                    <groupId>log4j</groupId>
+                    <artifactId>log4j</artifactId>
+                </exclusion>
+            </exclusions>
         </dependency>
         <dependency>
             <groupId>org.apache.hadoop</groupId>
@@ -106,6 +117,15 @@
                 <exclusion>
                     <groupId>org.apache.avro</groupId>
                     <artifactId>avro-ipc</artifactId>
+                </exclusion>
+                <!-- Conflicts with Reload4j -->
+                <exclusion>
+                    <groupId>org.slf4j</groupId>
+                    <artifactId>slf4j-log4j12</artifactId>
+                </exclusion>
+                <exclusion>
+                    <groupId>log4j</groupId>
+                    <artifactId>log4j</artifactId>
                 </exclusion>
             </exclusions>
         </dependency>
--- a/pom.xml
+++ b/pom.xml
@@ -56,7 +56,7 @@
         <junit.platform.suite.version>1.6.2</junit.platform.suite.version>
         <mockito.version>3.3.3</mockito.version>
         <failsafe.version>3.0.0-M5</failsafe.version>
-        <slf4j.api.version>1.7.25</slf4j.api.version>
+        <slf4j.api.version>1.7.36</slf4j.api.version>
 
         <koryphe.version>1.14.0</koryphe.version>
         <accumulo.version>1.9.3</accumulo.version>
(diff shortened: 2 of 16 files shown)


pom.xml
```diff
@@ -909,6 +958,30 @@
...
+                            <goal>enforce</goal>
+                        </goals>
+                        <configuration>
+                            <rules>
+                                <bannedDependencies>
+                                    <excludes>
+                                        <exclude>*log4j*:*log4j*</exclude>
+                                        <exclude>org.slf4j:slf4j-log4j12</exclude>
+                                    </excludes>
+                                    <message>Log4j conflicts with reload4j, add an exclusion</message>
```
t92549, line 978: 👍

pom.xml
```diff
@@ -233,9 +255,9 @@
                 <version>${javax-activation.version}</version>
             </dependency>
             <dependency>
-                <groupId>log4j</groupId>
-                <artifactId>log4j</artifactId>
-                <version>${log4j.version}</version>
+                <groupId>ch.qos.reload4j</groupId>
+                <artifactId>reload4j</artifactId>
```
t92549, line 259: Sounds silly, but as there are no tests that check the logging works as expected, have you run locally and verified the logs are still being written?

example/basic/basic-model/pom.xml
```diff
@@ -67,6 +67,17 @@
             <artifactId>accumulo-minicluster</artifactId>
             <version>${accumulo.version}</version>
             <scope>test</scope>
+            <exclusions>
+                <!-- Conflicts with Reload4j -->
```
t92549, line 71: Am I right in thinking that, not only will this stop dependency conflicts, but remove our log4j vulnerability - not only because Gaffer won't use it directly but it won't be using it indirectly through its dependencies?

If so, then will accumulo still be writing logs? And they will just be written by Reload4j rather than log4j?

pom.xml
t92549, line 259: Another question: are all log4j.xml property files also compatible with reload4j? Is it possible that my logging set up won't work now because of this change?

t92549 (commented):
Just some questions

example/basic/basic-model/pom.xml
GCHQDeveloper314: So far as I understand, that is correct. These changes cause log4j to be excluded from all transitive dependencies. Therefore the log4j JAR will not be on the classpath when any part of Gaffer is run. Any code in external dependencies (e.g. Accumulo) which makes calls to log4j or slf4j will instead be using Reload4j. Reload4j is a continuation of the log4j codebase and is compatible with code complied against log4j.

t92549: I think it is key that it works with code that calls both slf4j AND log4j directly, because I think some of the dependencies (hdfs or Accumulo, I can't remember) still call log4j directly.

pom.xml
t92549: Just a thought that maybe we should run a manual before / after and try to verify the outputs are the same or near enough, because we don't have automated logging testing

r32575 (approved):

example/basic/basic-model/pom.xml
GCHQDeveloper314: Yes, a few of the dependencies use log4j directly and not through SLF4J at all. They all work as before. The dependencies don't know log4j is not used, the classfiles for reload4j have the same names I believe.

pom.xml
GCHQDeveloper314: Reload4J has removed support for some loggers in later versions, due to the code for them being in a poor state. This will affect only very rarely used loggers and won't affect this PR as the version of Reload4J used here still has support for them - one of the current test dependencies requires (but doesn't actually use) one of these loggers.

t11947 (approved):

pom.xml
GCHQDeveloper314: I've run through the tests with this PR and `develop` and compared the output. There's no differences which are related to the logging change.

t92549 (approved):

PR #2612, opened 2022-03-21
Upgrade Hazelcast version to 5.1.1

### LINKED ISSUE #2611: Port Hazelcast upgrade to Gaffer 1.*
A PR #2599 was raised by dependabot because of a security vulnerability in older versions of Hazelcast. The dependabot PR couldn't be merged because migrating to the latest Hazelcast version requires minor code changes.

Changes to upgrade Hazelcast have already been made in the Gaffer 2 dependencies upgrade branch, these need to be ported to Gaffer 1.* to fix any potential security vulnerability.

GCHQDeveloper314:
# Related Issue

- Resolve #2611

### DIFF
--- a/library/cache-library/hazelcast-cache-service/src/test/resources/hazelcast.xml
+++ b/library/cache-library/hazelcast-cache-service/src/test/resources/hazelcast.xml
@@ -1,9 +1,6 @@
 <?xml version="1.0" encoding="UTF-8" ?>
 <hazelcast xmlns="http://www.hazelcast.com/schema/config">
-    <group>
-        <name>test</name>
-        <password>password</password>
-    </group>
+    <cluster-name>test</cluster-name>
     <network>
         <port port-count="3" auto-increment="true">34562</port>
     </network>
--- a/library/cache-library/hazelcast-cache-service/pom.xml
+++ b/library/cache-library/hazelcast-cache-service/pom.xml
@@ -38,7 +38,7 @@
         <dependency>
             <groupId>com.hazelcast</groupId>
             <artifactId>hazelcast</artifactId>
-            <version>3.8</version>
+            <version>5.1.1</version>
         </dependency>
 
         <dependency>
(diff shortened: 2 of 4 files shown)


library/cache-library/hazelcast-cache-service/src/test/resources/hazelcast.xml
```diff
@@ -1,9 +1,6 @@
 <?xml version="1.0" encoding="UTF-8" ?>
 <hazelcast xmlns="http://www.hazelcast.com/schema/config">
-    <group>
-        <name>test</name>
-        <password>password</password>
-    </group>
+    <cluster-name>test</cluster-name>
```
r32575, line 3: Is there still the option to have a user name/password?

GCHQDeveloper314: [According to this migration guide there is not](https://docs.hazelcast.com/imdg/4.2/migration-guides#changes-in-security-configurations), it appears to have been deprecated for a while.

GCHQDeveloper314:
From what I can tell - I've only had a very quick look at the code - there's nothing for users to do as the code cannot connect to any external Hazelcast. I'm not sure why this module is called a 'cache service' at all. It's not what I would describe as a service, it's simply an implementation of `ICacheService` which uses a `HazelcastCache`. It doesn't offer any way to connect to an external Hazelcast cluster. It just [creates a single local 'node' inside the JVM and uses that](https://docs.hazelcast.org/docs/5.1/javadoc/com/hazelcast/core/Hazelcast.html#newHazelcastInstance-com.hazelcast.config.Config-).

[The code is very simple](https://github.com/gchq/Gaffer/tree/develop/library/cache-library/hazelcast-cache-service/src/main/java/uk/gov/gchq/gaffer/cache/impl) and this is why it is easy to migrate to this much newer version. I'm not sure this module is actually very useful to anyone as it is, it feels more like a proof of concept. Maybe we should consider deprecating it?

GCHQDev404 (approved):
No objections.

GCHQDeveloper314:
After having a closer look at this, it seems that a cluster can in fact be joined by the node created by Gaffer - if this is [specified in the config](https://docs.hazelcast.com/hazelcast/latest/clusters/discovering-by-tcp). A user currently running a cluster using an old version of Hazelcast would need to upgrade their cluster to continue using it with Gaffer. In a cluster created by other instances of Gaffer, so long as the other instances were also updated, the only change needed would be updates to the Hazelcast config to be compatible with the new version.

r32575 (approved):

t92549 (approved):

PR #2621, opened 2022-03-25
Gh-2620: Cherry-pick GetElementsWithinSet fix (#2597)

### LINKED ISSUE #2620: Cherry-pick GetElementsWithinSet fix into v2-alpha
Specifically https://github.com/gchq/Gaffer/commit/b8b00162fdaca8d67a4b2179aa2d80a4dab66a74

t92549:
* Fixed bug gh-2562. View properties are now being filtered. Added tests

# Related Issue

- Resolve #2620

### DIFF
--- a/store-implementation/accumulo-store/src/main/java/uk/gov/gchq/gaffer/accumulostore/retriever/AccumuloSetRetriever.java
+++ b/store-implementation/accumulo-store/src/main/java/uk/gov/gchq/gaffer/accumulostore/retriever/AccumuloSetRetriever.java
@@ -40,6 +40,7 @@ import uk.gov.gchq.gaffer.data.element.Edge;
 import uk.gov.gchq.gaffer.data.element.Element;
 import uk.gov.gchq.gaffer.data.element.Entity;
 import uk.gov.gchq.gaffer.data.element.id.EntityId;
+import uk.gov.gchq.gaffer.data.elementdefinition.view.ViewUtil;
 import uk.gov.gchq.gaffer.operation.graph.GraphFilters;
 import uk.gov.gchq.gaffer.operation.io.InputOutput;
 import uk.gov.gchq.gaffer.store.StoreException;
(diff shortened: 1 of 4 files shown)


store-implementation/accumulo-store/src/main/java/uk/gov/gchq/gaffer/accumulostore/retriever/AccumuloSetRetriever.java
```diff
@@ -399,7 +408,7 @@ private boolean _hasNext() throws RetrieverException {
                 updateScanner();
             }
             if (!scannerIterator.hasNext()) {
-                scanner.close();
+                CloseableUtil.close(scanner);
```
t11947, line 411: Worth noting that this swallows the exception whereas the previous method doesn't. Missed raising this in the original.

r32575: Not sure... would you care about it not closing correctly? I'll do some digging...

t92549: I am not sure either but it seems that, for now, this can just get merged and then be looked at as part of your closeable PR?

GCHQDeveloper314 (approved):
This is the same as #2597. If we want to make any improvements it might be best to do so in a new PR.

store-implementation/accumulo-store/src/main/java/uk/gov/gchq/gaffer/accumulostore/retriever/AccumuloSetRetriever.java
t92549: Think maybe this should be looked at separately as suggested here: https://github.com/gchq/Gaffer/pull/2621#pullrequestreview-932162393

PR #2628, opened 2022-04-08
gh-2627: Upgrade spring-boot version to 2.5.12

### LINKED ISSUE #2627: Upgrade to Spring Boot 2.5.12
Even though Gaffer 2 shouldn't be effected by the [RCE](https://spring.io/blog/2022/03/31/spring-framework-rce-early-announcement) due to not using a war, this should ensure it is never a problem.

t92549:
# Related Issue

- Resolve #2627


GCHQDeveloper314 (approved):
Only a patch version update, tests are passing. Approved

PR #2636, opened 2022-04-27
Gh 2201 upgrade accumulo version to 2.x.x

r32575:


GCHQDeveloper314:
What's the significance of the `ClosableIterable` changes? I had thought this was already removed before, but it seems it's still in use in various places.

In relation to the renamed/duplicated modules, are we sure the Accumulo 2 version of the code isn't also compatible with Accumulo 1? Also, I assume modules other than the Accumulo store were duplicated because they also need different incompatible code, not just dependencies? It's a little hard to tell from the diff.

r32575:
Superseded by #2679

Issue #2498, opened 2021-08-26
Upgrade store integration tests to use JUnit 5


t92549:
The 21 integration tests within `integration-test/` extend [AbstractStoreIT](https://github.com/gchq/Gaffer/blob/23c323e894c8dba3e25029c3699b6fb5a050a39e/integration-test/src/test/java/uk/gov/gchq/gaffer/integration/AbstractStoreIT.java#L79).

The 6 store types within `store-implementation/` extend [AbstractStoreITs](https://github.com/gchq/Gaffer/blob/23c323e894c8dba3e25029c3699b6fb5a050a39e/integration-test/src/test/java/uk/gov/gchq/gaffer/integration/AbstractStoreITs.java#L44). The `AbstractStoreITs` are then ran with a custom extension of JUnit 4's `Suite`:
https://github.com/gchq/Gaffer/blob/23c323e894c8dba3e25029c3699b6fb5a050a39e/integration-test/src/test/java/uk/gov/gchq/gaffer/integration/AbstractStoreITs.java#L117

The `StoreTestSuite` collects all of the `AbstractStoreIT` subclasses and runs them all with the specified store properties from the AbstractStoreITs implementation. 

This works in JUnit 4 but is hard to replicate in JUnit 5. This is because Suites work differently, both using the current [workaround](https://junit.org/junit5/docs/5.7.2/user-guide/#running-tests-junit-platform-runner-test-suite) and the future release's [full implementation](https://junit.org/junit5/docs/5.8.0-M1/user-guide/#test-suite). More testing could be done with these options but I think we should look to change the structure of the operation ITs and their interaction with the various stores to make it less complicated.

This approach will probably be necessary anyway during the Gaffer 2 upgrade due to the huge changes coming to how Operations work in #2487.

t92549:
Looks like some (or all) of this has been done in [develop-v2](https://github.com/gchq/Gaffer/tree/develop-v2), but I think this could be a Gaffer 1 issue

t92549:
Of note: #2530 reverted a test back to JUnit4 to fix a strange test failure, this should be fixed. https://github.com/gchq/Gaffer/blob/0f76569a17ac953dbfd48a0e06f558721c32607c/store-implementation/federated-store/src/test/java/uk/gov/gchq/gaffer/federatedstore/FederatedStoreGetTraitsTest.java#L22

GCHQDeveloper314:
[...]

PR #2648, opened 2022-05-17
gh-2498: Upgrade store integration tests (and everything else) to use JUnit 5

GCHQDeveloper314:
Also removes unimplemented Generics and associated Javadoc from store integration tests.


t92549 (approved):
Nice rework of the ITs to properly use JUnit5

Issue #2618, opened 2022-03-24
Cleanup inconsistent use of logging strings


t92549:
In many places in Gaffer, correct logging string formatting is used:
https://github.com/gchq/Gaffer/blob/fb63fc25be2f01d6476d1010780500cb0856b6c4/store-implementation/accumulo-store/src/main/java/uk/gov/gchq/gaffer/accumulostore/operation/hdfs/handler/AddElementsFromHdfsHandler.java#L254

However, in some places, string concatenation is incorrectly used instead:
https://github.com/gchq/Gaffer/blob/fb63fc25be2f01d6476d1010780500cb0856b6c4/store-implementation/accumulo-store/src/main/java/uk/gov/gchq/gaffer/accumulostore/operation/hdfs/handler/AddElementsFromHdfsHandler.java#L66

Although concatenation works, it is considered bad practice as the concatenated string object is always created, whether or not that string will be logged (based on the logging level). However, with the former method of formatting, the object is only created if that string will be logged. This means better performance.
As well as that I find the formatting easier to read.

RobertG111:
Hi. I have made a pull request regarding this issue #2651. Thanks.

PR #2651, opened 2022-05-19
gh-2618: Cleanup inconsistent use of logging strings

RobertG111:
Hi. This is my first ever contribution, thanks for the opportunity. The other pull request I created was closed due to having my personal email address. I have fixed the minor issues in relation to inconsistent use of logging strings. All logging functions used the correct format and was able to find only 1 extra mistake apart from the mentioned one in the issue. Thanks.

### DIFF
--- a/core/common-util/src/main/java/uk/gov/gchq/gaffer/commonutil/ExecutorService.java
+++ b/core/common-util/src/main/java/uk/gov/gchq/gaffer/commonutil/ExecutorService.java
@@ -39,7 +39,7 @@ public final class ExecutorService {
 
     public static synchronized void initialise(final int jobExecutorThreadCount) {
         if (service == null) {
-            LOGGER.debug("Initialising ExecutorService with {} threads", jobExecutorThreadCount);
+            LOGGER.debug("Initialising ExecutorService with {} threads",jobExecutorThreadCount);
             service = Executors.newScheduledThreadPool(jobExecutorThreadCount, runnable -> {
                 final Thread thread = new Thread(runnable);
                 thread.setDaemon(true);
--- a/core/common-util/src/main/java/uk/gov/gchq/gaffer/commonutil/ExecutorService.java
+++ b/core/common-util/src/main/java/uk/gov/gchq/gaffer/commonutil/ExecutorService.java
@@ -39,7 +39,7 @@ public final class ExecutorService {
 
     public static synchronized void initialise(final int jobExecutorThreadCount) {
         if (service == null) {
-            LOGGER.debug("Initialising ExecutorService with " + jobExecutorThreadCount + " threads");
+            LOGGER.debug("Initialising ExecutorService with {} ", jobExecutorThreadCount ," threads");
             service = Executors.newScheduledThreadPool(jobExecutorThreadCount, runnable -> {
                 final Thread thread = new Thread(runnable);
                 thread.setDaemon(true);
--- a/core/common-util/src/main/java/uk/gov/gchq/gaffer/commonutil/ExecutorService.java
+++ b/core/common-util/src/main/java/uk/gov/gchq/gaffer/commonutil/ExecutorService.java
@@ -39,7 +39,7 @@ public final class ExecutorService {
 
     public static synchronized void initialise(final int jobExecutorThreadCount) {
         if (service == null) {
-            LOGGER.debug("Initialising ExecutorService with {} ", jobExecutorThreadCount ," threads");
+            LOGGER.debug("Initialising ExecutorService with {} threads", jobExecutorThreadCount);
             service = Executors.newScheduledThreadPool(jobExecutorThreadCount, runnable -> {
                 final Thread thread = new Thread(runnable);
                 thread.setDaemon(true);
--- a/core/common-util/src/main/java/uk/gov/gchq/gaffer/commonutil/ExecutorService.java
+++ b/core/common-util/src/main/java/uk/gov/gchq/gaffer/commonutil/ExecutorService.java
@@ -39,7 +39,7 @@ public final class ExecutorService {
 
     public static synchronized void initialise(final int jobExecutorThreadCount) {
         if (service == null) {
-            LOGGER.debug("Initialising ExecutorService with {} threads",jobExecutorThreadCount);
+            LOGGER.debug("Initialising ExecutorService with {} threads", jobExecutorThreadCount);
             service = Executors.newScheduledThreadPool(jobExecutorThreadCount, runnable -> {
                 final Thread thread = new Thread(runnable);
                 thread.setDaemon(true);
--- a/core/graph/src/main/java/uk/gov/gchq/gaffer/graph/Graph.java
+++ b/core/graph/src/main/java/uk/gov/gchq/gaffer/graph/Graph.java
@@ -342,7 +342,7 @@ public final class Graph {
                 try {
                     result = graphHook.onFailure(result, clonedOpChain, clonedContext, e);
                 } catch (final Exception graphHookE) {
-                    LOGGER.warn("Error in graphHook {}", graphHook.getClass().getSimpleName() ,": ", graphHookE.getMessage(), graphHookE);
+                    LOGGER.warn("Error in graphHook {} : {}", graphHook.getClass().getSimpleName() ,graphHookE.getMessage(), graphHookE);
                 }
             }
             CloseableUtil.close(clonedOpChain);
--- a/core/graph/src/main/java/uk/gov/gchq/gaffer/graph/Graph.java
+++ b/core/graph/src/main/java/uk/gov/gchq/gaffer/graph/Graph.java
@@ -285,7 +285,7 @@ public final class Graph {
                 try {
                     result = graphHook.onFailure(result, clonedOpChain, clonedContext, e);
                 } catch (final Exception graphHookE) {
-                    LOGGER.warn(String.format("Error in graphHook %s : %s", graphHook.getClass().getSimpleName(), graphHookE.getMessage()), graphHookE);
+                    LOGGER.warn("Error in graphHook {} : {}", graphHook.getClass().getSimpleName(), graphHookE.getMessage(), graphHookE);
                 }
             }
             CloseableUtil.close(clonedOpChain);
@@ -342,7 +342,7 @@ public final class Graph {
                 try {
                     result = graphHook.onFailure(result, clonedOpChain, clonedContext, e);
                 } catch (final Exception graphHookE) {
-                    LOGGER.warn(String.format("Error in graphHook %s : %s", graphHook.getClass().getSimpleName(), graphHookE.getMessage()), graphHookE);
+                    LOGGER.warn("Error in graphHook {} : {}", graphHook.getClass().getSimpleName(), graphHookE.getMessage(), graphHookE);
                 }
             }
             CloseableUtil.close(clonedOpChain);
--- a/core/graph/src/main/java/uk/gov/gchq/gaffer/graph/Graph.java
+++ b/core/graph/src/main/java/uk/gov/gchq/gaffer/graph/Graph.java
@@ -285,7 +285,7 @@ public final class Graph {
                 try {
                     result = graphHook.onFailure(result, clonedOpChain, clonedContext, e);
                 } catch (final Exception graphHookE) {
-                    LOGGER.warn("Error in graphHook " + graphHook.getClass().getSimpleName() + ": " + graphHookE.getMessage(), graphHookE);
+                    LOGGER.warn("Error in graphHook {}", graphHook.getClass().getSimpleName() ,": ", graphHookE.getMessage(), graphHookE);
                 }
             }
             CloseableUtil.close(clonedOpChain);
@@ -342,7 +342,7 @@ public final class Graph {
                 try {
                     result = graphHook.onFailure(result, clonedOpChain, clonedContext, e);
                 } catch (final Exception graphHookE) {
-                    LOGGER.warn("Error in graphHook " + graphHook.getClass().getSimpleName() + ": " + graphHookE.getMessage(), graphHookE);
+                    LOGGER.warn("Error in graphHook {}", graphHook.getClass().getSimpleName() ,": ", graphHookE.getMessage(), graphHookE);
                 }
             }
             CloseableUtil.close(clonedOpChain);
--- a/core/graph/src/main/java/uk/gov/gchq/gaffer/graph/Graph.java
+++ b/core/graph/src/main/java/uk/gov/gchq/gaffer/graph/Graph.java
@@ -285,7 +285,7 @@ public final class Graph {
                 try {
                     result = graphHook.onFailure(result, clonedOpChain, clonedContext, e);
                 } catch (final Exception graphHookE) {
-                    LOGGER.warn(String.format("Error in graphHook %s : %s", graphHook.getClass().getSimpleName() , graphHookE.getMessage()), graphHookE);
+                    LOGGER.warn(String.format("Error in graphHook %s : %s", graphHook.getClass().getSimpleName(), graphHookE.getMessage()), graphHookE);
                 }
             }
             CloseableUtil.close(clonedOpChain);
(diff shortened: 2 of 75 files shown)


GCHQDeveloper314 (changes requested):
Hello, thanks for your contribution. I've renamed the PR title to fit the naming scheme we use, this is partly described on our new [ways of working](https://gchq.github.io/gaffer-doc/latest/ways-of-working/#workflow) page.

In your commit you've missed out the `{}` mentioned in the issue. This is required to add the parameter to the formatted log string.

Also, the issue calls for the problem to be fixed across the repo. You've found one other instance, but there are many others. Searching the repo using `LOGGER.*\+` as a regex query, I can see 20+ other instances of string concatenation (some of which are partly using parameters).

RobertG111:
Noted. Thanks for the fast reply.

core/common-util/src/main/java/uk/gov/gchq/gaffer/commonutil/ExecutorService.java
```diff
@@ -39,7 +39,7 @@ private ExecutorService() {
 
     public static synchronized void initialise(final int jobExecutorThreadCount) {
         if (service == null) {
-            LOGGER.debug("Initialising ExecutorService with " + jobExecutorThreadCount + " threads");
+            LOGGER.debug("Initialising ExecutorService with {} ", jobExecutorThreadCount ," threads");
```
t92549, line 42: As explained in the issue (https://github.com/gchq/Gaffer/issues/2618), these should look like: 
```java
LOGGER.debug("Initialising ExecutorService with {} threads", jobExecutorThreadCount);
```
It uses SLF4J's [parameterized logging](https://www.tutorialspoint.com/slf4j/slf4j_parameterized_logging.htm), which is similar to Java's string format method.

t92549 (changes requested):
Thanks for the PR! Just a small change of syntax is required to get this working, I have linked an article in the review comment.

core/common-util/src/main/java/uk/gov/gchq/gaffer/commonutil/ExecutorService.java
RobertG111: Hi. I have made the required changes as stated above. If there are anymore changes required do let me know. Do you want me to squash the commits as well? Thanks a lot for your help and reviews.

core/graph/src/main/java/uk/gov/gchq/gaffer/graph/Graph.java
```diff
@@ -285,7 +285,7 @@ public JobDetail executeJob(final Job job, final Context context) throws Operati
                 try {
                     result = graphHook.onFailure(result, clonedOpChain, clonedContext, e);
                 } catch (final Exception graphHookE) {
-                    LOGGER.warn("Error in graphHook " + graphHook.getClass().getSimpleName() + ": " + graphHookE.getMessage(), graphHookE);
+                    LOGGER.warn("Error in graphHook {}", graphHook.getClass().getSimpleName() ,": ", graphHookE.getMessage(), graphHookE);
```
t92549, line 288: Confusingly, these usages are using the last argument, `graphHookE`, as a `Throwable`, and the first argument as a `String`. They look like [this](https://www.slf4j.org/api/org/slf4j/Logger.html#warn-java.lang.String-java.lang.Throwable-) rather than the other examples that look like [this](https://www.slf4j.org/api/org/slf4j/Logger.html#warn-java.lang.String-java.lang.Object...-).
A way to keep this working might be to change it like: 
```java
LOGGER.warn(String.format("Error in graphHook %s : %s", graphHook.getClass().getSimpleName(), graphHookE.getMessage()), graphHookE);
```

library/cache-library/hazelcast-cache-service/src/main/java/uk/gov/gchq/gaffer/cache/impl/HazelcastCacheService.java
```diff
@@ -42,7 +42,7 @@ private void configureHazelcast(final Properties properties) {
         if (null == hazelcast || !Hazelcast.getAllHazelcastInstances().contains(hazelcast)) {
             String configFile = properties.getProperty(CACHE_CONFIG_FILE);
             if (null == configFile) {
-                LOGGER.warn("Config file not set using system property: " + CACHE_CONFIG_FILE
+                LOGGER.warn("Config file not set using system property: {}", CACHE_CONFIG_FILE
```
t92549, line 45: This is still using `+`

rest-api/spring-rest/src/main/java/uk/gov/gchq/gaffer/rest/controller/GraphConfigurationController.java
```diff
@@ -96,7 +96,7 @@ public Set<Class> getFilterFunctions(@PathVariable("inputClass") final String in
             try {
                 predicate = (Predicate) predicateClass.newInstance();
             } catch (final IllegalAccessException | InstantiationException e) {
-                LOGGER.warn("Failed to create new instance of " + predicateClass, e);
+                LOGGER.warn("Failed to create new instance of {}", predicateClass, e);
```
t92549, line 99: Like above, this is using `e` as a `Throwable` so this won't work. A good tell is that there are more arguments than `{}`s. You can use the same solution as above with `String.format`

store-implementation/accumulo-store/src/main/java/uk/gov/gchq/gaffer/accumulostore/operation/hdfs/handler/AddElementsFromHdfsHandler.java
```diff
@@ -63,7 +63,7 @@ public void doOperation(final AddElementsFromHdfs operation,
 
         if (null == operation.getSplitsFilePath()) {
             final String splitsFilePath = getPathWithSlashSuffix(operation.getWorkingPath()) + context.getJobId() + "/splits";
-            LOGGER.info("Using working directory for splits files: " + splitsFilePath);
+            LOGGER.info("Using working directory for splits files: ", splitsFilePath);
```
t92549, line 66: You forgot to add the `{}` here.

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/FederatedGraphStorage.java
```diff
@@ -653,7 +653,7 @@ private boolean changeGraphId(final String graphId, final String newGraphId, fin
                     federatedStoreCache.addGraphToCache(newGraphSerialisable, key, true);
                 } catch (final CacheOperationException e) {
                     String s = "Contact Admin for recovery. Error occurred updating graphId. GraphStorage=updated, Cache=outdated graphId.";
-                    LOGGER.error(s + " graphStorage graphId:{} cache graphId:{}", newGraphId, graphId);
+                    LOGGER.error(s ,"{} graphStorage graphId:{} cache graphId:{}", newGraphId, graphId);
```
t92549, line 656: This needs to be fixed like you did on line [577](https://github.com/gchq/Gaffer/pull/2651/files#diff-f63625e74223d0e72648e1658bf66b918b576b71f5228eba98044b277155013aR577).

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/operation/handler/FederatedAddGraphHandlerParent.java
```diff
@@ -95,13 +95,13 @@ protected void addGenericHandler(final FederatedStore store, final Graph graph)
                     try {
                         outputClass = supportedOutputOperation.newInstance().getOutputClass();
                     } catch (final InstantiationException | IllegalAccessException e) {
-                        LOGGER.warn("Exception occurred while trying to create a newInstance of operation: " + supportedOperation, e);
+                        LOGGER.warn("Exception occurred while trying to create a newInstance of operation: {}", supportedOperation, e);
```
t92549, line 98: Same issue with the `Throwable` at the end.

store-implementation/map-store/src/main/java/uk/gov/gchq/gaffer/mapstore/impl/AddElementsHandler.java
```diff
@@ -65,7 +65,7 @@ private void addElements(final Iterable<? extends Element> elements, final MapSt
             // Add all elements directly
             addBatch(mapImpl, schema, elements);
         } else {
-            LOGGER.info("Adding elements in batches, batch size = " + bufferSize);
+            LOGGER.info("Adding elements in batches, batch size = ", bufferSize);
```
t92549, line 68: Missing a `{}`

t92549 (changes requested):
Thanks for all the updates! Still a couple more issues

core/common-util/src/main/java/uk/gov/gchq/gaffer/commonutil/ExecutorService.java
t92549: We will squash the commits when we merge into develop

t92549: No worries, thank you for contributing!

core/graph/src/main/java/uk/gov/gchq/gaffer/graph/hook/FunctionAuthoriser.java
```diff
@@ -71,7 +71,7 @@ public void preExecute(final OperationChain<?> opChain, final Context context) {
             // This should never happen in real life as operation chains should
             // always be json serialisable. However this could happen if using a
             // mock in testing. To account for this, it will be logged.
-            LOGGER.warn("Failed to serialise operation chain: " + opChain + " due to " + e.getMessage());
+            LOGGER.warn("Failed to serialise operation chain: {} due to {}", opChain , e.getMessage());
```
t92549, line 74: Your PR is failing the checkstyle stage of the CI, mostly due to whitespace issues around commas. You can run the checkstyle locally to see what exactly is wrong with `mvn checkstyle:checkstyle`.

t92549 (changes requested):

core/graph/src/main/java/uk/gov/gchq/gaffer/graph/hook/FunctionAuthoriser.java
RobertG111: Will fix this. Thanks for the notice.

core/graph/src/main/java/uk/gov/gchq/gaffer/graph/Graph.java
```diff
@@ -285,7 +285,7 @@ public JobDetail executeJob(final Job job, final Context context) throws Operati
                 try {
                     result = graphHook.onFailure(result, clonedOpChain, clonedContext, e);
                 } catch (final Exception graphHookE) {
-                    LOGGER.warn("Error in graphHook " + graphHook.getClass().getSimpleName() + ": " + graphHookE.getMessage(), graphHookE);
+                    LOGGER.warn(String.format("Error in graphHook %s : %s", graphHook.getClass().getSimpleName(), graphHookE.getMessage()), graphHookE);
```
GCHQDeveloper314, line 288: @t92549 In the issue (#2618), you say that concatenation "is considered bad practice as the concatenated string object is always created, whether or not that string will be logged". Yet this solution has exactly the same problem. The `String.format` static call will always be made regardless of logging level.

SLF4J doesn't offer a variant which accepts a formatting string, objects to be added into the string **and** an exception/throwable. Although it could offer it, considering how the formatting is done (see [here](https://github.com/qos-ch/slf4j/blob/82a592ed790cc8072aef1455f1065b6c9d207501/slf4j-api/src/main/java/org/slf4j/helpers/AbstractLogger.java) & [here](https://github.com/qos-ch/slf4j/blob/82a592ed790cc8072aef1455f1065b6c9d207501/slf4j-reload4j/src/main/java/org/slf4j/reload4j/Reload4jLoggerAdapter.java#L133)).

In this example, `graphHookE` is the exception and I presume we want to log the stack trace (as this is what happens when we include it in the call to the logger). Why is it that we sometimes do this and not others? If we don't really need the stack trace then we shouldn't include it and thereby remove a need to create the string object.

RobertG111: @GCHQDeveloper314 Is there anyway to work around the fact that we need to use String.format or + to concatenate the strings? Maybe make a variable with the msg value and assign it to the function?

t92549: You are right, just did some testing and I apologise, indeed the best way to go about this is something like:
```java
LOGGER.warn("Error in graphHook {} : {}", graphHook.getClass().getSimpleName(), graphHookE.getMessage(), graphHookE); 
```

GCHQDeveloper314: There's not an easy way to work around it. [As is written in the Javadoc](https://github.com/qos-ch/slf4j/blob/82a592ed790cc8072aef1455f1065b6c9d207501/slf4j-api/src/main/java/org/slf4j/Logger.java#L613), the varargs variant still causes an object to be created (although this is still better than concatenation). It is better for readability, but I think it might be arguable that if logging is enabled it could be less efficient vs `String.format` considering the extra work required to determine if the final object is a thowable. Another different option would be to convert the exception to a string before passing it in ([could use this](https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/exception/ExceptionUtils.html#getStackTrace-java.lang.Throwable-)), although that always creates an extra object which might not get logged and is probably less efficient.

The best (most efficient) solution would be to make a helper class (that extends SLF4J) which would take all the parameters we want and not do any unnecessary processing. That is far out of scope of this issue and most likely not something we'd want to do anyway. Ensuring we only log exceptions we need (to reduce overhead more generally) could be raised as another issue but it's pretty trivial really.

Ultimately, the above is mostly out of scope of this issue. As you have pointed out, there's no need for the String format and the `Exception` can be added as the final argument.

So far as this PR is concerned, `String.format` is not required and so shouldn't be used. With that done, we just need to remove the remaining instances of string concatenation. Using `grep -EIr "LOGGER.*\+"` I can see another 8 remaining in the codebase. Once that is all done this can be merged in.

t92549 (changes requested):
Great changes, just a few more I found that are missed:
- https://github.com/gchq/Gaffer/blob/6a0c8636f71c47096fc3d99401688b8815a32933/example/road-traffic/road-traffic-generators/src/main/java/uk/gov/gchq/gaffer/traffic/generator/RoadTrafficDataLoader.java#L124
- https://github.com/gchq/Gaffer/blob/6a0c8636f71c47096fc3d99401688b8815a32933/example/road-traffic/road-traffic-rest/src/main/java/uk/gov/gchq/gaffer/traffic/listeners/DataLoader.java#L60
- https://github.com/gchq/Gaffer/blob/6a0c8636f71c47096fc3d99401688b8815a32933/example/road-traffic/road-traffic-rest/src/test/java/uk.gov.gchq.gaffer.traffic/RoadTrafficTestQueries.java#L116
- https://github.com/gchq/Gaffer/blob/6a0c8636f71c47096fc3d99401688b8815a32933/store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/FederatedGraphStorage.java#L507
- https://github.com/gchq/Gaffer/blob/6a0c8636f71c47096fc3d99401688b8815a32933/example/road-traffic/road-traffic-rest/src/main/java/uk/gov/gchq/gaffer/traffic/listeners/ConsoleBanner.java#L36

GCHQDeveloper314:
The latest error is caused by the logging in `RoadTrafficDataLoader.java` being done from Java's util logger. This is used in a few places across road-traffic. Ideally SLF4J would be used, however that is not within scope of this ticket. Reverting any changes to the util logger (because it doesn't support parameterised logging) is what we'll want to do in this PR. Investigating changing to SLF4J is something we might raise as a new ticket.

t92549:
> The latest error is caused by the logging in `RoadTrafficDataLoader.java` being done from Java's util logger. This is used in a few places across road-traffic. Ideally SLF4J would be used, however that is not within scope of this ticket. Reverting any changes to the util logger (because it doesn't support parameterised logging) is what we'll want to do in this PR. Investigating changing to SLF4J is something we might raise as a new ticket.

That's my bad sorry, I said it needs changing but didn't realise it was using the java.util logger, apologies!

t92549:
I think all cases are now covered, to get the CI passing the road-traffic example cases need to be reverted because they aren't using SLF4J, sorry!

GCHQDeveloper314 (approved):
I can see a few remaining string concatenations across multiple lines, but there's nothing we can do about that as we are not using a new enough version of Java to use avoid this. PR looks to be ready to merge.

RobertG111:
Thanks @t92549 @GCHQDeveloper314 for the help and reviews. It has been great working on this and will continue contributing. Many thanks.

t92549 (approved):
Thanks for the PR! Apologies that a couple of times I suggested things that were actually incorrect!

PR #2658, opened 2022-05-26
gh-2657: Cherry-pick gh-2618: Cleanup inconsistent use of logging strings (#2651)

### LINKED ISSUE #2657: Cherry-pick logging strings into v2-alpha

t92549:
# Related Issue

- Resolve #2657

### DIFF
--- a/example/road-traffic/road-traffic-generators/src/main/java/uk/gov/gchq/gaffer/traffic/generator/RoadTrafficDataLoader.java
+++ b/example/road-traffic/road-traffic-generators/src/main/java/uk/gov/gchq/gaffer/traffic/generator/RoadTrafficDataLoader.java
@@ -121,7 +121,7 @@ public class RoadTrafficDataLoader {
             dataLoader.load(new File(dataFile));
             LOGGER.info("Data has been loaded");
         } catch (final Exception e) {
-            LOGGER.info("Unable to load data:" + e.getMessage());
+            LOGGER.info("Unable to load data: " + e.getMessage());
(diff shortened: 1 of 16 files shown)


example/road-traffic/road-traffic-generators/src/main/java/uk/gov/gchq/gaffer/traffic/generator/RoadTrafficDataLoader.java
```diff
@@ -121,7 +121,7 @@ public static void main(final String[] args) {
             dataLoader.load(new File(dataFile));
             LOGGER.info("Data has been loaded");
         } catch (final Exception e) {
-            LOGGER.info("Unable to load data: " + e.getMessage());
```
GCHQDeveloper314, line 124: This file shouldn't have been changed. I think this will have been caused by adding something and not reverting correctly.

GCHQDeveloper314 (changes requested):
One file that doesn't need changing. Also needs copyright 😉

example/road-traffic/road-traffic-generators/src/main/java/uk/gov/gchq/gaffer/traffic/generator/RoadTrafficDataLoader.java
t92549: Correct, this was incorrectly reverted, have removed change and fixed copyright

GCHQDeveloper314 (approved):

PR #2659, opened 2022-05-27
Updated Koryphe version to 2.4.0

t92549:


t92549:
Automatic PR must be created manually (this will stop happening after alphas)

Issue #2201, opened 2019-10-22
Upgrade Accumulo version to 2.x.x


d47853:
Upgrade version of Accumulo so that we can use the Accumulo Docker container as the baseline for Gaffer containerisation.

There is an accumulo-docker project that works for Accumulo 2.0.0+ which is supported apache accumulo but is not released to Dockerhub
There is a branch of accumulo-docker which supports accumulo 1.9.3+

Upgrading to 1.9.3 is fairly simple but 2.0.0 would need more work as MockAccumulo  is removed in this version. We would need to replace all the MockAccumuloStores from Gaffer should we choose to change this. As this would be a breaking change, it would have to be done in Gaffer 2.x.x

Helpful information available here: https://accumulo.apache.org/docs/2.x/administration/upgrading#upgrading-from-189-to-20

Linked to #2200

cybermaggedon:
I would be willing to work on this, would need some guidance with the code structure. I'm familiar with Accumulo deply, and have Accumulo 2 working in containers which I can push somewhere.

d47853:
Excellent! We recently updated all of our [Docker codebase](https://github.com/gchq/gaffer-docker). We also have an Accumulo (albeit 1.9.3) image that is pushed to docker hub. So we'll update those images when we roll out a version of Gaffer which supports Accumulo 2.x.

n3101:
Accumulo 2.0.0 has critical bugs; 2.0.1 is not a Long Term Maintenance version; there are no other Accumulo 2 versions today. We need to do a Gaffer 2 release this year to support federation. To simplify this process, Accumulo 2 will be left out of the Gaffer 2.0.0 release. We will come back to this later.

n3101:
Ah, no it seems we have to include Accumulo 2 in the Gaffer 2.0.0 release

n3101:
The requirement now is for Gaffer to work with either an Accumulo 1 or an Accumulo 2 backing store. Also, for federation,  a query must be able to federate across a [Gaffer 2.0 + Accumulo 1 instance] and a [Gaffer2.0 + Accumulo 2 instance].

t92549:
Closed by https://github.com/gchq/Gaffer/pull/2679

PR #2679, opened 2022-06-17
Gh-2201: Add support for Accumulo 2 while retaining Accumulo 1

GCHQDeveloper314:
This make Gaffer compatible with Accumulo 2 and retains support for Accumulo 1.

Uses a utility class to dynamically import Accumulo classes which are incompatible between versions 1 & 2, plus handles creation of certain objects associated with these classes. Gaffer can be built for Accumulo 1 and Hadoop 2 using a 'classic' profile which swaps dependency versions. Other minor POM changes were required to ensure compatibility for existing libraries.

### DIFF
--- a/store-implementation/accumulo-store/src/main/java/uk/gov/gchq/gaffer/accumulostore/utils/IngestUtils.java
+++ b/store-implementation/accumulo-store/src/main/java/uk/gov/gchq/gaffer/accumulostore/utils/IngestUtils.java
@@ -20,7 +20,6 @@ import org.apache.accumulo.core.client.AccumuloException;
 import org.apache.accumulo.core.client.AccumuloSecurityException;
 import org.apache.accumulo.core.client.Connector;
 import org.apache.accumulo.core.client.TableNotFoundException;
-import org.apache.accumulo.core.client.mock.MockConnector;
 import org.apache.commons.codec.binary.Base64;
 import org.apache.hadoop.fs.FSDataInputStream;
 import org.apache.hadoop.fs.FileStatus;
@@ -81,12 +80,6 @@ public final class IngestUtils {
         }
         // This should have returned at most maxSplits splits, but this is not implemented properly in MockInstance.
         if (splits.size() > maxSplits) {
-            if (conn instanceof MockConnector) {
-                LOGGER.info("Manually reducing the number of splits to {} due to MockInstance not implementing"
-                        + " listSplits(table, maxSplits) properly", maxSplits);
-            } else {
-                LOGGER.info("Manually reducing the number of splits to {} (number of splits was {})", maxSplits, splits.size());
-            }
             final Collection<Text> filteredSplits = new TreeSet<>();
             final int outputEveryNth = splits.size() / maxSplits;
             LOGGER.info("Outputting every {}-th split from {} total", outputEveryNth, splits.size());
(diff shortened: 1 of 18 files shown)


store-implementation/accumulo-store/src/main/java/uk/gov/gchq/gaffer/accumulostore/utils/IngestUtils.java
```diff
@@ -81,12 +81,6 @@ public static int createSplitsFile(final Connector conn, final String table, fin
         }
         // This should have returned at most maxSplits splits, but this is not implemented properly in MockInstance.
         if (splits.size() > maxSplits) {
-            if (conn instanceof MockConnector) {
-                LOGGER.info("Manually reducing the number of splits to {} due to MockInstance not implementing"
-                        + " listSplits(table, maxSplits) properly", maxSplits);
-            } else {
-                LOGGER.info("Manually reducing the number of splits to {} (number of splits was {})", maxSplits, splits.size());
```
t92549, line 88: Is it not worth logging this anymore? As we are never using a `MockConnector`?

store-implementation/accumulo-store/src/test/java/uk/gov/gchq/gaffer/accumulostore/utils/LegacySupportTest.java
```diff
@@ -0,0 +1,234 @@
...
+ * utility class is able to create and invoke Accumulo methods.
+ */
+public class LegacySupportTest {
+
+    @TempDir
+    static File tempDir;
+
+    @Test
+    public void shouldReflectForSetScanAuthorizations() {
+        // With
```
t92549, line 60: Very small point but tests are usually given / when / then

GCHQDeveloper314: I still have work to do on the tests. I'll be adding AssertJ as well 😀

store-implementation/accumulo-store/src/main/java/uk/gov/gchq/gaffer/accumulostore/utils/IngestUtils.java
GCHQDeveloper314: I copied this from the old PR for upgrading to Accumulo 2. Accumulo 2 doesn't have `MockConnector` or a direct equivalent so this needed to be dropped.

t92549: I meant the else part, as it effectively should always be hitting that else block now

GCHQDeveloper314: I simply copied this change over from the old PR, without taking too much of a look at it. You're right that this doesn't make much sense. Depending on what happened to `MockInstance` it might be that more code should be removed. As the comment implies the situation of `splits.size() > maxSplits` is related to `MockInstance`. If there's no `MockInstance` in use anymore then the whole `if` could potentially be removed. I'll take a further look at this.

r32575 (approved):

t92549 (approved):

PR #2687, opened 2022-06-27
Gh 2686 remove unnecessary hadoop dependencies

r32575:
There are `hadoop` dependencies in the `Gaffer` core `common-util`. These should be removed as the core `Gaffer` components should not have a dependency on one of the niche dependencies of the `accumulo-store`.


pom.xml
```diff
@@ -228,6 +229,11 @@
                 <artifactId>commons-math</artifactId>
                 <version>${commons-math.version}</version>
             </dependency>
+            <dependency>
+                <groupId>commons-collections</groupId>
+                <artifactId>commons-collections</artifactId>
+                <version>${commons-collections.version}</version>
+            </dependency>
```
GCHQDeveloper314, line 236: This version of commons collections is old and based on a very old version. We should instead use commons collections version 4 which is much more recent.
````
<dependency>
    <groupId>org.apache.commons</groupId>
    <artifactId>commons-collections4</artifactId>
    <version>4.4</version>
</dependency>
````
This requires a `4` to be added to import statements in a couple of files.

PR #2698, opened 2022-07-01
Gh-2695: Unsilence CI output and fix timeout

### LINKED ISSUE #2695: Unsilence CI build output and investigate CI timeout
Currently all Gaffer tests in the GitHub Actions workflows use the Maven `-q` flag for silent output.
However, despite this option a large amount of test output is still generated, most likely because of logging which bypasses Maven. The consequence is a log with output from tests but none of the structure describing which test is running. The silent flag should be removed to improve the usefulness of this output.

There's an inconsistent timeout problem with one of the action steps ([can be seen in this run](https://github.com/gchq/Gaffer/actions/runs/2591050164/attempts/1)), due to the lack of logging the cause is unclear. It doesn't occur locally and doesn't happen 2/3 of the time, so is presumably related to GitHub Actions itself. This needs to be fixed.

GCHQDeveloper314:
# Related Issue

- Resolve #2695

### DIFF
--- a/.github/workflows/continuous-integration.yaml
+++ b/.github/workflows/continuous-integration.yaml
@@ -31,8 +31,8 @@ jobs:
 
     - name: Build Javadoc
       run: |
-        mvn -q clean install -Pquick -Dskip.jar-with-dependencies=true -Dshaded.jar.phase=true
-        mvn -q javadoc:javadoc -Pquick
+        mvn clean install -Pquick -Dskip.jar-with-dependencies=true -Dshaded.jar.phase=true
+        mvn javadoc:javadoc -Pquick
 
   build:
     runs-on: ubuntu-latest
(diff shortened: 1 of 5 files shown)


GCHQDeveloper314:
So far these changes fix a problem with the reactor build order but not the inconsistent CI timeout problem.

GCHQDeveloper314:
The inconsistent CI timeout problem is now fixed. This was caused by a change in the order which JUnit ran certain core-rest tests, the root cause of this is unclear because the way JUnit selects the test execution order is not well documented, it was presumably triggered by a change to the REST API code.

This shouldn't have been a problem, except there was a preexisting problem which required tests to be run in a certain order - this was disrupted by the change above. The problematic test execution order has now be set manually.

t92549 (approved):

PR #2715, opened 2022-07-21
Gh 2708 move integration tests from test to main

### LINKED ISSUE #2708: Move integration tests from test to main
At present the integration tests are part of `integration-test/src/test`. With the raising of issue #2654 unit tests are required to test integration test extensions. However, at present this would mean that the unit tests would be bundled up with the integration tests which is undesirable. Therefore, we should move the integration tests to main and alter dependencies.

r32575:
# Related Issue

- Resolve #2708


library/hdfs-library/src/test/java/uk/gov/gchq/gaffer/hdfs/integration/loader/AddElementsFromHdfsLoaderIT.java
```diff
@@ -128,18 +127,17 @@ public void _setup() throws Exception {
...
-                setup(testInfo);
-                fail("Exception expected");
-            } catch (final OperationException e) {
-                assertThat(e.getCause().getMessage()).isEqualTo("Failure directory is not empty: " + failureDir);
+                assertThatExceptionOfType(Exception.class)
+                        .isThrownBy(() -> setup(testInfo))
+                        .withStackTraceContaining("Failure directory is not empty: " + failureDir);
             } finally {
                 tearDown();
             }
```
GCHQDeveloper314, line 143: This can be further improved by removing the `try` and `finally` entirely. The `assertThatExceptionOfType` doesn't throw anything, meaning the `try` isn't needed, therefore the `finally` is not needed either.

library/hdfs-library/src/test/java/uk/gov/gchq/gaffer/hdfs/integration/loader/AddElementsFromHdfsLoaderIT.java
```diff
@@ -170,10 +168,9 @@ public void shouldThrowExceptionWhenAddElementsFromHdfsWhenOutputDirectoryContai
...
-                setup(testInfo);
-                fail("Exception expected");
-            } catch (final Exception e) {
-                assertThat(e.getMessage()).as(e.getMessage()).contains("Output directory exists and is not empty: " + outputDir);
+                assertThatExceptionOfType(Exception.class)
+                        .isThrownBy(() -> setup(testInfo))
+                        .withStackTraceContaining("Output directory exists and is not empty: " + outputDir);
             } finally {
                 tearDown();
             }
```
GCHQDeveloper314, line 176: As previous, `try` and `finally` can be removed.

integration-test/src/main/java/uk/gov/gchq/gaffer/integration/AbstractStoreITs.java
```diff
@@ -46,8 +46,6 @@ public AbstractStoreITs(final StoreProperties storeProperties, final Schema sche
         this.skipTestMethods = AbstractStoreIT.getSkipTestMethods();
         AbstractStoreIT.setStoreSchema(schema);
         AbstractStoreIT.setStoreProperties(storeProperties);
-        System.out.println(storeProperties);
-        System.out.println(schema);
```
GCHQDeveloper314, line 50: I noticed these when I was restructuring this class a few months ago. I wasn't sure what they were for so left them in. There doesn't seem to be any purpose for these statements, so if there's no other objection I don't see why we shouldn't remove them.

library/hdfs-library/src/test/java/uk/gov/gchq/gaffer/hdfs/integration/loader/AddElementsFromHdfsLoaderIT.java
GCHQDeveloper314: To be clear, the `tearDown()` should be kept, it's the `finally` surrounding it that can be removed.

r32575: If the test fails will the `tearDown` still be called if it is in a `finally` block? If it fails and the `tearDown` is not in a `finally` block then I figure it will not be. Seems like it should be in an `@AfterEach` method instead...

GCHQDeveloper314: You're correct. If the assertion fails, then the lines below will not be called. I'm not sure if this throws an exception though, it may stop execution using another method which would prevent the `finally` from being run.

The good news is that is you dig down to the class where `tearDown` is declared, it turns out this is executed `@AfterAll` tests anyway:
https://github.com/gchq/Gaffer/blob/9986d3b0dc54e95fcaf24128758da4d23c793954/integration-test/src/test/java/uk/gov/gchq/gaffer/integration/AbstractStoreIT.java#L186-L189

The calls to `tearDown` at the end of the tests in this class are redundant and can be removed entirely as JUnit is already calling this method after every test method. I'm not sure why it is being called at the start of the method either, as it will have just been called at the end of the previous test. No other test methods in the codebase do this.

r32575: `tearDown` is called by 2 test methods so would need to be put in a `@AfterEach` block instead/as well?

GCHQDeveloper314: I was getting mixed up between `@AfterAll` and `@AfterEach`. Adding an `@AfterEach` is required if we want to ensure it always runs, even if the assertion fails.

However, quite why this class needs to call `tearDown` or `setup` is unclear. Calling `tearDown` simply sets the graph to `null` and `setup` then reconfigures it. Before each test is run `setup` is called anyway, so why set it to `null` and then call `setup` if `setup` was already called before the method was run?

For now I think doing as you have suggested is the best approach. Improving this class to remove the apparently redundant `tearDown` or `setup` should be a separate issue.

GCHQDeveloper314 (approved):

t92549 (approved):

t92549:
Have removed the labels as usually these only go onto issues

Issue #2711, opened 2022-07-14
Add capability to run tests on Windows


t92549:
Currently, PRs that affect Windows must be manually tested, e.g: https://github.com/gchq/Gaffer/pull/2709.

Instead, the matrix in the build stage of the ci workflow could be changed to build on [both](https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs#using-a-matrix-strategy) ubuntu and Windows:
https://github.com/gchq/Gaffer/blob/580cf2bb045a8a8fee5803597310243b0cd08b50/.github/workflows/continuous-integration.yaml#L37-L44

Even better, this could be done using a variable to conditionally run it on Windows depending on some outside factor, for example if the branch name contains the string "windows" or "release-branch". This would mean tests are only run on windows specific PRs or before releases, instead of all the time. This is because Windows is not a huge priority for us right now and we don't want tests to take longer.

t92549:
Closed by https://github.com/gchq/Gaffer/pull/2716

PR #2716, opened 2022-07-22
Gh-2711: conditionally run tests on windows

lb324567:


lb324567:
Commit b0df840 should pass, as its only running the tests on Ubuntu (because of the keywords it's looking for)
Commit dd6a883 should fail because of issues with the tests running on windows, but does show that if the keywords are in the branch name the tests will run on windows.

GCHQDeveloper314:
I've just raised #2717 to cover the remaining Windows test failures. However, because it is not a priority issue, I don't think we should run this new Windows variant of CI for releases yet.
The next release is not planned to include the Windows fixes. If this PR were merged ahead of that release, the CI would fail when run on the release. We don't really want the CI to routinely fail on releases as this is misleading.

Also, under #2717 there will be extra changes required to the CI for Windows only (to install appropriate Hadoop libraries). If the release trigger were to remain disabled for now, it could then be enabled when #2717 is fixed (as there would be no more failures after this point).

t92549:
> I've just raised #2717 to cover the remaining Windows test failures. However, because it is not a priority issue, I don't think we should run this new Windows variant of CI for releases yet. The next release is not planned to include the Windows fixes. If this PR were merged ahead of that release, the CI would fail when run on the release. We don't really want the CI to routinely fail on releases as this is misleading.
> 
> Also, under #2717 there will be extra changes required to the CI for Windows only (to install appropriate Hadoop libraries). If the release trigger were to remain disabled for now, it could then be enabled when #2717 is fixed (as there would be no more failures after this point).

So for now, this should only run when "windows" is in the branch name?

GCHQDeveloper314:
> 
> So for now, this should only run when "windows" is in the branch name?

Yes that's the idea, this way we can create branches and PRs to work on the Windows problems, but the releases will continue to build as usual and will pass. Then once the existing Windows issues are fixed, we can enable it for releases as well.

From a practical perspective, we could have `release-disabled-for-now` instead of `release` so that there's only a minor edit required to enable, that way this PR could be merged pretty much as it is, instead of taking the `release` bit out.

lb324567:
> > So for now, this should only run when "windows" is in the branch name?
> 
> Yes that's the idea, this way we can create branches and PRs to work on the Windows problems, but the releases will continue to build as usual and will pass. Then once the existing Windows issues are fixed, we can enable it for releases as well.
> 
> From a practical perspective, we could have `release-disabled-for-now` instead of `release` so that there's only a minor edit required to enable, that way this PR could be merged pretty much as it is, instead of taking the `release` bit out.

I've made the change to use release-disabled-for-now as you suggested :)

GCHQDeveloper314 (approved):

GCHQDeveloper314:
Because the branch for this PR contains the word 'windows', the tests are run on Windows and they fail because Windows isn't fully supported yet. This PR is ok to be merged despite the CI showing a failure.

t92549 (approved):

PR #2727, opened 2022-08-08
FederatedStore Alpha4 staging: pt2 Chain and Output

GCHQDev404:


store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/operation/handler/impl/FederatedNoOutputHandler.java
```diff
@@ -0,0 +1,58 @@
...
+     * @param context   the operation chain context, containing the user who executed the operation
+     * @param store     the {@link uk.gov.gchq.gaffer.federatedstore.FederatedStore} the operation should be run on.
+     * @return null, no output.
+     * @throws OperationException thrown if the operation fails
+     */
+    @Override
+    public Void doOperation(final PAYLOAD operation, final Context context, final Store store) throws OperationException {
+        FederatedOperation<Object, Void> federatedOperation = getFederatedOperation(operation);
+
+        Object ignore = store.execute(federatedOperation, context);
```
r32575, line 52: Could you remove `ignore`?

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/operation/handler/impl/FederatedOperationHandler.java
```diff
@@ -0,0 +1,114 @@
...
+
+    private Iterable getAllGraphResults(final FederatedOperation<INPUT, OUTPUT> operation, final Context context, final FederatedStore store) throws OperationException {
+        try {
+            List<Object> results;
+            final Collection<Graph> graphs = getGraphs(operation, context, store);
+            results = new ArrayList<>(graphs.size());
+            for (final Graph graph : graphs) {
+
+                final Operation updatedOp = FederatedStoreUtil.updateOperationForGraph(operation.getUnClonedPayload(), graph);
+                if (null != updatedOp) {
```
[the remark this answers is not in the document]
r32575: Or `if (nonNull(updatedOp))`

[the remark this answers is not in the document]
GCHQDev404: I'm out of practice with long hand.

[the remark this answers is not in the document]
GCHQDev404:
```suggestion
                if (updatedOp != null) {
```

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/operation/handler/impl/FederatedNoOutputHandler.java
GCHQDev404: Its to make it very clear and obvious, there is "something" being returned but it is deliberately meant to be ignored.

r32575: Ah yeah, I believe you said about this before...

GCHQDeveloper314: I was thinking the same - that this was redundant. I wouldn't have thought having this value sitting around doing nothing will impact on performance, so it can stay - although I wouldn't do this.

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/operation/handler/impl/FederatedOperationHandler.java
```diff
@@ -0,0 +1,114 @@
...
+
+    private Iterable getAllGraphResults(final FederatedOperation<INPUT, OUTPUT> operation, final Context context, final FederatedStore store) throws OperationException {
+        try {
+            List<Object> results;
+            final Collection<Graph> graphs = getGraphs(operation, context, store);
+            results = new ArrayList<>(graphs.size());
+            for (final Graph graph : graphs) {
+
+                final Operation updatedOp = FederatedStoreUtil.updateOperationForGraph(operation.getUnClonedPayload(), graph);
+                if (updatedOp != null) {
```
GCHQDeveloper314, line 62: If `updatedOp` is `null` then `results` remains an empty `ArrayList`, is that correct? Seems odd that if `updatedOp` exists but is not an instance of `Output`, then `null` is added to the results, but this isn't done if there's no `updatedOp`.

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/operation/handler/impl/FederatedOperationHandler.java
```diff
@@ -0,0 +1,114 @@
...
+                            throw new OperationException(FederatedStoreUtil.createOperationErrorMsg(operation, graph.getGraphId(), e), e);
+                        }
+                    }
+                }
+            }
+
+            return results;
+        } catch (
+                final Exception e) {
+            throw new OperationException(ERROR_WHILE_RUNNING_OPERATION_ON_GRAPHS, e);
```
GCHQDeveloper314, line 83: Isn't this only needed for line 61? It looks like that is the only line that can throw anything and isn't part of the other try/catch. As the message is `ERROR_WHILE_RUNNING_OPERATION_ON_GRAPHS` it looks like this is intended to cover only line 61.

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/operation/handler/impl/FederatedOperationHandler.java
```diff
@@ -0,0 +1,114 @@
...
+
+            //Reduce
+            for (final Object resultFromAGraph : resultsFromAllGraphs) {
+                rtn = mergeFunction.apply(resultFromAGraph, rtn);
+            }
+
+            return rtn;
+        } catch (final Exception e) {
+            String message = e.getMessage();
+            throw new OperationException(String.format("Error while merging results. %s", isNull(message) ? "" : message), e);
```
GCHQDeveloper314, line 103: Why is the ternary operator used here? What's wrong with just using `e.getMessage()` directly? Seems to me that a message of `null` would be more useful than nothing.
In any case, [`Objects.toString`](https://docs.oracle.com/javase/8/docs/api/java/util/Objects.html#toString-java.lang.Object-java.lang.String-) can be used to make this more concise and easier to read.
```suggestion
            throw new OperationException(String.format("Error while merging results. %s", toString(e.getMessage(),"")), e);
```

store-implementation/federated-store/src/test/java/uk/gov/gchq/gaffer/federatedstore/FederatedStoreTestUtil.java
```diff
@@ -0,0 +1,189 @@
...
+    public static final String FORMAT_VALUE_STRING = "value%s";
+    public static final String VALUE_1 = value(1);
+    public static final String VALUE_2 = value(2);
+    static final String CACHE_SERVICE_CLASS_STRING = "uk.gov.gchq.gaffer.cache.impl.HashMapCacheService";
+    static final Set<String> GRAPH_AUTHS_ALL_USERS = ImmutableSet.of(ALL_USERS);
+    public static final String INTEGER = "integer";
+
+    private FederatedStoreTestUtil() {
+        //no instance
+    }
```
GCHQDeveloper314, line 94: Is this needed?

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/operation/handler/impl/FederatedOperationHandler.java
GCHQDev404: I think the reasoning was/is:

If there is no updated Op. 
It has now View which is legal for that graph.
The operation is not performed on that graph. 
nothing is done, nothing is repainted to results.

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/operation/handler/impl/FederatedOperationHandler.java
GCHQDev404: Its BIG catch for everything and anything (like dumb nullpointers) that will happen now or later to identify concisely where and what stage of federation something went wrong.


there are two parts "getting results from graphs" and "merging"

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/operation/handler/impl/FederatedOperationHandler.java
GCHQDev404: Nullpointers don't have message. appending "null" to a upstream message provides no value.


Sadly looks like you must have "Objects.toString" for scoping. but otherwise yeah I can use that.

store-implementation/federated-store/src/test/java/uk/gov/gchq/gaffer/federatedstore/FederatedStoreTestUtil.java
GCHQDev404: The utility class is blocked from having an instance created, the private constructor is not accessible.
comment shows its deliberate choice that is empty.

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/operation/handler/impl/FederatedNoOutputHandler.java
t92549, line 52: Could this just be `store.execute(federatedOperation, context);`?

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/operation/handler/impl/FederatedOperationHandler.java
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    private Object mergeResults(final Iterable resultsFromAllGraphs, final FederatedOperation operation, final FederatedStore store) throws OperationException {
+        try {
+            Object rtn = null;
+
+            //TODO FS map of merge
+            final BiFunction mergeFunction = nonNull(operation.getMergeFunction()) ? operation.getMergeFunction() : store.getDefaultMergeFunction();
+
+            //Reduce
```
t92549, line 95: May be too late but wondering if this could've used `Stream.reduce`

store-implementation/federated-store/src/test/java/uk/gov/gchq/gaffer/federatedstore/operation/handler/impl/FederatedOperationChainHandlerTest.java
```diff
@@ -46,19 +45,23 @@
...
+import static uk.gov.gchq.gaffer.federatedstore.FederatedStoreTestUtil.GRAPH_ID_ACCUMULO_WITH_ENTITIES;
+import static uk.gov.gchq.gaffer.federatedstore.util.FederatedStoreUtil.getFederatedOperation;
+import static uk.gov.gchq.gaffer.federatedstore.util.FederatedStoreUtil.getHardCodedDefaultMergeFunction;
 
 public class FederatedOperationChainHandlerTest {
 
-    public static final String GRAPH_IDS = String.format("%s,%s", PredefinedFederatedStore.ACCUMULO_GRAPH_WITH_ENTITIES, PredefinedFederatedStore.ACCUMULO_GRAPH_WITH_EDGES);
+    public static final String GRAPH_IDS = String.format("%s,%s", GRAPH_ID_ACCUMULO_WITH_ENTITIES, GRAPH_ID_ACCUMULO_WITH_EDGES);
 
-    private final Element[] elements = new Element[] {
```
t92549, line 61: Does checkstyle not enforce space before `{`?

store-implementation/federated-store/src/test/java/uk/gov/gchq/gaffer/federatedstore/operation/handler/impl/FederatedOperationChainHandlerTest.java
```diff
@@ -99,18 +102,21 @@ public void shouldHandleChainWithoutSpecialFederation() throws OperationExceptio
         final Context context = new Context();
 
         final OperationChain<Iterable<? extends Element>> opChain = new OperationChain.Builder()
-                .first(new GetAllElements.Builder()
-                        // Ensure the elements are returned form the graphs in the right order
-                        .option(KEY_OPERATION_OPTIONS_GRAPH_IDS, GRAPH_IDS)
+                .first(new FederatedOperation.Builder()
```
t92549, line 105: I know this is the new way, but is there a test so that the old way still works? (Ask for a GetAllElements which gets wrapped)?

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/operation/handler/impl/FederatedNoOutputHandler.java
t92549: Sorry I duplicated this question

store-implementation/federated-store/src/test/java/uk/gov/gchq/gaffer/federatedstore/operation/handler/impl/FederatedOperationChainHandlerTest.java
GCHQDev404: probably, this won't compile until the end and then that's a problem for checkstyle.

store-implementation/federated-store/src/test/java/uk/gov/gchq/gaffer/federatedstore/operation/handler/impl/FederatedOperationChainHandlerTest.java
GCHQDev404: I think...... the standard IT tests covers that.

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/operation/handler/impl/FederatedNoOutputHandler.java
GCHQDev404: > I was thinking the same - that this was redundant. I wouldn't have thought having this value sitting around doing nothing will impact on performance, so it can stay - although I wouldn't do this.

The value falls out of scope almost immediately, the java compiler if smart enough (dun oh) may ignore it.

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/operation/handler/impl/FederatedOperationHandler.java
GCHQDev404: instead of the loop?

t92549 (approved):

PR #2731, opened 2022-08-10
V2 alpha federated store alpha4 pr pt3

GCHQDev404:


t92549 (approved):

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/FederatedStore.java
```diff
@@ -334,46 +329,68 @@ public Set<StoreTrait> getTraits() {
...
         }
+        return rtn;
+    }
+
+    private String getKeyForProcessedFedStore() {
+        return FEDERATED_STORE_PROCESSED + id;
+    }
 
+    private boolean addFedStoreId(final Operation operation, final String optionKey) {
+        boolean rtn = false;
```
GCHQDeveloper314, line 367: I know this was already like this, but why it is `rtn` and not something more descriptive. It looks like it could be called `hasOperationOrPayloadPreexistingFedStoreId`?

GCHQDev404: For a "do" method which is returning a byproduct like a `isSuccess` or similar like this method. **I agree**. Its not clear what the byproduct is meant to be at a glance. so I can change this.
I think this method started life out as two methods a "do" `addID` and "get" `didIDexist` and was joined up.

for "get/create/make/fetch" methods the name of the method should be descriptive enough to know what is being returned. So having a `rtn` as a blanket standard variable name across most/all methods is clean clutter free way of writing methods. When reading methods with `rtn` it is very easy to identify the object you probably actually only care about and can easily see when it is being made/created/modified amongst a crowd of transient variables names which may be similar in noise. In the below example it should be clear what the method is getting, so having a variable called "magicSpell" might get lost amongst "stars", "magic", "phrases" and such.

``` java
public String getSuperMagicIncantationSpellPhrase(String magic, String phases, String[] stars){
    String rtn = getStartingPhrase()  //the only object you care about;
    String magic = // noise noise
    for(stars){
        if(magic){
              rtn =+ // Ah something is happening to what care about
        }
    }
    //something something
  
    return rtn; //send it!
}
```

PR #2737, opened 2022-08-16
Gh-2614: Enable alternative authentication to Accumulo using Kerberos

### LINKED ISSUE #2614: Enable alternative authentication to Accumulo
Gaffer authenticates to Accumulo using stored username & password. 
There is now a requirement to offer a choice of authentication either using Kerberos or the existing username/password. 

(https://github.com/gchq/Gaffer/issues/2348 previously enquired about setting up Gaffer to authenticate to Accumulo using Kerberos)

This needs to work with Accumulo 2. If possible also make it work for Accumulo 1.

GCHQDeveloper314:
# Related Issue

- Resolve #2614

### DIFF
--- a/store-implementation/accumulo-store/src/main/java/uk/gov/gchq/gaffer/accumulostore/utils/TableUtils.java
+++ b/store-implementation/accumulo-store/src/main/java/uk/gov/gchq/gaffer/accumulostore/utils/TableUtils.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016-2020 Crown Copyright
+ * Copyright 2016-2022 Crown Copyright
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/store-implementation/accumulo-store/src/main/java/uk/gov/gchq/gaffer/accumulostore/utils/TableUtils.java
+++ b/store-implementation/accumulo-store/src/main/java/uk/gov/gchq/gaffer/accumulostore/utils/TableUtils.java
@@ -258,7 +258,7 @@ public final class TableUtils {
             KerberosToken token = new KerberosToken();
             Connector conn = instance.getConnector(token.getPrincipal(), token);
             return conn;
-        } catch (final AccumuloException | AccumuloSecurityException | IOException e) {
+        } catch (final Exception e) {
             throw new StoreException("Failed to create accumulo connection (using Kerberos authentication)", e);
         }
     }
@@ -292,7 +292,7 @@ public final class TableUtils {
     public static Authorizations getCurrentAuthorizations(final Connector connection) throws StoreException {
         try {
             return connection.securityOperations().getUserAuthorizations(connection.whoami());
-        } catch (final AccumuloException | AccumuloSecurityException e) {
+        } catch (final Exception e) {
             throw new StoreException(e.getMessage(), e);
         }
     }
--- a/store-implementation/accumulo-store/src/main/java/uk/gov/gchq/gaffer/accumulostore/utils/TableUtils.java
+++ b/store-implementation/accumulo-store/src/main/java/uk/gov/gchq/gaffer/accumulostore/utils/TableUtils.java
@@ -240,14 +240,14 @@ public final class TableUtils {
         final Instance instance = new ZooKeeperInstance(instanceName, zookeepers);
         try {
             // Configure Hadoop UGI to use Kerberos (if it's not already)
-            if (!UserGroupInformation.isSecurityEnabled()){
+            if (!UserGroupInformation.isSecurityEnabled()) {
                 Configuration conf = new Configuration();
                 conf.set("hadoop.security.authentication", "kerberos");
                 conf.set("hadoop.security.authorization", "true");
                 UserGroupInformation.setConfiguration(conf);
             }
             // If already logged in with Keytab, then check if it needs renewal, else do initial login
-            if (UserGroupInformation.isLoginKeytabBased()){
+            if (UserGroupInformation.isLoginKeytabBased()) {
                 UserGroupInformation.getCurrentUser().checkTGTAndReloginFromKeytab();
                 LOGGER.info("Already logged into Kerberos, TGT rechecked for principal '{}'", UserGroupInformation.getCurrentUser().getUserName());
             } else {
--- a/store-implementation/accumulo-store/src/main/java/uk/gov/gchq/gaffer/accumulostore/utils/TableUtils.java
+++ b/store-implementation/accumulo-store/src/main/java/uk/gov/gchq/gaffer/accumulostore/utils/TableUtils.java
@@ -246,15 +246,13 @@ public final class TableUtils {
                 conf.set("hadoop.security.authorization", "true");
                 UserGroupInformation.setConfiguration(conf);
             }
-            // If already logged in with Keytab, then check if it needs renewal, else do initial login
+            // If already logged in with Keytab, then check if ticket needs renewal, else do initial login
             if (UserGroupInformation.isLoginKeytabBased()) {
                 UserGroupInformation.getCurrentUser().checkTGTAndReloginFromKeytab();
-                LOGGER.info("Already logged into Kerberos, TGT rechecked for principal '{}'", UserGroupInformation.getCurrentUser().getUserName());
+                LOGGER.debug("Already logged into Kerberos, TGT rechecked for principal '{}'", UserGroupInformation.getCurrentUser().getUserName());
             } else {
                 LOGGER.info("Attempting Kerberos login with principal '{}' & keytab path '{}'", principal, keytabPath);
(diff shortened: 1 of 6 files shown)


t92549 (approved):

p3430233 (approved):

store-implementation/accumulo-store/src/main/java/uk/gov/gchq/gaffer/accumulostore/utils/TableUtils.java
```diff
@@ -216,7 +221,63 @@ public static Connector getConnector(final String instanceName, final String zoo
...
+        final Instance instance = new ZooKeeperInstance(instanceName, zookeepers);
+        try {
+            // Configure Hadoop UGI to use Kerberos (if it's not already)
+            if (!UserGroupInformation.isSecurityEnabled()) {
+                Configuration conf = new Configuration();
+                conf.set("hadoop.security.authentication", "kerberos");
+                conf.set("hadoop.security.authorization", "true");
+                UserGroupInformation.setConfiguration(conf);
+            }
+            // If already logged in with Keytab, then check if ticket needs renewal, else do initial login
```
t92549, line 250: Is a connection made for every Operation or just when the graph is created?

GCHQDeveloper314: It is done for every operation. However the Kerberos renewal/login is only done as often as is required. If the connection was only created once then there would be no easy way to ensure it did not expire. This isn't an issue for user/pass which cannot expire like this.

r32575 (approved):

store-implementation/accumulo-store/src/main/java/uk/gov/gchq/gaffer/accumulostore/utils/TableUtils.java
```diff
@@ -216,7 +221,63 @@ public static Connector getConnector(final String instanceName, final String zoo
...
+     * @throws StoreException failure to create an accumulo connection
+     */
+    public static Connector getConnectorKerberos(final String instanceName, final String zookeepers,
+                                                 final String principal, final String keytabPath) throws StoreException {
+        final Instance instance = new ZooKeeperInstance(instanceName, zookeepers);
+        try {
+            // Configure Hadoop UGI to use Kerberos (if it's not already)
+            if (!UserGroupInformation.isSecurityEnabled()) {
+                Configuration conf = new Configuration();
+                conf.set("hadoop.security.authentication", "kerberos");
```
GCHQDev404, line 246: Use a string constant.

store-implementation/accumulo-store/src/main/java/uk/gov/gchq/gaffer/accumulostore/utils/TableUtils.java
```diff
@@ -216,7 +221,63 @@ public static Connector getConnector(final String instanceName, final String zoo
...
+                UserGroupInformation.getCurrentUser().checkTGTAndReloginFromKeytab();
+                LOGGER.debug("Already logged into Kerberos, TGT rechecked for principal '{}'", UserGroupInformation.getCurrentUser().getUserName());
+            } else {
+                LOGGER.info("Attempting Kerberos login with principal '{}' & keytab path '{}'", principal, keytabPath);
+                UserGroupInformation.loginUserFromKeytab(principal, keytabPath);
+            }
+            KerberosToken token = new KerberosToken();
+            Connector conn = instance.getConnector(token.getPrincipal(), token);
+            return conn;
+        } catch (final AccumuloException | AccumuloSecurityException | IOException e) {
```
GCHQDev404, line 261: make this final Exception

store-implementation/accumulo-store/src/main/java/uk/gov/gchq/gaffer/accumulostore/utils/TableUtils.java
```diff
@@ -216,7 +221,63 @@ public static Connector getConnector(final String instanceName, final String zoo
         try {
             return instance.getConnector(userName, new PasswordToken(password));
         } catch (final AccumuloException | AccumuloSecurityException e) {
```
GCHQDev404, line 223: make this final Exception

store-implementation/accumulo-store/src/test/java/uk/gov/gchq/gaffer/accumulostore/AccumuloPropertiesTest.java
```diff
@@ -54,6 +56,86 @@ public void shouldMergeAccumuloJsonModulesAndDeduplicate() {
         assertEquals(SketchesJsonModules.class.getName() + "," + TestCustomJsonModules1.class.getName(), modules);
     }
 
+
+    @Test
+    public void getEnabledKerberosShouldReturnFalseIfSetFalse() {
+        // Given
+        final AccumuloProperties props = new AccumuloProperties();
+        final boolean ENABLE_KERBEROS = false;
```
GCHQDev404, line 64: wrong case.

store-implementation/accumulo-store/src/main/java/uk/gov/gchq/gaffer/accumulostore/utils/TableUtils.java
```diff
@@ -216,7 +221,63 @@ public static Connector getConnector(final String instanceName, final String zoo
...
+     * an accumulo instance using the provided parameters
+     *
+     * @param instanceName the instance name
+     * @param zookeepers   the zoo keepers
+     * @param principal    the Kerberos principal
+     * @param keytabPath   the path to the Kerberos keytab
+     * @return A connection to an accumulo instance
+     * @throws StoreException failure to create an accumulo connection
+     */
+    public static Connector getConnectorKerberos(final String instanceName, final String zookeepers,
```
GCHQDev404, line 239: Where is the test for this connection? is it possible to test ?

GCHQDev404 (changes requested):

store-implementation/accumulo-store/src/main/java/uk/gov/gchq/gaffer/accumulostore/utils/TableUtils.java
GCHQDeveloper314: It is not possible to test with a Unit test. This is because `MiniAccumuloCluster` doesn't support Kerberos. It is covered by the integration tests.

store-implementation/accumulo-store/src/test/java/uk/gov/gchq/gaffer/accumulostore/AccumuloPropertiesTest.java
GCHQDeveloper314: Do you mean it being all caps? This style was copied form the preexisting test under these new ones - `shouldSetProperties`.

t92549: I think he is suggesting only static variables should be all caps.
Should be automatically caught by check style though.

GCHQDev404: Yeah, why isn't check style picking up on this, is something broken?

store-implementation/accumulo-store/src/main/java/uk/gov/gchq/gaffer/accumulostore/utils/TableUtils.java
GCHQDeveloper314: I would, except that would add an extra two lines, 200 odd lines above here and wouldn't provide any benefit because these values are not going to be used anywhere else in the class or in Gaffer.

store-implementation/accumulo-store/src/main/java/uk/gov/gchq/gaffer/accumulostore/utils/TableUtils.java
GCHQDeveloper314: This was already like this. It could be changed in the same way as the others, but this PR is not the place to do it as the code hasn't been introduced or modified. There are another 5 instances in this class where the least upper bound (most specific supertype of the types in question) of a multi-catch clause is Exception. If we want to replace these (not worth doing) then it should come under a separate code clean up issue.

store-implementation/accumulo-store/src/test/java/uk/gov/gchq/gaffer/accumulostore/AccumuloPropertiesTest.java
GCHQDeveloper314: Fixed, agree it should be picked up by checkstyle. The other instances of all caps in this class can be fixed in a different PR - probably one which adds checking for this to checkstyle.

GCHQDev404 (approved):

PR #2742, opened 2022-08-19
FederatedStore Alpha4 staging: pt6-8

GCHQDev404:


store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/operation/handler/impl/FederatedGetTraitsHandler.java
```diff
@@ -45,35 +37,15 @@
...
-            }
-
-            return rtn.keySet();
+            return (Set<StoreTrait>) store.execute(
+                    new FederatedOperation.Builder()
+                            .op(operation)
+                            .mergeFunction(new CollectionIntersect<Object>())
+                            .graphIds(getDeprecatedGraphIds(operation)) // deprecate this line.
+                            .build(), context);
         } catch (final Exception e) {
```
GCHQDeveloper314, line 49: Why `Exception`? The thrown message is about traits. Doesn't `FederatedOperation.Builder` throw something more specific than generic `Exception`?

GCHQDev404: It is to catch Any and All exception from execution of the operation not just the builder.
This is to breadcrumb the stack trace for Users and Devs.

PR #2752, opened 2022-09-16
Fix package structure for traffic tests

### LINKED ISSUE #2736: Fix inconsistent uk.gov.gchq.gaffer.traffic package
There are some tests, under [example/road-traffic/road-traffic-rest/src/test/java/uk.gov.gchq.gaffer.traffic/](https://github.com/gchq/Gaffer/blob/580cf2bb045a8a8fee5803597310243b0cd08b50/example/road-traffic/road-traffic-rest/src/test/java/uk.gov.gchq.gaffer.traffic/RoadTrafficRestApiITs.java), which are in a directory where the package is `.` delimited rather than `/`. This should be fixed as it can break some IDEs/builders like VSCode, but also because it is inconsistent.

t92549:
# Related Issue

- Resolve #2736


GCHQDeveloper314 (approved):
Unfortunate that it is requiring a copyright update here. If you want you could not do it and merge anyway. This won't get picked up in future if you did this as it only compares with `v2-alpha`.

PR #2756, opened 2022-09-22
Gh 2755 fix operation output type

### LINKED ISSUE #2755: Add more testing around Operation Output type
There is currently not much testing around the `getOutputTypeReference` function that must be overridden by all Output Operations. This can lead to errors that go undiscovered, for example, in this new Operation, the function returns null:
https://github.com/gchq/Gaffer/blob/9a66133ec8f0abe5a3758989f3f5ecdd73f6deeb/core/operation/src/main/java/uk/gov/gchq/gaffer/operation/impl/export/localfile/ExportToLocalFile.java#L96-L98
This may seem logical as the Operation does not return anything in reality, but when null is used here, it causes a null pointer exception in the getOperationDetails rest endpoint:
https://github.com/gchq/Gaffer/blob/9a66133ec8f0abe5a3758989f3f5ecdd73f6deeb/rest-api/common-rest/src/main/java/uk/gov/gchq/gaffer/rest/model/OperationDetail.java#L92
I suggest that this Operation is fixed, as well as better handling added for a null pointer exception.
To stop this from happening in the future, however, more tests should be added around getOutputTypeReference as well as the getOperationDetails rest endpoint.

t92549:
# Related Issue

- Resolve #2755

### DIFF
--- a/core/operation/src/main/java/uk/gov/gchq/gaffer/operation/impl/export/localfile/ExportToLocalFile.java
+++ b/core/operation/src/main/java/uk/gov/gchq/gaffer/operation/impl/export/localfile/ExportToLocalFile.java
@@ -24,7 +24,7 @@ import org.apache.commons.lang3.exception.CloneFailedException;
 import uk.gov.gchq.gaffer.commonutil.Required;
 
 import uk.gov.gchq.gaffer.operation.export.ExportTo;
-import uk.gov.gchq.gaffer.operation.serialisation.TypeReferenceImpl;
+import uk.gov.gchq.gaffer.operation.serialisation.TypeReferenceImpl.IterableString;
 import uk.gov.gchq.koryphe.Since;
 import uk.gov.gchq.koryphe.Summary;
 
@@ -95,7 +95,7 @@ public class ExportToLocalFile implements ExportTo<Iterable<? extends String>> {
 
     @Override
     public TypeReference<Iterable<? extends String>> getOutputTypeReference() {
-        return new TypeReferenceImpl.IterableString();
+        return new IterableString();
     }
 
 
--- a/core/operation/src/main/java/uk/gov/gchq/gaffer/operation/impl/export/localfile/ExportToLocalFile.java
+++ b/core/operation/src/main/java/uk/gov/gchq/gaffer/operation/impl/export/localfile/ExportToLocalFile.java
@@ -24,6 +24,7 @@ import org.apache.commons.lang3.exception.CloneFailedException;
 import uk.gov.gchq.gaffer.commonutil.Required;
 
 import uk.gov.gchq.gaffer.operation.export.ExportTo;
+import uk.gov.gchq.gaffer.operation.serialisation.TypeReferenceImpl;
 import uk.gov.gchq.koryphe.Since;
 import uk.gov.gchq.koryphe.Summary;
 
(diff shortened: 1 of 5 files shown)


core/operation/src/main/java/uk/gov/gchq/gaffer/operation/impl/export/localfile/ExportToLocalFile.java
```diff
@@ -24,6 +24,7 @@
 import uk.gov.gchq.gaffer.commonutil.Required;
 
 import uk.gov.gchq.gaffer.operation.export.ExportTo;
+import uk.gov.gchq.gaffer.operation.serialisation.TypeReferenceImpl;
```
GCHQDeveloper314, line 27: We can import the inner class `IterableString` directly.
```suggestion
import uk.gov.gchq.gaffer.operation.serialisation.TypeReferenceImpl.IterableString;
```

core/operation/src/main/java/uk/gov/gchq/gaffer/operation/impl/export/localfile/ExportToLocalFile.java
```diff
@@ -83,23 +84,23 @@ public void setKey(final String key) {
...
-    public void setInput(final Iterable<String> input) {
+    public void setInput(final Iterable<? extends String> input) {
         this.input = input;
     }
 
     @Override
-    public TypeReference<Iterable<String>> getOutputTypeReference() {
-        return null;
+    public TypeReference<Iterable<? extends String>> getOutputTypeReference() {
+        return new TypeReferenceImpl.IterableString();
```
GCHQDeveloper314, line 98: We can use `IterableString` directly.
```suggestion
    public IterableString getOutputTypeReference() {
        return new IterableString();
```

GCHQDeveloper314 (changes requested):

core/operation/src/main/java/uk/gov/gchq/gaffer/operation/impl/export/localfile/ExportToLocalFile.java
t92549: I do prefer your way but did it more for consistency, around 65 places in Gaffer do it this way but 3 the better way. I suppose doing the better change is more meaningful than the consistent one..

core/operation/src/main/java/uk/gov/gchq/gaffer/operation/io/Output.java
```diff
@@ -47,19 +47,31 @@ default Class<?> getOutputClass() {
...
         final TypeReference<O> outputType = getOutputTypeReference();
-        if (null != outputType) {
-            Type type = outputType.getType();
-            if (type instanceof ParameterizedType) {
-                type = ((ParameterizedType) type).getRawType();
-            }
-            if (type instanceof Class) {
-                outputClass = (Class) type;
-            }
+        if (null == outputType) {
```
lb324567, line 50:
```suggestion
        if (outputType == null) {
```

core/operation/src/main/java/uk/gov/gchq/gaffer/operation/io/Output.java
```diff
@@ -47,19 +47,31 @@ default Class<?> getOutputClass() {
...
+            outputClass = (Class) type;
         }
 
         return outputClass;
     }
 
+    @JsonIgnore
+    default Type getOutputType() {
+        final TypeReference<O> outputType = getOutputTypeReference();
+        if (null == outputType) {
```
lb324567, line 68:
```suggestion
        if (outputType == null) {
```

lb324567 (changes requested):
Can we make sure we avoid using 'null == '

t92549:
> Can we make sure we avoid using 'null == '

I have changed this as I agree I prefer `x == null`, I think here I had just copy and pasted from an equivalent method.
If this is a code style thing we are going to fail PRs over though perhaps it should be automated.

GCHQDeveloper314 (approved):

lb324567 (approved):

PR #2785, opened 2022-10-23
gh-2632 federated store remove graph and delete accumulo table

### LINKED ISSUE #2632: FederatedStore RemoveGraphDeleteAccumuloTableHandler
Handler to force Accumulo to drop the table.

---
Update!
 - New Federated Operation: `RemoveAndDeleteGraph`
 - New Federated Handler: `RemovAndDeleteGraphHandler`
-- extend RemoveGraphHandler.
-- Before executing super.doOperation(), each graph is asked if it supports `TearDownOperation`
-- If true sends `TearDownOperation`
-- then executes super.doOperation()
- New Operation: `DeleteDataOperation`
 - New Handler: `DeleteDataHandler`
 -- AccumuloStore implements this handler to do Accumulo specific drop tables.
 -- MapStore implements a token gesture of clear map (MapStore data deleted anyway when mapstore is removed and scope is lost)

GCHQDev404:
This handler will require MANUAL testing against an Accumulo instance.
Remember to add the Handler to the FederatedStore config.

# Related Issue

- Resolve #2632


GCHQDev404:
blocked waiting on #2769

GCHQDev404:
blocked waiting on #2631

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/operation/handler/impl/FederatedRemoveGraphDeleteAccumuloTableHandler.java
```diff
@@ -0,0 +1,133 @@
...
+
+/**
+ * A handler for RemoveGraph operation for the FederatedStore.
+ * And delete associated Accumulo Tables.
+ * <p>
+ *
+ * @see FederatedStore
+ * @see RemoveGraph
+ */
+public class FederatedRemoveGraphDeleteAccumuloTableHandler extends FederatedRemoveGraphHandler {
```
t92549, line 52: Could this be done as a DeleteGraph operation that is implemented by MapStore and AccumuloStore, and the FederatedStore can pass that to the relevant store? As part of remove graph?

GCHQDev404: A Mapstore just stops existing.
AccumuloStore possibly it deletes the table, but the Accumlo instance continues to live.
This is possibly useful.

Federation would need to send `DeleteGraph` operation to all stores. Only AccumuloStore would support the operation.

Federation would still require a second Handler. I think.

*additional comment taken offline

GCHQDev404:
Approach requires a re-work see ticket.

PR #2799, opened 2022-10-28
Gh 2632 federated store remove graph and delete accumulo table

### LINKED ISSUE #2632: FederatedStore RemoveGraphDeleteAccumuloTableHandler

GCHQDev404:
# Related Issue

- Resolve #2632


store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/operation/RemoveAndDeleteGraph.java
```diff
@@ -0,0 +1,28 @@
...
+
+import com.fasterxml.jackson.annotation.JsonPropertyOrder;
+
+
+@JsonPropertyOrder(value = {"class", "graphId"}, alphabetic = true)
+public class RemoveAndDeleteGraph extends RemoveGraph {
+    /*
+     * This class is completely empty.
+     * This highlights need for Operation.class to be final with field "String id = operationName"
+     */
```
GCHQDeveloper314, line 27: Can this be fixed with some other method? An empty class like this is not ideal.

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/operation/handler/impl/FederatedRemoveAndDeleteGraphHandler.java
```diff
@@ -0,0 +1,62 @@
...
+
+    @Override
+    public Boolean doOperation(final RemoveGraph operation, final Context context, final Store store) throws OperationException {
+        try {
+            //Get the graph before they are removed.
+            final List<Graph> graphsToRemove = ((FederatedStore) store).getGraphs(context.getUser(), Collections.singletonList(operation.getGraphId()), operation);
+
+            //Ask graphs to delete all data.
+            for (final Graph graph : graphsToRemove) {
+                if (graph.isSupported(DeleteAllData.class)) {
```
GCHQDeveloper314, line 49: If someone has specifically asked for remove **and** delete, might it be useful to warn them if the relevant store doesn't support deletion? Otherwise they might just assume data was deleted when it wasn't because the operation ran successfully.
One option could be to log a warning, but as this might not make its way to the user immediately, you could instead throw an exception if the operation is not supported. This could state that the store doesn't support deleting and therefore the RemoveGraph operation should be used instead.

store-implementation/accumulo-store/src/main/java/uk/gov/gchq/gaffer/accumulostore/operation/handler/DeleteAllDataHandler.java
```diff
@@ -0,0 +1,58 @@
...
+
+    @Override
+    public Object doOperation(final DeleteAllData operation, final Context context, final Store store) throws OperationException {
+        try {
+            final String removeId = store.getGraphId();
+            try {
+                Connector connection = getConnector((AccumuloProperties) store.getProperties());
+                if (connection.tableOperations().exists(removeId)) {
+                    connection.tableOperations().offline(removeId);
+                    connection.tableOperations().delete(removeId);
```
GCHQDeveloper314, line 45: I don't see any test/ITs for this.

GCHQDeveloper314 (changes requested):
The only comment which definitely requires changes is the lack of tests for the Accumulo Store table deletion.

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/operation/RemoveAndDeleteGraph.java
GCHQDev404: The comment signifies the Empty class is deliberate.
Do you want the "this highlights" part to be removed?

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/operation/handler/impl/FederatedRemoveAndDeleteGraphHandler.java
GCHQDev404: Logging will never make its way to the user.

store-implementation/accumulo-store/src/main/java/uk/gov/gchq/gaffer/accumulostore/operation/handler/DeleteAllDataHandler.java
GCHQDev404: I don't think it  be possible to test this easily. 
An accumulo instance will need to be set up. 
This will need to be manually tested by developers.

GCHQDeveloper314: Should be possible to test with the MiniAccumuloStore as used by other tests. This doesn't require a real/full Accumulo instance/cluster.

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/operation/handler/impl/FederatedRemoveAndDeleteGraphHandler.java
GCHQDeveloper314: I'm thinking more about awareness for an admin. As it is written currently, if a store doesn't support deleting then the data will be left undeleted and admins would need to know this was a possibility and manually check for this as there's no logging to inform about this.
The easiest option would be to **fail if the store doesn't support deletion**. There's no reason to use this operation over the remove operation if store doesn't support deletion.

t92549 (approved):

GCHQDeveloper314:
Is it correct that when there's no federation involved, this situation can now occur as a side effect of the `DeleteAllData` operation?

1. An instance of Gaffer is created with a new schema/properties to use an Accumulo Store.
2. Gaffer is started and connects to the Accumulo Store. The table does not yet exist, so it is created.
3. User adds data to the graph, etc.
4. User calls the `DeleteAllData` operation. The Accumulo table is removed.
5. User then tries to add data. **I think this could fail because the table doesn't exist anymore and is only created when Gaffer starts up.** Restarting Gaffer would probably cause the table to be recreated.

I don't think this would happen with federation as once the graph is removed from federation it no longer exists anywhere and there is no way to send any operations to it. If the user wants to use it again they'd have to add it which would trigger table creation.

t92549:
This can happen, but that's part of the reason that this `DeleteAllData` operation isn't enabled by default for all Accumulo stores.

GCHQDev404:
> Is it correct that when there's no federation involved, this situation can now occur as a side effect of the `DeleteAllData` operation?
> 
> 1. An instance of Gaffer is created with a new schema/properties to use an Accumulo Store.
> 2. Gaffer is started and connects to the Accumulo Store. The table does not yet exist, so it is created.
> 3. User adds data to the graph, etc.
> 4. User calls the `DeleteAllData` operation. The Accumulo table is removed.
> 5. User then tries to add data. **I think this could fail because the table doesn't exist anymore and is only created when Gaffer starts up.** Restarting Gaffer would probably cause the table to be recreated.
> 
> I don't think this would happen with federation as once the graph is removed from federation it no longer exists anywhere and there is no way to send any operations to it. If the user wants to use it again they'd have to add it which would trigger table creation.

This Solo AccumuloStore probably shouldn't have DeleteAllData enabled by the system admin.
The Solo AccumuloStore should be shutdown after a DeleteAllData if it was enabled.
restarting the Solo AccumuloStore would re-connect and remake a new empty table in Accumulo.

store-implementation/accumulo-store/src/main/java/uk/gov/gchq/gaffer/accumulostore/operation/handler/DeleteAllDataHandler.java
GCHQDev404: This test does no work.
the last assert throws a null pointer.
I do not think testing is easily possible.

```java
FederatedStoreTestUtil.resetForFederatedTests();

//Make a FederatedStore
final FederatedStore federatedStore = new FederatedStore();
federatedStore.initialise(GRAPH_ID_TEST_FEDERATED_STORE, new Schema(), new FederatedStoreProperties());

//Add MiniAccumulo
federatedStore.execute(new AddGraph.Builder()
.graphId(DELETE_ALL_DATA_GRAPH)
.schema(loadSchemaFromJson(SCHEMA_EDGE_BASIC_JSON))
.storeProperties(loadAccumuloStoreProperties(PROPERTIES_ACCUMULO_STORE_PROPERTIES))
.build(), contextBlankUser());

//Make a Local MiniAccumulo with Table collision.
final Store directToMiniAccumulo = MiniAccumuloStore.createStore(DELETE_ALL_DATA_GRAPH, loadSchemaFromJson(SCHEMA_EDGE_BASIC_JSON), loadAccumuloStoreProperties(PROPERTIES_ACCUMULO_STORE_PROPERTIES));

//Local Mini is empty
assertThat(directToMiniAccumulo.execute(new GetAllElements(), contextBlankUser()))
.containsExactly();

//Add Element to FederatedStore
federatedStore.execute(new AddElements.Builder().input(edgeBasic()).build(), contextBlankUser());

//Check the element is Local MiniAccumulo confirming table collision.
assertThat(directToMiniAccumulo.execute(new GetAllElements(), contextBlankUser()))
.containsExactly(edgeBasic());

//Federated RemoveGraph with no Deletion of data or tables
federatedStore.execute(new RemoveGraph.Builder().graphId(DELETE_ALL_DATA_GRAPH).build(), contextBlankUser());

//Confirm Local mini is still working and connected to the table.
assertThatNoException()
.isThrownBy(() -> directToMiniAccumulo.execute(new GetAllElements(), contextBlankUser()));
```

GCHQDeveloper314 (approved):
The lack of complete testing has been raised as #2801 and doesn't need to be addressed in this PR.

lb324567 (approved):

PR #2808, opened 2022-11-08
Gh-2773: Tests to document deprecated graphId behaviour

### LINKED ISSUE #2773: FederatedStore: Examine old style operation chain with graphIds
Write or find a test that covers the behaviour of old style Operation chains.

`chain{ [ 1stOp{}, 2ndOp{}, 3rdOp{options:[graphId:A]}, 4thOp{} ]}`
I suspect the whole operationChain gets sent to all graphs, breaking change.

`chain{ [ 1stOp{}, 2ndOp{}, 3rdOp{}, 4thOp{} ], options:[graphId:A]}`
I suspect the whole operationChain get sent to graph A, not breaking change.

t92549:
# Related Issue

- Resolve #2773

### DIFF
--- a/store-implementation/federated-store/src/test/java/uk/gov/gchq/gaffer/federatedstore/FederatedStoreGraphIDsTest.java
+++ b/store-implementation/federated-store/src/test/java/uk/gov/gchq/gaffer/federatedstore/FederatedStoreGraphIDsTest.java
@@ -1,0 +1,248 @@
+/*
+ * Copyright 2022 Crown Copyright
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package uk.gov.gchq.gaffer.federatedstore;
+
+import org.junit.jupiter.api.AfterAll;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+import uk.gov.gchq.gaffer.data.element.Element;
+import uk.gov.gchq.gaffer.data.element.Entity;
+import uk.gov.gchq.gaffer.federatedstore.operation.AddGraph;
+import uk.gov.gchq.gaffer.federatedstore.operation.FederatedOperation;
+import uk.gov.gchq.gaffer.operation.OperationChain;
+import uk.gov.gchq.gaffer.operation.impl.DiscardOutput;
+import uk.gov.gchq.gaffer.operation.impl.Limit;
+import uk.gov.gchq.gaffer.operation.impl.add.AddElements;
+import uk.gov.gchq.gaffer.operation.impl.export.set.ExportToSet;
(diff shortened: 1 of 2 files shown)


store-implementation/federated-store/src/test/java/uk/gov/gchq/gaffer/federatedstore/FederatedStoreGraphIDsTest.java
```diff
@@ -0,0 +1,248 @@
...
+                                .build())
+                        .build(), contextBlankUser());
+
+        // Then
+        assertThat(results).hasSize(2);
+        assertThat((Iterable<Entity>) results).containsExactlyInAnyOrder(EXPECTED_ENTITY, thirdEntity);
+    }
+
+    @Test
+    public void shouldNotCorrectlySelectGraphUsingGraphIdsOptionOnOperationChain() throws Exception {
```
GCHQDeveloper314, line 230: I'm confused by the name of this test.

> should Not Correctly Select Graph Using GraphIds Option On OperationChain

Does this mean the same as "should Not Select Graph Using GraphIds Option On OperationChain" or "should Ignore Select Graph Using GraphIds Option On OperationChain"?

store-implementation/federated-store/src/test/java/uk/gov/gchq/gaffer/federatedstore/FederatedStoreGraphIDsTest.java
```diff
@@ -0,0 +1,248 @@
...
+                                .input(EXPECTED_ENTITY)
+                                .build())
+                        .then(new Limit<>(10))
+                        .option(DEPRECATED_GRAPHIDS_OPTION, GRAPH_ID_A)
+                        .build(), contextBlankUser());
+
+        // Then
+        // Pre Gaffer 2.0 this would have sent the whole chain to graph A
+        // Now, the chain's graphIds option is ignored as every individual
+        // operation is wrapped in a FederatedOperation
```
GCHQDeveloper314, line 244: So the addition of the option doesn't have any effect here because it is attached to the chain and not to the individual operation inside it. The important bit here being that `DEPRECATED_GRAPHIDS_OPTION` only has any effect when used on individual operations and not on a chain?
Could have some more info on that, seeing as there is a comment.

store-implementation/federated-store/src/test/java/uk/gov/gchq/gaffer/federatedstore/FederatedStoreGraphIDsTest.java
t92549: Both! Wording is subjective and awkward here, I am trying to convey that the graphIds options is ignored when set on an OperationChain and that this is weird/incorrect/breaking behaviour but we are accepting it for now

store-implementation/federated-store/src/test/java/uk/gov/gchq/gaffer/federatedstore/FederatedStoreGraphIDsTest.java
t92549: Needs to be documented for sure

GCHQDev404 (dismissed):

store-implementation/federated-store/src/test/java/uk/gov/gchq/gaffer/federatedstore/FederatedStoreGraphIDsTest.java
GCHQDeveloper314: I see. Could this be changed to "should ignore historic graphIds option on operationchain", or something similar which implies why this is not doing anything. Alternatively, could add a comment under the line explaining this.

GCHQDev404 (approved):

GCHQDeveloper314 (approved):

PR #2810, opened 2022-11-10
gh-2809 federatedstore fix incorrect test refracting

### LINKED ISSUE #2809: FederatedStore fix incorrect test refracting
FederatedStoreGraphVisibilityTest was incorrectly refactored in previous alpha4 test.

GCHQDev404:
# Related Issue

- Resolve #2809

### DIFF
--- a/core/store/src/main/java/uk/gov/gchq/gaffer/store/library/GraphLibrary.java
+++ b/core/store/src/main/java/uk/gov/gchq/gaffer/store/library/GraphLibrary.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2017-2020 Crown Copyright
+ * Copyright 2017-2022 Crown Copyright
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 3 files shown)


t92549 (dismissed):
Nice refactor of `GraphLibrary` and tests make a lot more sense now

core/store/src/main/java/uk/gov/gchq/gaffer/store/library/GraphLibrary.java
```diff
@@ -327,44 +327,47 @@ private void nullCheck(final String graphId, final Schema schema, final StorePro
 
 
     public Schema resolveSchema(final Schema schema, final List<String> parentSchemaIds) {
-        Schema resultSchema = null;
+        Schema.Builder rtn = new Schema.Builder();
```
GCHQDeveloper314, line 330: Why is this `rtn`. It's not being returned, one of it's methods run and the result of that is returned. Even if it was being returned, `rtn` is not a good name.

PR #2811, opened 2022-11-10
gh-1834 federated store different visability properties

### LINKED ISSUE #1834: Unable to federate over 2 graphs that have different visibility properties
If you create 2 graphs in the FederatedStore with different visibility property names ("visibility1" and "visibility2") then you cannot run a single operation over both graphs at the same time.

This is because, in order to federated over the 2 graphs at the same time, we need to be able to merge the schemas together. But, because the visibilityProperty value is different the schemas cannot be merged together.

A workaround is to run 2 separate queries and merge the results, GetElements on graph1 and then GetElements on graph2. This way the queries are done separately and the schemas don't need to be merged together. It isn't very pretty but currently it is the only way of doing it. For example:

```
GetElements(options:{graphId: graph1})
ExportToSet
GetElements(options:{graphId: graph2})
ExportToSet
DiscardOutput
GetSetExport
```

GCHQDev404:
…improve temp merge maps control of visibility due to possible aggregation that could happen in the wild.

# Related Issue

- Resolve #1834


t92549:
Superseded by https://github.com/gchq/Gaffer/pull/2818

PR #2818, opened 2022-11-11
Gh 1834: federated store multiple visibility properties

### LINKED ISSUE #1834: Unable to federate over 2 graphs that have different visibility properties

t92549:
# Related Issue

- Resolve #1834


store-implementation/federated-store/src/test/java/uk/gov/gchq/gaffer/federatedstore/FederatedStoreVisibilityTest.java
```diff
@@ -0,0 +1,339 @@
...
+
+        // When
+        Iterable<? extends Element> aggregatedResults = (Iterable<? extends Element>) federatedGraph.execute(
+                new FederatedOperation.Builder()
+                        .op(new GetAllElements())
+                        .build(), USER);
+        Iterable<? extends Element> concatenatedResults = (Iterable<? extends Element>) federatedGraph.execute(
+                new FederatedOperation.Builder()
+                        .op(new GetAllElements())
+                        .mergeFunction(getDefaultMergeFunction())
```
GCHQDeveloper314, line 213: I thought the default merge function was not concatenation? Also if the merge function is being set to the default, then what does this do? I would have thought the default merge function value would already be the default merge function.
If what's happening is the actual default is not what is returned by "getDefault", then that seems to be needlessly confusing.

t92549: The `getDefaultMergeFunction` returns the basic merge function that will be used by an Operation that has not been configured to use any other merge function. This is the `ConcatenateListMergeFunction`.

However, `GetAllElements` has had a merge function set by it in `FederatedStore` which overrides the default. It uses the `ApplyViewToElementsFunction`:
https://github.com/gchq/Gaffer/blob/2b1f8d235c38099b9e269565553896a7e2a64356/store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/FederatedStore.java#L158

GCHQDev404 (dismissed):

store-implementation/federated-store/src/test/java/uk/gov/gchq/gaffer/federatedstore/FederatedStoreVisibilityTest.java
GCHQDeveloper314: That makes sense, however as this test is intended to use `ConcatenateListMergeFunction`, then wouldn't it be better to use it directly, instead of using a method which happens to return it? That way it would be clear at a glance what the merge function being used was, without having to dig into the codebase.

t92549: I think @GCHQDev404 advised against this? But I agree it would be more readable

GCHQDev404 (approved):

store-implementation/federated-store/src/test/java/uk/gov/gchq/gaffer/federatedstore/FederatedStoreVisibilityTest.java
t92549: Checked with @GCHQDev404 and have changed to `ConcatenateListMergeFunction`

GCHQDeveloper314 (approved):

PR #2827, opened 2022-11-21
Gh-2779: FederatedStore remove handlers on RemoveGraph

### LINKED ISSUE #2779: FederatedStore FederatedAddGraphHandlerParent appending supported operations to FederatedStore
Within the `FederatedAddGraphHandlerParent` there is logic to add Handlers for operations that are unsupported by the FederatedStore, WHEN adding a new graph that DOES support the new operation, which the FederatedStore did not previously support.

This Logic should be explored to confirm that it is still required.

THEN it likely in `RemoveGraph` when the LAST graph which supports an operation is removed, it should remove the supported operations from the FederatedStore.

Additional info
The handlers which are added this way to support these new operations are just hollow gateway handlers which typically just wrap the operation in a `FederatedOperation`. Attention should  be paid to if the `AdminConfiguredMergeMap` requires some kind of updating within the FederatedStore constructor.

t92549:
# Related Issue

- Resolve #2779

### DIFF
--- a/core/store/src/main/java/uk/gov/gchq/gaffer/store/Store.java
+++ b/core/store/src/main/java/uk/gov/gchq/gaffer/store/Store.java
@@ -174,6 +174,7 @@ import uk.gov.gchq.koryphe.util.ReflectionUtil;
 
 import java.util.ArrayList;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.LinkedHashMap;
 import java.util.LinkedHashSet;
 import java.util.List;
@@ -580,7 +581,7 @@ public abstract class Store {
      * @return a collection of all the supported {@link Operation}s.
      */
     public Set<Class<? extends Operation>> getSupportedOperations() {
-        return operationHandlers.keySet();
+        return new HashSet<>(operationHandlers.keySet());
     }
 
     public Set<Class<? extends Operation>> getNextOperations(final Class<? extends Operation> operation) {
--- a/store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/FederatedStore.java
+++ b/store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/FederatedStore.java
@@ -135,6 +135,7 @@ public class FederatedStore extends Store {
     private List<String> storeConfiguredGraphIds;
     @JsonTypeInfo(use = JsonTypeInfo.Id.CLASS, property = "class")
     private Map<String, BiFunction> storeConfiguredMergeFunctions;
+    private Set<Class<? extends Operation>> externallySupportedOperations = new HashSet<>();
 
     @JsonCreator
     public FederatedStore(@JsonProperty("customPropertiesAuths") final Set<String> customPropertiesAuths,
@@ -313,6 +314,21 @@ public class FederatedStore extends Store {
         return asAdmin
                 ? graphStorage.remove(graphId, user, this.getProperties().getAdminAuth())
                 : graphStorage.remove(graphId, user);
+    }
+
+    /**
+     * @return a collection of all the {@link Operation}s supported by subgraphs.
+     */
+    public Set<Class<? extends Operation>> getExternallySupportedOperations() {
+        return new HashSet<>(externallySupportedOperations);
+    }
+
+    public void removeExternallySupportedOperation(final Class<? extends Operation> operation) {
+        externallySupportedOperations.remove(operation);
+    }
+
+    public void addExternallySupportedOperation(final Class<? extends Operation> operation) {
+        externallySupportedOperations.add(operation);
     }
 
     /**
--- a/store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/operation/handler/impl/FederatedRemoveGraphHandler.java
+++ b/store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/operation/handler/impl/FederatedRemoveGraphHandler.java
@@ -18,11 +18,16 @@ package uk.gov.gchq.gaffer.federatedstore.operation.handler.impl;
 
 import uk.gov.gchq.gaffer.federatedstore.FederatedStore;
 import uk.gov.gchq.gaffer.federatedstore.operation.RemoveGraph;
+import uk.gov.gchq.gaffer.graph.Graph;
+import uk.gov.gchq.gaffer.graph.GraphSerialisable;
+import uk.gov.gchq.gaffer.operation.Operation;
 import uk.gov.gchq.gaffer.operation.OperationException;
 import uk.gov.gchq.gaffer.store.Context;
 import uk.gov.gchq.gaffer.store.Store;
 import uk.gov.gchq.gaffer.store.operation.handler.OutputOperationHandler;
 
+import java.util.HashSet;
+import java.util.Set;
 
 /**
  * A handler for RemoveGraph operation for the FederatedStore.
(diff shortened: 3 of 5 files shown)


t92549:
This implementation is a bit messy, but necessary without a large rework to how operation handlers and supported operations are implemented by the FederatedStore.

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/operation/handler/impl/FederatedRemoveGraphHandler.java
```diff
@@ -35,10 +40,34 @@
...
+                        store.addOperationHandler(removedOperation, null);
+                        ((FederatedStore) store).removeExternallySupportedOperation(removedOperation);
+                    }
+                }
+            }
+        }
+        return removed;
+    }
+
+    private Set<Class<? extends Operation>> getSupportedOperations(final Context context, final FederatedStore store) {
```
GCHQDev404, line 65:
```suggestion
    private Set<Class<? extends Operation>> getSupportedOperationsFromSubGraphs(final Context context, final FederatedStore store) {
```

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/operation/handler/impl/FederatedRemoveGraphHandler.java
```diff
@@ -35,10 +40,34 @@
...
+            removed = ((FederatedStore) store).remove(operation.getGraphId(), context.getUser(), operation.isUserRequestingAdminUsage());
         } catch (final Exception e) {
             throw new OperationException("Error removing graph: " + operation.getGraphId(), e);
         }
+        if (removed) {
+            Set<Class<? extends Operation>> updatedSupportedOperations = getSupportedOperations(context, (FederatedStore) store);
+            if (!supportedOperations.equals(updatedSupportedOperations)) {
+                supportedOperations.removeAll(updatedSupportedOperations);
+                for (final Class<? extends Operation> removedOperation : supportedOperations) {
+                    if (((FederatedStore) store).getExternallySupportedOperations().contains(removedOperation)) {
```
GCHQDev404, line 55: else? what happens?
is that illegal state? do we care?

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/FederatedStore.java
```diff
@@ -135,6 +135,7 @@ public class FederatedStore extends Store {
     private List<String> storeConfiguredGraphIds;
     @JsonTypeInfo(use = JsonTypeInfo.Id.CLASS, property = "class")
     private Map<String, BiFunction> storeConfiguredMergeFunctions;
+    private Set<Class<? extends Operation>> externallySupportedOperations = new HashSet<>();
```
GCHQDev404, line 138: You can delete this ;) Its not used anywhere.
You set and get it, but it has not effect on `graph.getSupported operations()` 

you check for the diff before and after removing a graph. you update `supportedOperations` and you also update this superfluous object which isn't used anywhere.

GCHQDev404 (changes requested):

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/operation/handler/impl/FederatedRemoveGraphHandler.java
t92549: The if statement is needed because it checks that the operations that are no longer supported by subgraphs are not operations that should still be supported by the federated store itself.
Otherwise if you have 2 subgraphs, an Accumulo store and a federated store, and you removed the sub federated store, all of the federated operation handlers would be deleted. This is why the `getExternallySupportedOperations` flag is required.

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/FederatedStore.java
t92549: See my other comment below, this is required

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/operation/handler/impl/FederatedRemoveGraphHandler.java
t92549: So to answer "else what happens?": If one store is added and then removed, the `GetElementsHandler` would be detected as removed from the sub graphs, but we see that this is not in the `externallySupportedOperations` as the federated store can support this itself, so we do not delete the handler.

GCHQDev404: Can you review this logic.

1) AddGraph(FedStore) // has AddGraph
2) FederatedAddGraphParentHandler -> addExternallySupportedOperations() // AddGraph is added.
3) RemoveGraphHandler line44 -> getSupportedOperations // includes AddGraph from (1)
4) RemoveGraphHandler line51 -> getSupportedOperations // does NOT include AddGraph
5) line 52 not equal
6) line 53 supported list has AddGraph (it wasn't removed)
7) line 55 externallySupportedOperations contains AddGraph 
8) line 56 remove AddGraph from current FederatedStore.

FederatedStore no supports AddGraphs.

t92549: See new test `shouldKeepFederatedStoreSupportedOperations`

t92549: `AddGraph` is never added to `externallySupportedOperations` because in the [`FederatedAddGraphHandlerParent::addGenericHandler`](https://github.com/gchq/Gaffer/blob/dce22c074b9fcc89cb6f4b2c9cc8e193217e8066/store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/operation/handler/FederatedAddGraphHandlerParent.java#L94) it only gets added when `!store.isSupported(supportedOperation)`

GCHQDev404: thanks. I think I'm satisfied by this.

GCHQDev404 (approved):

core/store/src/main/java/uk/gov/gchq/gaffer/store/Store.java
```diff
@@ -580,7 +581,7 @@ public boolean isSupported(final Class<? extends Operation> operationClass) {
      * @return a collection of all the supported {@link Operation}s.
      */
     public Set<Class<? extends Operation>> getSupportedOperations() {
-        return operationHandlers.keySet();
+        return new HashSet<>(operationHandlers.keySet());
```
GCHQDeveloper314, line 584: Why is this needed? It seems unrelated to this PR.

GCHQDeveloper314 (approved):

core/store/src/main/java/uk/gov/gchq/gaffer/store/Store.java
t92549: I wanted to safely return a copy of the keys so I could modify the result without modifying the underlying operationHandlers map

Issue #2829, opened 2022-11-23
ConcatenateListMergeFunction should return lazy iterables


t92549:
**Describe the new feature you'd like**
The federated store's `ConcatenateListMergeFunction` should return results in a lazy iterable.

**Why do you want this feature?**
Pre v2.0 the results were in a lazy iterable so this functionality should be retained.

**Additional context**
Currently, the koryphe function `ToList` is used to turn all results into iterables. However, this stops the results from being returned lazily as they are read into a list:
https://github.com/gchq/Gaffer/blob/2a33be8bea67e3864b7872f5dfb6f8dcee95e98b/store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/util/ConcatenateListMergeFunction.java#L36
As well as this, `Lists.newArrayList` has the same issue: https://github.com/gchq/Gaffer/blob/2a33be8bea67e3864b7872f5dfb6f8dcee95e98b/store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/util/ConcatenateListMergeFunction.java#L39
Koryphe functions do not really need to be used at all here in fact, and `IterableConcat` can be replaced by directly using a `ChainedIterable`, which is what used to happen and what is happening indirectly anyway.

If the proposed changes were implemented, then https://github.com/gchq/koryphe/pull/286 can be closed.

As well as these functionality changes, there should be proper javadoc and tests added to this class to ensure it works as expected and the expected behaviour is documented.

Additionally, I would argue the class should be renamed to `ConcatenateMergeFunction` and `ConcatenateSetMergeFunction` should be deleted.
This is because results are concatenated whether they are in a list or not, and returned in an Iterable. The results are not necessarily returned in a list. If a user specifically wanted a list or set return type, they could use ToList or ToSet in their operation chain.

GCHQDev404:
ChainedIterable requires an Array of Iterables to be given to the constructor.

You don't know what is being returned from the individual graphs, until it's returned and they may not be Iterables.

3 graphs could return:
3 nulls
3 iterates of elements
3 schemas
3 strings
3 lists of strings.

Also you need to have all the iterates available as the same time which doesn't work the same way as an BiFunction being looped and updated.

t92549:
[...]

PR #2831, opened 2022-11-24
Gh-2829: Refactor ConcatenateListMergeFunction to use lazy iterables

GCHQDeveloper314:
Remove creation of list/array/collection except for when object is not already an iterable.

### DIFF
--- a/store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/util/ConcatenateListMergeFunction.java
+++ b/store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/util/ConcatenateListMergeFunction.java
@@ -16,27 +16,29 @@
 
 package uk.gov.gchq.gaffer.federatedstore.util;
 
-import com.google.common.collect.Lists;
 import org.apache.commons.lang3.builder.HashCodeBuilder;
 
-import uk.gov.gchq.koryphe.impl.function.IterableConcat;
-import uk.gov.gchq.koryphe.impl.function.ToList;
+import uk.gov.gchq.koryphe.iterable.ChainedIterable;
 
 import java.util.Collections;
 import java.util.function.BiFunction;
 
-import static java.util.Objects.isNull;
 import static java.util.Objects.nonNull;
 
 public class ConcatenateListMergeFunction implements BiFunction<Object, Iterable<Object>, Iterable<Object>> {
 
     @Override
     public Iterable<Object> apply(final Object update, final Iterable<Object> state) {
-        //When update=null, Then this STOPS ToList returning an Iterable with a null in it.
-        final Iterable<Object> updateSafe = isNull(update) ? Collections.emptyList() : (Iterable<Object>) new ToList().apply(update);
-        return isNull(state)
-                ? updateSafe
-                : new IterableConcat<>().apply(Lists.newArrayList(updateSafe, state));
+        final Iterable<Object> updateSafe;
+        // When update is null, don't return an Iterable containing a null.
+        if (update == null) {
+            updateSafe = Collections.emptyList();
+        } else if (!(update instanceof Iterable)) {
+            updateSafe = Collections.singletonList(update);
+        } else {
+            updateSafe = (Iterable<Object>) update;
+        }
+        return (state == null) ? updateSafe : new ChainedIterable<>(updateSafe, state);
     }
 
     @Override


t92549:
Marked as draft as tests and javadoc to be added

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/util/ConcatenateSetMergeFunction.java
```diff
@@ -1,53 +0,0 @@
...
-
-public class ConcatenateSetMergeFunction implements BiFunction<Object, Iterable<Object>, Iterable<Object>> {
-
-    @Override
-    public Iterable<Object> apply(final Object update, final Iterable<Object> state) {
-        //When update=null, Then this STOPS ToSet returning an Iterable with a null in it.
-        final Iterable<Object> updateSafe = isNull(update) ? Collections.emptySet() : (Iterable<Object>) new ToSet().apply(update);
-        return isNull(state)
-                ? updateSafe
-                : new IterableConcat<>().apply(Sets.newHashSet(updateSafe, state));
```
GCHQDev404, line 39: How do I get a concatenation of of type set ?
(distinct ?)

t92549: If you need a Set return type, put a `ToSet` operation after the one that was merged with this.
It will always be inefficient with a lazy iterable though as you need to empty the iterable into a Set to make it distinct.

store-implementation/federated-store/src/test/java/uk/gov/gchq/gaffer/federatedstore/util/ConcatenateMergeFunctionTest.java
```diff
@@ -0,0 +1,113 @@
...
+        // When
+        Object update = STRING;
+        Iterable<Object> state = Collections.singletonList(STRING);
+        Iterable<Object> results = mergeFunction.apply(update, state);
+
+        // Then
+        assertThat(results)
+                .containsExactly(STRING, STRING);
+    }
+}
```
GCHQDev404, line 113: Can you add a test for
update = iterable[null]
state = iterable[null]

assertThat(output).containsExactly(null, null)

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/util/ConcatenateMergeFunction.java
```diff
@@ -16,27 +16,41 @@
...
-        return isNull(state)
-                ? updateSafe
-                : new IterableConcat<>().apply(Lists.newArrayList(updateSafe, state));
+        final Iterable<Object> updateSafe;
+        // When update is null, don't add an Iterable containing a null.
+        if (update == null) {
+            updateSafe = Collections.emptyList();
+        } else if (!(update instanceof Iterable)) {
+            // When update is not Iterable, wrap it in a List.
+            updateSafe = Collections.singletonList(update);
```
GCHQDev404, line 47: So an array becomes an Iterable with one object of type array in it?
`Iterable{[1,2,3]}`

I think the old would return the contents of the array as the iterable.
`Iterable{1,2,3}`

GCHQDev404: See: `ToList` line 44

GCHQDeveloper314: If by the old one, you mean the way it was done prior to the introduction of this class, then yes. @t92549 and I had a look at this and the old merge did not convert a result to an iterable if it wasn't one already ([see this code](https://github.com/gchq/Gaffer/blob/9421d9dfed70cfa14c475b8f8d9144589bc135ab/store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/operation/handler/impl/FederatedOperationChainHandler.java#L66)).

That behaviour changed with the introduction of this class. The current behaviour seems better.

Old behaviour with results from a graph A & B, where one graph doesn't return an `Iterable` for some reason:
`A -> Array [1,2,3]` _not an `Iterable`_
`B -> Iterable {4,5,6}`
When merged returns `Iterable{[1,2,3],{4,5,6}}`, these are not actually merged because a non `Iterable` was found (see the `break` in the for loop).

New (current) behaviour with those same results returns `Iterable{[1,2,3],4,5,6}`. The array is still an array, but the second set of results have been merged better.

An example where the current behaviour is more obviously better is this:
`A -> Int 1` _not an `Iterable`_
`B -> Iterable {4,5,6}`
When merged returns `Iterable{1,4,5,6}`, old way would have given `Iterable{1,{4,5,6}}`.

This is why it was considered useful to add tests, although they don't seem to cover any of those use cases yet.

We also thought it was uncommon for results to not be an `Iterable`, so these differences are unlikely to be noticed.

GCHQDev404: I'm talking in reference to class your replacing `ConcatenateListMergeFunction`

```java
final String[] s = new String[]{"one","two","three"};
//        final Iterable<Object> apply = new ConcatenateListMergeFunction().apply(s, null);
        final Iterable<Object> apply = new ConcatenateMergeFunction().apply(s, null);
        org.assertj.core.api.Assertions.assertThat(apply)
                .containsExactly("one","two","three");
```
                
Toggle between ConcatenateListMergeFunction (passes) and ConcatenateMergeFunction (fails).

GCHQDeveloper314: I have another look and you're correct, Arrays are now handled differently. I'll fix this. The way they were handled before this class was introduced was also different.

t92549: Added a test to verify it will work how @GCHQDev404 wants.
This will mean it behaves differently to Gaffer 1 though so perhaps another ticket should be made to compare before and after?

store-implementation/federated-store/src/test/java/uk/gov/gchq/gaffer/federatedstore/util/ConcatenateMergeFunctionTest.java
```diff
@@ -78,11 +79,42 @@ public void shouldReturnUpdateWrappedInIterableWhenStateNull() {
...
+    public void shouldReturnIntegerUpdateWrappedInIterableWhenStateNull() {
+        // Given
+        final ConcatenateMergeFunction mergeFunction = new ConcatenateMergeFunction();
+
+        // When
+        Object update = 25;
+        Iterable<Object> results = mergeFunction.apply(update, null);
+
+        // Then
+        assertThat(results)
```
t92549, line 96: `assertThat(results).hasSameClassAs(results)` - won't that always pass?

GCHQDeveloper314: You're right, it will. I was using this as a lazy way of doing `assertThat(results).isInstanceOf(Iterable.class)`, but as we're assigning to an object of that class, if it was anything else we'd have an exception thrown prior to the test.

I'll remove these asserts.

t92549 (approved):

GCHQDev404 (approved):

Issue #2764, opened 2022-10-06
FederatedStore: Review and test "isDisabledByDefault"


GCHQDev404:
**UPDATE: We have now decided that isDisabledByDefault will be removed from Gaffer. As such this ticket is now to delete isDisabledByDefault;**

Class: FederatedStoreDefaultGraphsTest
Add tests and examine the interaction between `isDisabledByDefault` and `StoreConfiguredGraphIds` List
Add tests and examine the interaction between `isDisabledByDefault` and `getGraphs` with Admin privilege.

---

Class: FederatedGraphStorage
Method: GetStream
Line: ~331 `.filter(entry -> !entry.getKey().isDisabledByDefault())`
Review the requirement, usage and test coverage of isDisabledByDefault.

----

Class: FederatedStore
Method: GetDefaultGraphs
line: ~ 549 `if (isNull(adminConfiguredDefaultGraphIdsCSV) || isAdminRequestingOverridingDefaultGraphs) {`
Review the interaction, usage and test coverage with `isDisabledByDefault`. is disabled logic preserved or is it required. 

---

Class: FederatedGraphStorageTest
Method: shouldGetDisabledGraphWhenGetAll
line: ~217
disabledByDefault access is used, but `getAll()` has auths parameter so when and why is it disabled?

---
Class: FederatedGraphStorageTest
Method: shouldNotGetDisabledGraphForAuthUserWhenNoIdsProvided
line: ~287
examine the use of `null` why is it only missing when `get(null)`

GCHQDev404:
I suspect that isDisabledByDefault is not required and can be deleted from Gaffer. due to `FederatedStore.storeConfiguredGraphsIds`.

lb324567:
Completed by #2834

PR #2834, opened 2022-12-01
Gh-2764: federated store review and test is disabled by default

r32575:


store-implementation/federated-store/src/test/java/uk/gov/gchq/gaffer/federatedstore/FederatedStoreConfiguredGraphIdsTest.java
```diff
@@ -0,0 +1,299 @@
...
+    void shouldFailConfiguredGraphIdsFileInvalid() {
+        assertThatExceptionOfType(IllegalArgumentException.class)
+                .isThrownBy(() -> getFederatedStoreFromProperties(FEDERATED_STORE_CONFIGURED_GRAPH_IDS_FAIL_FILE_INVALID_PROPERTIES))
+                .withMessageContaining("Could not initialise the store with provided arguments.")
+                .withStackTraceContaining("Cannot deserialize value of type `java.util.ArrayList<java.lang.Object>` from String value");
+    }
+
+    @Test
+    void shouldReturnSelectedGraphIdByDefaultFromJsonConfig()
+            throws Exception {
```
GCHQDeveloper314, line 100: Why is the throw on another line? It looks odd and there's plenty of space for it in the usual place.

store-implementation/federated-store/src/test/java/uk/gov/gchq/gaffer/federatedstore/FederatedStoreConfiguredGraphIdsTest.java
```diff
@@ -0,0 +1,299 @@
...
+        assertNotNull(result); // TODO: Fails as result == null
+        JsonAssert.assertEquals(selectedSchema.toJson(true), result.toJson(true));
+    }
+
+    private static FederatedStore getFederatedStoreFromJson(final String path) throws IOException, StoreException {
+        final FederatedStore federatedStore = loadFederatedStoreFrom(path);
+        federatedStore.initialise(GRAPH_ID_SELECTED, new Schema(), new FederatedStoreProperties());
+        addDefaultLibrary(federatedStore);
+        return federatedStore;
+    }
```
GCHQDeveloper314, line 232: This seems like quite a generic utility test method for federated stores. In fact, looking at the `FederatedStoreDefaultGraphsTest.java` file which was deleted in this PR, there's this in there:

```
import static uk.gov.gchq.gaffer.federatedstore.FederatedStoreTestUtil.loadFederatedStoreFrom;
...
FederatedStore defaultedFederatedStore = loadFederatedStoreFrom("configuredGraphIds.json");
```

This new method appears to be a duplicate of that method.

store-implementation/federated-store/src/test/java/uk/gov/gchq/gaffer/federatedstore/FederatedStoreConfiguredGraphIdsTest.java
```diff
@@ -0,0 +1,299 @@
...
+                .merge(STRING_SCHEMA)
+                .build();
+
+        federatedStore.getGraphLibrary().addSchema(SCHEMA_UNSELECTED, unselectedSchema);
+        addGraph(federatedStore, GRAPH_ID_UNSELECTED, SCHEMA_UNSELECTED, contextTestUser());
+
+        return unselectedSchema;
+    }
+
+    private static void addGraph(final FederatedStore federatedStore, final String graphId,
```
GCHQDeveloper314, line 289: There's a method in `FederatedStoreTestUtil` very similar to this. If that method doesn't quite work another could be added there instead.

store-implementation/federated-store/src/test/java/uk/gov/gchq/gaffer/federatedstore/FederatedStoreTest.java
```diff
@@ -707,7 +709,8 @@ public void shouldNotAllowOverridingOfKnownGraphInLibrary() throws Exception {
     @Test
     public void shouldFederatedIfUserHasCorrectAuths() throws Exception {
         // Given
-        store.addGraphs(GRAPH_AUTHS_ALL_USERS, null, false, new GraphSerialisable.Builder()
+        final AccessPredicate noAccessPredicate = new NoAccessPredicate();
+        store.addGraphs(GRAPH_AUTHS_ALL_USERS, null, false, null, noAccessPredicate, new GraphSerialisable.Builder()
```
GCHQDeveloper314, line 713: Why is the addition of `NoAccessPredicate` required? Is that because the previous method in use was the one including disabledByDefault?

GCHQDeveloper314 (changes requested):
All looks fine except for the addition of those test utility functions. Ideally the existing utility functions would be used or improved.

I'm also interested in what happens to an existing federated store which is configured to use default graph ids after this is merged. If the store had them configured already and was upgraded to this version then would it continue to work correctly? Presumable any graphs that were using the disabled behaviour would become visible?

store-implementation/federated-store/src/test/java/uk/gov/gchq/gaffer/federatedstore/FederatedStoreConfiguredGraphIdsTest.java
r32575: `FederatedStoreDefaultGraphsTest.java` was not deleted but renamed but it seems that git didn't pick up on this. Just wondering whether any other test class would use the test methods if they were added to the test util class. That is why I kept them in the `FederatedStoreConfiguredGraphIdsTest`

store-implementation/federated-store/src/test/java/uk/gov/gchq/gaffer/federatedstore/FederatedStoreTest.java
r32575: @lb324567, think this was refactored by yourself. Any idea?

store-implementation/federated-store/src/test/java/uk/gov/gchq/gaffer/federatedstore/FederatedStoreConfiguredGraphIdsTest.java
r32575: The 2 are different in the way they call `AddGraph`. One is a `FederatedSore` direct call and other uses an operation. One presents the schema and properties where as the other stores the schema and properties in the cache and gives a sting representation that is then used later to get the schemas/properties and merge them returning a single schema. So, if you wanted to amalgamate the 2 you would have to convert to a single format and provide more options to the util class. If you were to go down the `Operation` route then it includes adding an exception catch to 2 other methods.

I have gone down the `Operation` route. It functions as before but you have lost the direct call as a result. Is that important to you? I have not merged the changes yet. The new code:

```
public static void addGraphToAccumuloStore(final FederatedStore federatedStore, final String graphId,
                                               final boolean isPublic, final Schema schema)
        throws StorageException, OperationException {
    addGraph(federatedStore, graphId, isPublic,
            loadAccumuloStoreProperties(ACCUMULO_STORE_SINGLE_USE_PROPERTIES),
            schema, null, null, new Context(new User(UNKNOWN_USER_ID)));
}

public static void addGraph(final FederatedStore federatedStore, final String graphId,
                            final boolean isPublic, final StoreProperties storeProperties, final Schema schema,
                            final String parentPropertyIds, final List<String> parentSchemaIds,
                            final Context context)
        throws OperationException {
    federatedStore.execute(OperationChain.wrap(
            new AddGraph.Builder()
                    .graphId(graphId)
                    .storeProperties(storeProperties)
                    .schema(schema)
                    .parentSchemaIds(parentSchemaIds)
                    .parentPropertiesId(parentPropertyIds)
                    .isPublic(isPublic)
                    .build()), context);
}
```

I altered the method name slightly as it doesn't just add the graph.

store-implementation/federated-store/src/test/java/uk/gov/gchq/gaffer/federatedstore/FederatedStoreConfiguredGraphIdsTest.java
r32575: I added a method to create the Federated Store as there is now the loading from a properties as well as the loading from `JSON` method mentioned.

GCHQDeveloper314: Ah ok, shame it doesn't recognise that - I've seen this before when there are too many changes. I think doing `git mv` and then making a commit straight away with just the rename _might_ help this in future.
If the methods were already in this class before the rename then don't worry about changing that. I'd assumed they were new. If the are new then it would make more sense to put them in the utility class, but you're right that nothing else would be using them.

It might be better to have a new ticket for refactoring all federation tests to ensure there are no duplicating test methods and that everything is in the test utility class.

store-implementation/federated-store/src/test/java/uk/gov/gchq/gaffer/federatedstore/FederatedStoreTest.java
lb324567: I am not sure what happened here, as i don't remember making that change and it doesn't seem to be necessary. I'll undo this change.

GCHQDeveloper314 (approved):

Issue #2841, opened 2022-12-09
Upgrade Spotbugs configuration


GCHQDeveloper314:
**Describe the new feature you'd like**
Spotbugs is currently used but the configuration has not been changed from when Findbugs was being used. It should be updated and improved - as mentioned in #2722 . The POM configuration could also allow for the complimentary PMD plugin to be used.

**Why do you want this feature?**
To improve static analysis for the project.

GCHQDeveloper314:
The complimentary PMD plugin has been tested as part of this issue but it requires many minor changes to bring the codebase up to standard. To keep this issue manageable, any PMD plugin should be done under a separate issue and is now out of scope for this issue.

t92549:
Closed by https://github.com/gchq/Gaffer/pull/2842

PR #2842, opened 2022-12-16
Gh-2841: Update Spotbugs

GCHQDeveloper314:


core/common-util/src/main/java/uk/gov/gchq/gaffer/commonutil/OneOrMore.java
```diff
@@ -101,6 +102,7 @@ public boolean addAll(final Collection<? extends T> items) {
         return result;
     }
 
+    @SuppressFBWarnings(value = "BC_BAD_CAST_TO_ABSTRACT_COLLECTION", justification = "Cast to List is required to use a List specific remove method")
     public void removeAnyItem() {
         if (null == collection) {
             singleItem = null;
```
t92549, line 108: Maybe it could get put on two lines instead? and also check that we can cast to List?

core/graph/src/main/java/uk/gov/gchq/gaffer/graph/Graph.java
```diff
@@ -776,9 +777,11 @@ public Builder addSchema(final Path schemaPath) {
             if (null != schemaPath) {
                 try {
                     if (Files.isDirectory(schemaPath)) {
```
t92549, line 779: Would try with resources work if the if was before the try?

integration-test/src/main/java/uk/gov/gchq/gaffer/integration/generators/BasicElementGenerator.java
```diff
@@ -36,10 +36,13 @@ public class BasicElementGenerator implements OneToOneElementGenerator<DomainObj
...
+        Element result = null;
         if (domainObject instanceof EntityDomainObject) {
-            return entityGenerator._apply((EntityDomainObject) domainObject);
+            result = entityGenerator._apply((EntityDomainObject) domainObject);
+        } else if (domainObject instanceof EdgeDomainObject) {
+            result = edgeGenerator._apply((EdgeDomainObject) domainObject);
         }
 
-        return edgeGenerator._apply((EdgeDomainObject) domainObject);
+        return result;
```
t92549, line 46: Maybe we should throw if we get here instead of return null

integration-test/src/main/java/uk/gov/gchq/gaffer/integration/impl/loader/schemas/SchemaLoader.java
```diff
@@ -39,6 +40,7 @@ public interface SchemaLoader {
     String B = "B";
     String C = "C";
     String D = "D";
+    @SuppressFBWarnings(value = "MS_OOI_PKGPROTECT", justification = "Value is not being changed")
     String[] VERTEX_PREFIXES = new String[]{A, B, C, D};
```
t92549, line 44: Quite a few tests use this, could we reuse `AbstractStoreIT::VERTEX_PREFIXES`

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/operation/IFederationOperation.java
```diff
@@ -32,8 +32,8 @@ public interface IFederationOperation extends Operation {
     boolean isUserRequestingAdminUsage();
 
     @JsonGetter("userRequestingAdminUsage")
```
t92549, line 34: I think this was being done to ensure that FederationOperations did not show the `userRequestingAdminUsage` field in json when it was false, probably as this is default. I think a test should be added to assert this behaviour and another method should be used to ensure this passes. I can look at this next week

t92549 (commented):
Very nice additions!

core/graph/src/main/java/uk/gov/gchq/gaffer/graph/Graph.java
GCHQDeveloper314: Yes it looks like it would. Everything in the `try` is inside an `if` and which can probably be merged with the other one as well as being taken out.

integration-test/src/main/java/uk/gov/gchq/gaffer/integration/generators/BasicElementGenerator.java
GCHQDeveloper314: We could, it I think it would make the flow nicer as well. Do you know of a suitable Gaffer specific exception to use?

core/common-util/src/main/java/uk/gov/gchq/gaffer/commonutil/OneOrMore.java
GCHQDeveloper314: The `@SuppressFBWarnings` on two lines? Checking if we can cast to List won't affect the warning.

t92549: Yeah if it can maybe the line would be a bit less long

integration-test/src/main/java/uk/gov/gchq/gaffer/integration/generators/BasicElementGenerator.java
t92549: There is `GafferRuntimeException` but not sure what would be best to use

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/operation/IFederationOperation.java
GCHQDeveloper314: Changing to use `@JsonInclude(value = JsonInclude.Include.NON_DEFAULT)` removing the need for the `null` return and has the same effect. This is because globally `null` is not included already. This also allows the extra method to be removed entirely.

There is a test (in `FederatedOperationTest`) which unintentionally checks that this field (and also `skipFailedFederatedExecution`) is not present.

integration-test/src/main/java/uk/gov/gchq/gaffer/integration/impl/loader/schemas/SchemaLoader.java
GCHQDeveloper314: Value from `AbstractStoreIT` now used instead, this field has been removed and there's no longer a need to suppress this warning.

core/operation/src/main/java/uk/gov/gchq/gaffer/operation/impl/add/AddElementsFromKafka.java
```diff
@@ -160,7 +161,7 @@ public void setOptions(final Map<String, String> options) {
     @Override
     public ValidationResult validate() {
         final ValidationResult result = Operation.super.validate();
-        if (null != bootstrapServers && bootstrapServers.length < 0) {
+        if (null != bootstrapServers && bootstrapServers.length < 1) {
```
lb324567, line 164:
```suggestion
        if (null != bootstrapServers != null && bootstrapServers.length < 1) {
```
As you've already made a change i think this might as well get changed too

core/store/src/main/java/uk/gov/gchq/gaffer/store/schema/SchemaElementDefinition.java
```diff
@@ -196,7 +197,7 @@ public ElementAggregator getOriginalAggregator() {
     }
 
     @JsonIgnore
-    public ElementAggregator getFullAggregator() {
+    public synchronized ElementAggregator getFullAggregator() {
```
t92549, line 200: If the `IS2_INCONSISTENT_SYNC` is a false positive, does this need to have `synchronized` added?

t92549 (approved):

lb324567 (approved):

core/operation/src/main/java/uk/gov/gchq/gaffer/operation/impl/export/localfile/LocalFileExporter.java
```diff
@@ -56,8 +57,9 @@ public Iterable<String> get(final String filePath) throws OperationException {
             } else {
                 fileToBeRead = StreamUtil.openStream(LocalFileExporter.class, filePath);
             }
-            BufferedReader reader = new BufferedReader(new InputStreamReader(fileToBeRead));
+            BufferedReader reader = new BufferedReader(new InputStreamReader(fileToBeRead, StandardCharsets.UTF_8));
             linesFromFile = reader.lines().collect(Collectors.toList());
+            reader.close();
         } catch (final IOException e) {
```
r32575, line 63: Should this be in another `try` loop in case an error occurs before the `reader` can be closed?

core/store/src/main/java/uk/gov/gchq/gaffer/store/library/FileGraphLibrary.java
```diff
@@ -95,7 +95,7 @@ public Pair<String, String> getIds(final String graphId) {
 
     @Override
     protected void _addIds(final String graphId, final Pair<String, String> schemaAndPropsIds) throws OverwritingException {
-        String schemaAndPropsIdsString = new String(schemaAndPropsIds.getFirst() + "," + schemaAndPropsIds.getSecond());
+        String schemaAndPropsIdsString = schemaAndPropsIds.getFirst() + "," + schemaAndPropsIds.getSecond();
```
r32575, line 98: `String.join` may be neater

core/store/src/main/java/uk/gov/gchq/gaffer/store/library/FileGraphLibrary.java
```diff
@@ -122,12 +122,12 @@ protected void _addSchema(final String schemaId,
...
                                   final StoreProperties properties) {
         if (null != properties) {
-            getPropertiesPath(propertiesId).toFile().getParentFile().mkdirs();
+            Boolean dirCreated = getPropertiesPath(propertiesId).toFile().getParentFile().mkdirs();
             try (FileOutputStream propertiesFileOutputStream = new FileOutputStream(getPropertiesPath(propertiesId).toFile())) {
                 properties.getProperties().store(propertiesFileOutputStream, null);
             } catch (final IOException e) {
                 throw new IllegalArgumentException("Could not write " +
-                        "properties to path: " + getPropertiesPath(propertiesId), e);
+                        "properties to path: " + getPropertiesPath(propertiesId) + ". Directory created: " + dirCreated, e);
```
r32575, line 130: Could use `String.format` instead...

store-implementation/proxy-store/src/main/java/uk/gov/gchq/gaffer/proxystore/operation/handler/OperationChainHandler.java
```diff
@@ -36,6 +38,7 @@
 
 import static java.util.Objects.isNull;
 
+@SuppressFBWarnings(value = "NM_SAME_SIMPLE_NAME_AS_SUPERCLASS", justification = "Investigate")
```
GCHQDev404, line 41: Yes there is argument to rename this. I'll add it to an existing ticket.

GCHQDev404 (changes requested):

core/data/src/main/java/uk/gov/gchq/gaffer/data/elementdefinition/view/View.java
```diff
@@ -309,7 +309,7 @@ private boolean hasEdgeFilters(final Function<ViewElementDefinition, Boolean> ha
     }
 
     public boolean canMerge(final View addingView, final View srcView) {
-        if (addingView instanceof View && !(srcView instanceof View)) {
+        if (addingView != null && srcView == null) {
```
r32575, line 312: Confused by this... Could you theoretically have `addingView == null && srcView != null` as you are checking the `addingView` exists?

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/operation/handler/impl/FederatedNoOutputHandler.java
```diff
@@ -49,7 +49,8 @@
     public Void doOperation(final PAYLOAD operation, final Context context, final Store store) throws OperationException {
         FederatedOperation<Object, Void> federatedOperation = getFederatedOperation(operation);
 
-        Object ignore = store.execute(federatedOperation, context);
+        // Execution result is not stored or returned
```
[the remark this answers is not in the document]
GCHQDev404: I thought an appropriately named variable more worthwhile than a  than sentence in the comments.

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/operation/FederatedOperation.java
```diff
@@ -102,15 +103,12 @@ public FederatedOperation<INPUT, OUTPUT> mergeFunction(final BiFunction mergeFun
         return this;
     }
 
+    @JsonGetter("skipFailedFederatedExecution")
```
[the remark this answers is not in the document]
GCHQDeveloper314: This is exactly the same behaviour as before. What the field is set as initially (`false`) is the 'default'. We tell jackson to only serialise if not the default by using `@JsonInclude(value = JsonInclude.Include.NON_DEFAULT)`. This gives us the same results as if we'd returned `null`, as `null` is also not serialised by jackson. But this way we don't need a separate method, only an annotation.

The reason for making this change is only because the old method caused a SpotBugs violation by returning `null` but with with return type of `Boolean`.

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/operation/handler/impl/FederatedNoOutputHandler.java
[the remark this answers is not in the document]
GCHQDeveloper314: Only reason for changing this is that SpotBugs considers how it was done before to be a violation.

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/FederatedStore.java
```diff
@@ -145,7 +146,7 @@ public FederatedStore(@JsonProperty("customPropertiesAuths") final Set<String> c
                           @JsonProperty("storeConfiguredMergeFunctions") final Map<String, BiFunction> storeConfiguredMergeFunctions) {
         Integer i = null;
         while (isNull(i) || ALL_IDS.contains(i)) {
-            i = new Random().nextInt();
+            i = RANDOM.nextInt();
```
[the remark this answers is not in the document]
GCHQDeveloper314: Unfortunately SpotBugs considers the practice of non field Random to be a violation. If reverted this would need to be suppressed.

So the FederatedStore object is only ever created once for each federated store an instance of Gaffer might use? So normally this constructer is called once per JVM?

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/operation/handler/impl/FederatedNoOutputHandler.java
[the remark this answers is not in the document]
GCHQDeveloper314: SpotBugs does the opposite, it considers the unused variable to be bad practice. See the other comment.

core/data/src/main/java/uk/gov/gchq/gaffer/data/elementdefinition/view/View.java
GCHQDeveloper314: I was also confused by this, my commit for this change has some more notes on this. My understanding is that you could have a non-null `srcView` and a null `addingView`, these could be merged. This change does not impact the behaviour, unless I'm wrong about what `instanceof` is doing.

A merge is not possible (returning false) if there is no `srcView` but there is an `addingView`. If both are `null` then you can merge because two nulls make a null. This doesn't seem ideal to me, but as explained below, this is what the previous code did anyway.

The reason for this change is SpotBugs considers the use of `instanceof` when null checking could be done instead to be a violation. I think the original is certainly more confusing, as what is being checked is if the objects are null and not if they are an `instanceof` anything.

In this case, the only object that can be assigned to either of these is a `View` object, if it is assigned then the object is non-null, if it is not it is null. Therefore `instanceof` can be replaced with null checks which are easier to understand and probably more efficient as well.

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/FederatedStore.java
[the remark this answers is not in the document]
GCHQDev404: Yes normally You'd only have one FederatedStore per JVM. I think this should be surpressed and put back.

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/operation/FederatedOperation.java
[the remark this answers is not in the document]
GCHQDev404: 👍

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/util/ApplyViewToElementsFunction.java
```diff
@@ -50,6 +50,7 @@ public class ApplyViewToElementsFunction implements BiFunction<Object, Iterable<
     public static final String SCHEMA = "schema";
     public static final String USER = "user";
     public static final String TEMP_RESULTS_GRAPH = "temporaryResultsGraph";
+    private static final Random RANDOM = new Random();
```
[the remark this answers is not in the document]
GCHQDev404: It's used in the constructor only.

core/operation/src/main/java/uk/gov/gchq/gaffer/operation/impl/export/localfile/LocalFileExporter.java
GCHQDeveloper314: I've now changed this part to be in a nested try with resources. It will now always be closed and any exception caught by the original catch. I've also tidied the code up in this class to be more concise in general.

core/store/src/main/java/uk/gov/gchq/gaffer/store/library/FileGraphLibrary.java
GCHQDeveloper314: That could be done, though it's a longer line of code (even with static import) and from what I've read there's no performance benefit.

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/util/ApplyViewToElementsFunction.java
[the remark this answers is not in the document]
GCHQDeveloper314: If it it created many times then the `Random` object is recreated many times over which is less much efficient than creating only once and reusing. There is also a problem with increased probabilities of the same random number being used.

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/FederatedStore.java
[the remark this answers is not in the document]
GCHQDeveloper314: I've put this back to how it was and added the suppression.

core/access/src/main/java/uk/gov/gchq/gaffer/access/predicate/AccessPredicate.java
```diff
@@ -36,6 +38,7 @@
  * A {@link BiPredicate} which will first check if the user is an admin according to the provided
  * admin role. If not it uses a predicate to determine if the user can access a resource.
  */
+@SuppressFBWarnings(value = "SE_BAD_FIELD", justification = "Investigate if this class needs to implement Serializable")
```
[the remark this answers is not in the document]
GCHQDeveloper314: I've replaced the justification with that.

core/store/src/main/java/uk/gov/gchq/gaffer/store/library/FileGraphLibrary.java
GCHQDeveloper314: That does look more readable here, I've made the change.

core/store/src/main/java/uk/gov/gchq/gaffer/store/schema/SchemaElementDefinition.java
GCHQDeveloper314: Good question, I was hoping that adding that would fix it, but it doesn't seem to have done. I'm not sure why this class has any `synchronized ` at all?

core/operation/src/main/java/uk/gov/gchq/gaffer/operation/impl/export/localfile/LocalFileExporter.java
r32575: It is still seems to be the same code to me... No `try-with-resources`

r32575: Seems it has been changed after all. Just Github not showing the change. Looks good to me

r32575 (dismissed):

lb324567 (dismissed):

core/common-util/src/main/java/uk/gov/gchq/gaffer/commonutil/OneOrMore.java
t11947: I can't work out why this (casting to list if deduplicate is false) to use the remove(int i) method is being used in this case over the iterator method used for deduplicate. What am I missing here?

core/store/src/main/java/uk/gov/gchq/gaffer/store/operation/handler/named/cache/NamedOperationCache.java
```diff
@@ -229,6 +230,7 @@ public NamedOperationDetail getFromCache(final String name) throws CacheOperatio
         throw new CacheOperationFailedException(String.format("No named operation with the name %s exists in the cache", name));
     }
 
+    @SuppressFBWarnings(value = "DCN_NULLPOINTER_EXCEPTION", justification = "Investigate an improved null checking approach")
```
t11947, line 233: Could this just be an Objects.isNull check/throw like for name below it?

core/common-util/src/main/java/uk/gov/gchq/gaffer/commonutil/OneOrMore.java
GCHQDeveloper314: @t11947 When deduplicate is false it means the `Collection` is always a list. I think it could probably just use the iterator. However there is a difference in the behaviour at least. For the list it removes the last item, but when the iterator is used I think it's removing the first item. Given this method is called `removeAnyItem` how this works is pretty odd.

I'm not sure what would happen if we started removing from the start of lists instead of the end, but it could cause strange things to happen, so probably best to keep the current suboptimal situation!

core/store/src/main/java/uk/gov/gchq/gaffer/store/operation/handler/named/cache/NamedOperationCache.java
GCHQDeveloper314: I didn't look into this one very closely, but I think the `NullPointerException` originates in the `getOperationName()` method's class and not this class. Looking at that class (`NamedOperationDetail.java`), it seems there's a situation where the `String` field returned is unset - which is not the same as having a value of null, causing a null pointer. This explains why there's that separate `if` check with the `isNull` for way the value is set but is null.

This cannot be fixed in this class I don't think. This is the only place where `getOperationName()` is used, so this method could (and probably should) be changed to not have this problem.

GCHQDeveloper314: It requires more investigation, but if the `NamedOperationDetail` is itself null, that could be causing this, in which case adding an initial check for that would fix this problem.

findbugs-exclude.xml
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!--
-  ~ Copyright 2016-2020 Crown Copyright
+  ~ Copyright 2016-2022 Crown Copyright
```
t11947, line 3:
```suggestion
  ~ Copyright 2016-2023 Crown Copyright
```

t11947 (commented):
One or two extra bits from me

t11947 (approved):

GCHQDev404 (approved):

Issue #2845, opened 2022-12-20
Why is NamedOperation::parameters a LinkedHashMap?


t92549:
This should just be Map:
https://github.com/gchq/Gaffer/blob/f88215c6d98b48a805309b83c2f08c4f8188a47b/core/operation/src/main/java/uk/gov/gchq/gaffer/named/operation/NamedOperation.java#L66

It is understandable the setter must use LinkedHashMap in order to retain order, but the class variable doesn't have to use the class implementation.
https://github.com/gchq/Gaffer/blob/f88215c6d98b48a805309b83c2f08c4f8188a47b/core/operation/src/main/java/uk/gov/gchq/gaffer/named/operation/NamedOperation.java#L93

PR #2846, opened 2022-12-20
Gh-2845: Update class variable to use Map interface

kvaithin:
Resolves issue: #2845 

Updated class variable `parameter` to be of type `Map`.


t92549 (approved):

t92549:
Thanks for the PR!

PR #2852, opened 2023-01-04
Gh-2849: Fix JUnit @TempDir annotation

### LINKED ISSUE #2849: Incorrect use of JUnit @TempDir annotation
**Describe the bug**
[Since JUnit 5.9.0](https://github.com/junit-team/junit5/commit/0dad5f937acd418f9a3601531b6297e88be14a61) the [`@TempDir` annotation](https://junit.org/junit5/docs/current/user-guide/#writing-tests-built-in-extensions-TempDirectory) cannot be used with `final` fields.
We are using this is a few places, eg: https://github.com/gchq/Gaffer/blob/9421d9dfed70cfa14c475b8f8d9144589bc135ab/store-implementation/proxy-store/src/test/java/uk/gov/gchq/gaffer/proxystore/integration/ProxyStoreResponseDeserialiserIT.java#L56-L57
This code causes tests to fail (when they are being run as expected - see additional context).

**To Reproduce**
Run `mvn integration-test -Dit.test=ProxyStoreResponseDeserialiserIT -pl :proxy-store` to run the IT above, it will fail with an error stating the field must not be declared as final.

**Expected behavior**
Test to not fail.

**Stack trace and errors**
```
[ERROR] Tests run: 1, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 2.299 s <<< FAILURE! - in uk.gov.gchq.gaffer.proxystore.integration.ProxyStoreResponseDeserialiserIT
[ERROR] uk.gov.gchq.gaffer.proxystore.integration.ProxyStoreResponseDeserialiserIT.shouldUseOperationsResponseDeserialiserToDeserialiseOperationsResponse  Time elapsed: 0.007 s  <<< ERROR!
org.junit.jupiter.api.extension.ExtensionConfigurationException: @TempDir field [public final java.io.File uk.gov.gchq.gaffer.proxystore.integration.ProxyStoreResponseDeserialiserIT.testFolder] must not be declared as final.
```

**Additional context**
This problem has been present since #2732 was merged. However, for unknown reasons, version `3.0.0-M4` of the failsafe plugin does not run the IT above and therefore passes (it does recognise that class but none of the tests inside are run - `Tests run: 0`). Newer and older versions of this plugin **do** run the tests inside. Therefore this problem was not detected previously. In #2841 no changes are made to the tests/ITs in `proxy-store`, but this IT started being run again for unknown reasons, thereby revealing this issue related to the JUnit change. The problem of failsafe not running tests will be raised and tracked separately.

GCHQDeveloper314:
Also includes a fix to cleanup method in associated ITs which wasn't working correctly.

# Related Issue

- Resolve #2849

### DIFF
--- a/store-implementation/proxy-store/src/test/java/uk/gov/gchq/gaffer/proxystore/integration/ProxyStoreResponseDeserialiserIT.java
+++ b/store-implementation/proxy-store/src/test/java/uk/gov/gchq/gaffer/proxystore/integration/ProxyStoreResponseDeserialiserIT.java
@@ -30,6 +30,7 @@ import uk.gov.gchq.gaffer.operation.OperationChainDAO;
 import uk.gov.gchq.gaffer.operation.impl.add.AddElements;
 import uk.gov.gchq.gaffer.proxystore.ProxyProperties;
 import uk.gov.gchq.gaffer.proxystore.ProxyStore;
+import uk.gov.gchq.gaffer.proxystore.SingleUseMapProxyStore;
 import uk.gov.gchq.gaffer.proxystore.response.deserialiser.ResponseDeserialiser;
 import uk.gov.gchq.gaffer.rest.RestApiTestClient;
 import uk.gov.gchq.gaffer.rest.service.v2.RestApiV2TestClient;
@@ -58,6 +59,7 @@ public class ProxyStoreResponseDeserialiserIT {
 
     @BeforeAll
     public static void beforeAll() {
+        SingleUseMapProxyStore.cleanUp();
         CLIENT.startServer();
     }
 
--- a/store-implementation/proxy-store/src/test/java/uk/gov/gchq/gaffer/proxystore/integration/ProxyStoreResponseDeserialiserIT.java
+++ b/store-implementation/proxy-store/src/test/java/uk/gov/gchq/gaffer/proxystore/integration/ProxyStoreResponseDeserialiserIT.java
@@ -21,7 +21,6 @@ import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.io.TempDir;
 
-import uk.gov.gchq.gaffer.commonutil.CommonTestConstants;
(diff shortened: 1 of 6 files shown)


lb324567 (approved):

r32575 (approved):

store-implementation/proxy-store/src/test/java/uk/gov/gchq/gaffer/proxystore/integration/ProxyStoreResponseDeserialiserIT.java
```diff
@@ -53,11 +54,9 @@ public class ProxyStoreResponseDeserialiserIT {
 
     private static final RestApiTestClient CLIENT = new RestApiV2TestClient();
 
-    @TempDir
-    public final File testFolder = CommonTestConstants.TMP_DIRECTORY;
```
t92549, line 57: Is this incorrect everywhere it is used? Because I think there are other uses

GCHQDeveloper314: If it is marked as `final`, then JUnit considers that an error which justifies failing the test. Having it set to a value is incorrect as the value is set by JUnit and is actually set to the same thing as `CommonTestConstants.TMP_DIRECTORY` anyway. Having the value as public is not required either. There are no other instances in the codebase where it is marked as `final`, there are some where it is assigned a value, but that's not the problem here and would be more of a general tidy up task to fix.

This is an example of how it should be done if the temp directory is going to be used/reused in multiple tests in that class: https://github.com/gchq/Gaffer/blob/71199d3ebfa403404ce23d4eb5e0d4151dd38e89/core/operation/src/test/java/uk/gov/gchq/gaffer/export/LocalFileExporterTest.java#L49-L54

In this PR it was used in only one test and so could be moved into the test method as a parameter instead. This way it will be cleaned up once that individual test is run.

t92549 (approved):

PR #2861, opened 2023-01-09
Gh-2858: Fix ImportExportIT and ToCsvHandler

### LINKED ISSUE #2858: Fix ImportExportIT
Due to a bug with `readFile` the test is only testing the input file is the same as itself.
https://github.com/gchq/Gaffer/blob/adfa1163842872765ac290d124062029f5d6c5ca/integration-test/src/main/java/uk/gov/gchq/gaffer/integration/impl/ImportExportIT.java#L115

t92549:
# Related Issue

- Resolve #2858

### DIFF
--- a/core/store/src/main/java/uk/gov/gchq/gaffer/store/operation/handler/output/ToCsvHandler.java
+++ b/core/store/src/main/java/uk/gov/gchq/gaffer/store/operation/handler/output/ToCsvHandler.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2017-2022 Crown Copyright
+ * Copyright 2017-2023 Crown Copyright
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/core/store/src/main/java/uk/gov/gchq/gaffer/store/operation/handler/output/ToCsvHandler.java
+++ b/core/store/src/main/java/uk/gov/gchq/gaffer/store/operation/handler/output/ToCsvHandler.java
@@ -19,6 +19,7 @@ package uk.gov.gchq.gaffer.store.operation.handler.output;
 import uk.gov.gchq.gaffer.data.element.IdentifierType;
 import uk.gov.gchq.gaffer.data.generator.CsvFormat;
 import uk.gov.gchq.gaffer.data.generator.CsvGenerator;
+import uk.gov.gchq.gaffer.data.generator.OpenCypherFormat;
 import uk.gov.gchq.gaffer.operation.OperationException;
 import uk.gov.gchq.gaffer.operation.impl.output.ToCsv;
 import uk.gov.gchq.gaffer.store.Context;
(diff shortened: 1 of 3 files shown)


integration-test/src/main/java/uk/gov/gchq/gaffer/integration/impl/ImportExportIT.java
```diff
@@ -96,23 +95,22 @@ public void shouldImportFromFileThenExportBackOutToDifferentFile() throws Operat
...
-
-        //   When
-
         Iterable<String> incomingFile = readFile(INCOMING_FILE_PATH);
         Iterable<String> outgoingFile = readFile(path.toFile().getAbsolutePath());
 
 
-        //  Then
+        // Then
+        assertThat(expectedData).containsExactlyInAnyOrderElementsOf(exportedData);
```
GCHQDeveloper314, line 105: Might be better if it was asserting that the exported data contained the expected data and not the other way around, as normally we assert that what we have contains what we expect.

core/store/src/main/java/uk/gov/gchq/gaffer/store/operation/handler/output/ToCsvHandler.java
```diff
@@ -73,15 +74,18 @@ private LinkedHashMap<String, String> getPropertyHeadersFromSchema(final Store s
...
             }
-            for (final String propertyName:schemaEdgeDefinition.getProperties()) {
-                String typeName = schemaEdgeDefinition.getPropertyTypeName(propertyName);
-                propertyHeadersFromSchema.put(propertyName, schema.getType(typeName).getClazz().getSimpleName());
+            for (final String propertyName : schemaEdgeDefinition.getProperties()) {
+                final String typeName = schemaEdgeDefinition.getPropertyTypeName(propertyName);
+                final String javaType = schema.getType(typeName).getClazz().getSimpleName();
+                final String openCypherType = OpenCypherFormat.typeMappings.getOrDefault(javaType, javaType);
+
+                propertyHeadersFromSchema.put(propertyName, openCypherType);
```
GCHQDeveloper314, line 88: This change to use the `OpenCypherFormat.typeMappings` is causing tests to fail because they don't expect the OpenCypher `Int`.

The test failure is confusing because the asserts in there are also backwards, causing the error to reported that it expected `Int` but was `Integer`, when it is actually the other way around in the code.

I think it would be better if the mapping between Java and OpenCypher types was done in a separate for loop after these ones. Is there a reason why the identifier type doesn't need to be converted to a OpenCypher one?

core/store/src/main/java/uk/gov/gchq/gaffer/store/operation/handler/output/ToCsvHandler.java
```diff
@@ -73,15 +74,18 @@ private LinkedHashMap<String, String> getPropertyHeadersFromSchema(final Store s
             }
         }
         for (final SchemaEdgeDefinition schemaEdgeDefinition : schema.getEdges().values()) {
-            for (final IdentifierType identifierType:schemaEdgeDefinition.getIdentifiers()) {
+            for (final IdentifierType identifierType : schemaEdgeDefinition.getIdentifiers()) {
                 if (identifierType.toString().equals(identifierType.DIRECTED.toString())) {
```
GCHQDeveloper314, line 78: This should also be improved. There's no reason to convert this all to strings and then compare. This can be done instead:
```suggestion
                if (identifierType.equals(IdentifierType.DIRECTED)) {
```

GCHQDeveloper314 (changes requested):

core/store/src/main/java/uk/gov/gchq/gaffer/store/operation/handler/output/ToCsvHandler.java
t92549: The type mapping is only relevant for properties as only property headings have types. I have sorted the broken tests now

core/store/src/main/java/uk/gov/gchq/gaffer/store/operation/handler/output/ToCsvHandler.java
t92549: Have implemented your suggestion

integration-test/src/main/java/uk/gov/gchq/gaffer/integration/impl/ImportExportIT.java
t92549: Managed to fix and get this in

integration-test/src/main/java/uk/gov/gchq/gaffer/integration/impl/ImportExportIT.java
```diff
@@ -96,23 +95,22 @@ public void shouldImportFromFileThenExportBackOutToDifferentFile() throws Operat
...
-
-        //   When
-
         Iterable<String> incomingFile = readFile(INCOMING_FILE_PATH);
         Iterable<String> outgoingFile = readFile(path.toFile().getAbsolutePath());
 
 
-        //  Then
+        // Then
+        assertThat((Iterable<String>) exportedData).containsExactlyInAnyOrderElementsOf(expectedData);
```
GCHQDeveloper314, line 105: It might be cleaner to put the cast with the `graph.execute` and not therefore not using the `?` at all.

GCHQDeveloper314:
The failing test should be fixed by changes in #2842, once that has been merged this PR should be ready to merge.

GCHQDeveloper314 (approved):

lb324567 (approved):

t11947 (approved):

GCHQDev404 (approved):

PR #2867, opened 2023-01-20
gh-2866 Adding port options to RealFederatedStore example.

### LINKED ISSUE #2866: FederatedStore add port option to RealFederatedStore
Add port number option to the real federatedstore example.

This will allow having multiple rest running locally.

GCHQDev404:
# Related Issue

- Resolve #2866

### DIFF
--- a/store-implementation/proxy-store/src/main/java/uk/gov/gchq/gaffer/proxystore/ProxyProperties.java
+++ b/store-implementation/proxy-store/src/main/java/uk/gov/gchq/gaffer/proxystore/ProxyProperties.java
@@ -37,7 +37,7 @@ public class ProxyProperties extends StoreProperties {
     public static final String READ_TIMEOUT = "gaffer.read-timeout";
 
     public static final String DEFAULT_GAFFER_HOST = "localhost";
-    public static final String DEFAULT_GAFFER_CONTEXT_ROOT = "/rest/v2";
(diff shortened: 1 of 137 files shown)


store-implementation/proxy-store/src/main/java/uk/gov/gchq/gaffer/proxystore/ProxyProperties.java
```diff
@@ -37,7 +37,7 @@ public class ProxyProperties extends StoreProperties {
     public static final String READ_TIMEOUT = "gaffer.read-timeout";
 
     public static final String DEFAULT_GAFFER_HOST = "localhost";
-    public static final String DEFAULT_GAFFER_CONTEXT_ROOT = "/rest/v2";
+    public static final String DEFAULT_GAFFER_CONTEXT_ROOT = "/rest";
```
t11947, line 40: What was the reasoning behind this change?

GCHQDev404: Spring doesn't have V2 by default.

t92549: We should document that proxy by default now wants to point to a spring-rest and not the older one

t92549 (approved):

PR #2870, opened 2023-01-25
gh-2580 remove dynamic schema.

### LINKED ISSUE #2580: Fix FederatedStore checks for DYNAMIC_SCHEMA
Due to a partial implementation of #2008, currently FederatedOperationChains only validate the View if the graph does not have the `DYNAMIC_SCHEMA` trait:
https://github.com/gchq/Gaffer/blob/40bb337d2ca91fea2dadcd04cddec46021c169ee/store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/operation/FederatedOperationChainValidator.java#L82
https://github.com/gchq/Gaffer/blob/40bb337d2ca91fea2dadcd04cddec46021c169ee/store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/util/FederatedStoreUtil.java#L134

This trait only exists in the FederatedStore as its getTraits() implementation returns ALL_TRAITS. This effectively means that only FederatedStores and ProxyStores which point to a FederatedStore will not be validated here. However, this cannot be used going forward because:
- FederatedStore fixes introduced in the FederatedStore alpha mean it will no longer return ALL_TRAITS
- the getTraits function is being replaced by the Operation which behaves differently, instead of returning ALL_TRAITS it will use the FederatedGetTraitsHandler which correctly returns the traits that the FederatedStore's stores have

Therefore, before the getTraits/getStoreTraits function can be completely removed, this usage has to be replaced (under this ticket). This can be done in various ways:
- deciding on a different solution to the partial implementation of #2008: either by fully implementing it or reverting it perhaps?
- replacing the FederatedStore check with something else smarter than this DYNAMIC_SCHEMA trait check

A solution to this must be discussed, decided on and implemented before getTraits can be completely removed and it seems that it will be best to wait until the other FederatedStore changes first to do so.

GCHQDev404:
# Related Issue

- Resolve #2580

### DIFF
--- a/store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/operation/FederatedOperationChainValidator.java
+++ b/store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/operation/FederatedOperationChainValidator.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016-2022 Crown Copyright
+ * Copyright 2016-2023 Crown Copyright
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -22,7 +22,6 @@ import uk.gov.gchq.gaffer.graph.GraphSerialisable;
 import uk.gov.gchq.gaffer.operation.Operation;
 import uk.gov.gchq.gaffer.store.Context;
 import uk.gov.gchq.gaffer.store.Store;
-import uk.gov.gchq.gaffer.store.StoreTrait;
 import uk.gov.gchq.gaffer.store.operation.OperationChainValidator;
 import uk.gov.gchq.gaffer.store.schema.Schema;
 import uk.gov.gchq.gaffer.store.schema.ViewValidator;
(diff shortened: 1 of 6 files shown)


store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/util/FederatedStoreUtil.java
```diff
@@ -146,10 +145,6 @@ public static <OP extends Operation> OP updateOperationForGraph(final OP operati
                     if (validView.hasGroups()) {
                         ((OperationView) resultOp).setView(validView);
                         // Deprecated function still in use due to Federated GetTraits bug with DYNAMIC_SCHEMA
```
t11947, line 147:
```suggestion
                        ((OperationView) resultOp).setView(validView);
```

t11947 (changes requested):
Just a single suggestion removing something referencing DYNAMIC SCHEMA in a comment.

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/operation/FederatedOperationChainValidator.java
```diff
@@ -101,10 +100,7 @@ protected void validateViews(final Operation op, final User user, final Store st
                 if (graphIdValid) {
                     currentResult = new ValidationResult();
                     clonedOp.graphIdsCSV(graphId);
-                    // Deprecated function still in use due to Federated GetTraits bug with DYNAMIC_SCHEMA
-                    if (!graphSerialisable.getGraph().getStoreTraits().contains(StoreTrait.DYNAMIC_SCHEMA)) {
```
t92549, line 105: This only used to run on proxy stores and federated stores but now on all, is this okay?

GCHQDev404: "**Not** (Traits contains Dynamic Schema)"

GCHQDev404:
superseded

PR #2884, opened 2023-02-08
gh-2881-Problems with getOriginalSchema in Federated and Proxy Stores

### LINKED ISSUE #2881: Problems with getOriginalSchema in Federated and Proxy Stores
Problems have been reported fetching schemas with Federated and Proxy Stores. The current implementation of these stores contain overridden methods for `getSchema` but not `getOriginalSchema`. It's likely that with certain combinations of graphs and stores, the `originalSchema` field could remain `null` or be outdated. This could cause problems with the getSchema operation when used directly on the store (not wrapped) and differences between using a `Store` directly and using it through a `Graph`.

Investigate and fix any problems with Federated and Proxy Store use of `getSchema`/`getOriginalSchema`. This might be as simple as adding overridden methods for `getOriginalSchema`.

**Additional context**
See #2791 for background.

GCHQDev404:
# Related Issue

- Resolve #2881


store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/util/FederatedStoreUtil.java
```diff
@@ -282,7 +282,12 @@ public static String getDeprecatedGraphIds(final Operation operation) throws Gaf
 
     @Deprecated
     public static FederatedOperation<Void, Iterable<Schema>> getFederatedWrappedSchema() {
-        return new FederatedOperation.Builder().<Void, Iterable<Schema>>op(new GetSchema()).build();
+        return getFederatedWrappedSchema(true);
+    }
+    public static FederatedOperation<Void, Iterable<Schema>> getFederatedWrappedSchema(final boolean getOptimisedSchema) {
+        final GetSchema.Builder getSchema = new GetSchema.Builder();
+        getSchema.option("UnsupportedGetOptimisedSchema", String.valueOf(getOptimisedSchema));
+        return new FederatedOperation.Builder().<Void, Iterable<Schema>>op(getSchema.build()).build();
```
GCHQDeveloper314, line 290: The option given doesn't exist. If you want the optimised version you use `GetSchema.Builder().compact(true)`. Maybe this can be:
```suggestion
    public static FederatedOperation<Void, Iterable<Schema>> getFederatedWrappedSchema(final boolean getCompactSchema) {
        final GetSchema.Builder getSchema = new GetSchema.Builder();
        getSchema.compact(getCompactSchema);
        return new FederatedOperation.Builder().<Void, Iterable<Schema>>op(getSchema.build()).build();
```

GCHQDeveloper314 (changes requested):

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/util/FederatedStoreUtil.java
t92549: I would say this should be false by default

GCHQDeveloper314: It should yes. It might also be worth looking at what happens to non-wrapped `GetSchema` operations with the compact option set when they are sent to a federated store. I'm not sure that this option is preserved at the moment. More tests would be needed for this.

GCHQDeveloper314:
This PR also includes changes to remove use of and delete the following deprecated methods from the Federated Store:

- `getSchema` from `FederatedGraphStorage.java`.
- `getFederatedWrappedSchema` from `FederatedStoreUtil.java`.

6 tests from `FederatedGraphStorageTest` which had been marked as deprecated (due to using the above `getSchema` method) have been migrated to StoreSchemaTest.

GCHQDeveloper314 (approved):

GCHQDev404:
Approved.

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/FederatedStore.java
```diff
@@ -324,34 +324,50 @@ public List<String> getAllGraphIds(final User user, final boolean userRequesting
...
+     * {@link Context}, or ideally use the {@link GetSchema}
+     * operation instead.
      *
-     * @return schema
+     * @return {@link Schema} blank schema
      */
     @Override
     public Schema getSchema() {
-        return getSchema((Context) null);
+        return getSchema(new Context(), true);
```
t92549, line 337: @GCHQDeveloper314 Has the default behaviour changed so that compact schema is now returned? Or was this always the behaviour?

store-implementation/federated-store/src/test/java/uk/gov/gchq/gaffer/federatedstore/FederatedStoreTest.java
```diff
@@ -335,15 +335,15 @@ public void shouldUpdateSchemaWhenNewGraphIsAdded() throws Exception {
     public void shouldUpdateSchemaWhenNewGraphIsRemoved() throws Exception {
         // Given
         addGraphWithPaths(ACC_ID_1, propertiesAlt, blankUserContext, SCHEMA_ENTITY_BASIC_JSON);
-        final Schema was = store.getSchema(new Context(blankUser));
+        final Schema was = store.getSchema(new Context(blankUser), true);
```
t92549, line 338: If there was a `getSchema` added which took no args and defaulted to true (or false; whichever was default behaviour), then there would be a lot less unnecessary changes

GCHQDeveloper314: I was considering adding one to do exactly that, but I believe this was done deliberately by @GCHQDev404 in order to ensure clarity?

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/FederatedStore.java
GCHQDeveloper314: The original behaviour for this method was actually to almost always return an empty schema. Although that was not clear from this class (new Javadoc clarifies this). It still does this (though no longer via a deprecated class), so the **default behaviour is unchanged from before**.

GCHQDeveloper314 (approved):

t92549 (approved):

PR #2888, opened 2023-02-13
Gh 2066: Add warning for edge directed validation

### LINKED ISSUE #2066: Add Edge Validation warning for directed field
Validation should be added to directed fields in schemas. This will be done as part of work for version 2 as adding in validation would cause breaking changes.

t92549:
# Related Issue

- Resolve #2066


core/store/src/test/java/uk/gov/gchq/gaffer/store/schema/SchemaElementDefinitionValidatorTest.java
```diff
@@ -624,4 +624,17 @@ public void shouldValidateAndReturnFalseWhenAggregatorSelectionHasUnknownPropert
                 ),
                 result.getErrors());
     }
+
+    @Test
+    public void shouldValidateAndReturnTrueWhenEdgeDirectionUnset() {
```
GCHQDeveloper314, line 629: It's unfortunate we cannot easily ensure there is actually a warning logged.

core/store/src/main/java/uk/gov/gchq/gaffer/store/schema/SchemaElementDefinitionValidator.java
```diff
@@ -168,10 +166,10 @@ protected ValidationResult validateFunctionArgumentTypes(
         return result;
     }
 
-    private ValidationResult validateDirection(final SchemaElementDefinition elementDef) {
+    protected ValidationResult validateDirection(final SchemaElementDefinition elementDef) {
```
GCHQDeveloper314, line 169: Why is this now protected? There are no children of the `SchemaElementDefinitionValidator` class. I also see other methods in this class which could be private.

core/store/src/main/java/uk/gov/gchq/gaffer/store/schema/SchemaElementDefinitionValidator.java
```diff
@@ -168,10 +166,10 @@ protected ValidationResult validateFunctionArgumentTypes(
...
         final ValidationResult result = new ValidationResult();
         if (elementDef instanceof SchemaEdgeDefinition && elementDef.getIdentifierTypeName(IdentifierType.DIRECTED) == null) {
-            result.addError("\"directed\" must be set");
+            LOGGER.warn("Edge definition does not have directed set.");
```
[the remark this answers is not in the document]
t92549: Within a definition you cannot know the name of the group that is using the definition (annoyingly). This is because there is a Map of group names to definitions.

core/store/src/test/java/uk/gov/gchq/gaffer/store/schema/SchemaElementDefinitionValidatorTest.java
t92549: I have tested manually but yes, no easy way to test logging

core/store/src/main/java/uk/gov/gchq/gaffer/store/schema/SchemaElementDefinitionValidator.java
t92549: Because it needs to be called in the Test, a few of the other functions are the same for this reason I think. It just makes the test cleaner with less mocking if I can call the method directly

core/store/src/test/java/uk/gov/gchq/gaffer/store/schema/SchemaElementDefinitionValidatorTest.java
GCHQDeveloper314: There are some ways to do it with Reload4j appenders and/or mocking. We wouldn't want to bring that in here though. It might be something for a future issue - code for testing that logging works could sit in a test util class.

core/store/src/main/java/uk/gov/gchq/gaffer/store/schema/SchemaElementDefinitionValidator.java
GCHQDeveloper314: Not exactly the best practice though, but I had suspected that was the reason.

GCHQDeveloper314 (approved):

PR #2889, opened 2023-02-13
Add tests to ensure element definition orders

### LINKED ISSUE #1804: Signature of methods in SchemaElementDefinition
In `SchemaElementDefinition` the signature of the method `getGroupBy` is to return `Set<String>`. The actual object that is returned is of type `LinkedHashSet`. Many callers of this method assume that the results are in a consistent order, e.g. when properties are serialised they need to always be serialised in the same order. As the result is a `LinkedHashSet`, the results are in a consistent order. But as the signature doesn't guarantee that, in theory the implementation of `SchemaElementDefinition` could change (to return a `HashSet` for example) and client code would break as the order would not be consistent.

There are similar issues with `getIdentifierMap` and `getProperties`.

EDIT 
This branch has attempted a workaround. @d21211122 says:
For Gaffer 2.0,  undo the workaround, deprecate these methods and replace them with ones that guarantee a consistent order and update all the callers of the methods in Gaffer to use the new ones.

t92549:
# Related Issue

- Resolve #1804


GCHQDeveloper314 (approved):
Could a comment be added to the issue explaining this PR?

PR #2893, opened 2023-02-19
Gh 2890 cache service static instance bug

### LINKED ISSUE #2890: Cache Service Static Instance Bug
**Describe the bug**
Cache service is Static, which means that any attempt to use a new cache can wipe/swap the cache.
This is horrendous for FederatedStore which uses the cache to store graphs. This is an Ancient bug but since the internal GraphStorage java object has been removed since #2595 this bug has appeared during docker containers testing.

Truth Table of error and behaviour below.

**To Reproduce**
  | AddGraph with no Cache | AddGraph with HashMap Cache | AddGraph with JCS Cache
-- | -- | -- | --
FederatedStore is currently using HashMap Cache | Graph is added to HashMap Cache | The Cache service continues to be HashMap Cache, the cache is initialised, FederatredStore only contains this new graph. | The Cache service is switched to JCS Cache, the cache is initialised, JCS cache is loaded from disc, FederatredStore contains this new graph and anything pre-existing in the cache.
FederatedStore is currently using JCS Cache | Graph is added to JCS Cache | The Cache service is switched to HashMap Cache, the cache is initialised, FederatredStore only contains this new graph. | The Cache service continues to be JCS Cache, the cache is initialised, JCS cache is loaded from disc, FederatredStore contains this new graph and anything pre-existing in the cache.

GCHQDev404:
# Related Issue

- Resolve #2890


core/store/src/test/java/uk/gov/gchq/gaffer/store/operation/handler/named/cache/NamedOperationCacheBackwardCompatibilityTest.java
```diff
@@ -1,68 +0,0 @@
...
-import uk.gov.gchq.gaffer.operation.OperationChain;
-import uk.gov.gchq.gaffer.operation.impl.add.AddElements;
-import uk.gov.gchq.gaffer.user.User;
-
-import java.util.Properties;
-
-import static java.util.Arrays.asList;
-import static org.junit.jupiter.api.Assertions.assertEquals;
-
-public class NamedOperationCacheBackwardCompatibilityTest {
```
t92549, line 35: How come 2 of the backward compatibility tests have been deleted?

core/cache/src/main/java/uk/gov/gchq/gaffer/cache/CacheServiceLoader.java
```diff
@@ -56,6 +56,12 @@ public static void initialise(final Properties properties) {
             }
             return;
         }
+
+        if (isEnabled()) {
```
t92549, line 60: I think there needs to be a test added which fails before this change and passes to prove it works

GCHQDev404: This was added because test were failing.

core/store/src/test/java/uk/gov/gchq/gaffer/store/operation/handler/named/cache/NamedOperationCacheBackwardCompatibilityTest.java
t92549: These will be re-added before 2.0 release: https://github.com/gchq/Gaffer/issues/2895

t92549 (approved):

PR #2901, opened 2023-02-22
Gh-2581: Replace getTraits usages with operation

### LINKED ISSUE #2581: Replace getTrait usages with Operation
https://github.com/gchq/Gaffer/issues/2552 removed deprecated code, however, the `getTraits` function is still used (although deprecated).
Similar to https://github.com/gchq/Gaffer/issues/2573, the function version of `getTraits` should be replaced by executing the `GetTraits` Operation.

This, however, cannot be done as part of alpha 1 and has to be done later. This is because it is reliant upon https://github.com/gchq/Gaffer/issues/2580 which needs to be done with the major FederatedStore changes.

GCHQDeveloper314:
A test has been disabled in Federated Store because it cannot work with these changes until the work in 2580 to remove `DYNAMIC_SCHEMA` has been done. This test can be re-enabled as part of that PR. This test has been marked with a TODO comment.

# Related Issue

- Resolve #2581

### DIFF
--- a/core/store/src/main/java/uk/gov/gchq/gaffer/store/SchemaOperationChainValidator.java
+++ b/core/store/src/main/java/uk/gov/gchq/gaffer/store/SchemaOperationChainValidator.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2018-2020 Crown Copyright
+ * Copyright 2018-2023 Crown Copyright
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/core/store/src/main/java/uk/gov/gchq/gaffer/store/SchemaOperationChainValidator.java
+++ b/core/store/src/main/java/uk/gov/gchq/gaffer/store/SchemaOperationChainValidator.java
@@ -47,7 +47,7 @@ public class SchemaOperationChainValidator extends OperationChainValidator {
     }
 
     @Override
-    protected Set<StoreTrait> getStoreTraits(final Store store) {
+    protected Set<StoreTrait> getStoreTraits(final Store store, final Context context) {
         return StoreTrait.ALL_TRAITS;
     }
 }
--- a/core/store/src/test/java/uk/gov/gchq/gaffer/store/operation/OperationChainValidatorTest.java
+++ b/core/store/src/test/java/uk/gov/gchq/gaffer/store/operation/OperationChainValidatorTest.java
@@ -17,6 +17,7 @@
 package uk.gov.gchq.gaffer.store.operation;
 
 import com.google.common.collect.Lists;
+import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 import org.mockito.Mockito;
 
@@ -25,6 +26,7 @@ import uk.gov.gchq.gaffer.commonutil.iterable.EmptyIterable;
 import uk.gov.gchq.gaffer.data.element.comparison.ElementPropertyComparator;
 import uk.gov.gchq.gaffer.operation.Operation;
 import uk.gov.gchq.gaffer.operation.OperationChain;
+import uk.gov.gchq.gaffer.operation.OperationException;
 import uk.gov.gchq.gaffer.operation.impl.DiscardOutput;
 import uk.gov.gchq.gaffer.operation.impl.compare.Max;
 import uk.gov.gchq.gaffer.operation.impl.export.GetExports;
@@ -33,6 +35,7 @@ import uk.gov.gchq.gaffer.operation.impl.generate.GenerateObjects;
 import uk.gov.gchq.gaffer.operation.impl.get.GetAdjacentIds;
 import uk.gov.gchq.gaffer.operation.impl.get.GetElements;
 import uk.gov.gchq.gaffer.operation.impl.output.ToVertices;
+import uk.gov.gchq.gaffer.store.Context;
 import uk.gov.gchq.gaffer.store.Store;
 import uk.gov.gchq.gaffer.store.schema.Schema;
 import uk.gov.gchq.gaffer.store.schema.ViewValidator;
@@ -40,6 +43,7 @@ import uk.gov.gchq.gaffer.user.User;
 import uk.gov.gchq.koryphe.ValidationResult;
 
 import java.util.Arrays;
+import java.util.HashSet;
 import java.util.Set;
 
 import static org.assertj.core.api.Assertions.assertThat;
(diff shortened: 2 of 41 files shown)


store-implementation/map-store/src/test/java/uk/gov/gchq/gaffer/mapstore/MapStoreTest.java
```diff
@@ -44,7 +45,7 @@ public void testTraits() throws StoreException {
                 StoreTrait.TRANSFORMATION,
                 StoreTrait.POST_TRANSFORMATION_FILTERING,
                 StoreTrait.MATCHED_VERTEX));
-        assertEquals(expectedTraits, mapStore.getTraits());
+        assertEquals(expectedTraits, mapStore.TRAITS);
```
t92549, line 48: I think maybe this can stay as `getTraits`? Or should at least be `MapStore.TRAITS`

store-implementation/proxy-store/src/main/java/uk/gov/gchq/gaffer/proxystore/ProxyStore.java
```diff
@@ -357,14 +345,14 @@ protected OperationHandler<? extends AddElements> getAddElementsHandler() {
...
-            throw new GafferRuntimeException(e.getMessage(), e);
+        } catch (final OperationException e) {
+            throw new GafferRuntimeException("Unable to fetch traits from remote store", e);
         }
     }
 
     @Override
     protected OperationHandler<? extends OperationChain<?>> getOperationChainHandler() {
-        return new OperationChainHandler<>(opChainValidator, opChainOptimisers);
+        return null;
```
t92549, line 355: 👍

store-implementation/proxy-store/src/test/java/uk/gov/gchq/gaffer/proxystore/integration/ProxyStoreBasicIT.java
```diff
@@ -216,13 +218,13 @@ public void shouldCatchAndThrowUsefulErrorMessages() throws Exception {
...
     @Test
-    public void shouldHaveAllOfDelegateStoreTraitsApartFromVisibility() {
+    public void shouldHaveAllOfDelegateStoreTraitsApartFromVisibility() throws OperationException {
         // Given
         final Set<StoreTrait> expectedTraits = new HashSet<>(MapStore.TRAITS);
         expectedTraits.remove(StoreTrait.VISIBILITY);
 
         // When
-        final Set<StoreTrait> storeTraits = graph.getStoreTraits();
+        final Set<StoreTrait> storeTraits = graph.execute(new GetTraits.Builder().currentTraits(false).build(), new Context());
```
t92549, line 227: If the Operation is meant to replace the function, perhaps `currentTraits` should default to false?

core/store/src/main/java/uk/gov/gchq/gaffer/store/SchemaOperationChainValidator.java
```diff
@@ -47,7 +47,7 @@ protected Schema getSchema(final Operation operation, final User user, final Sto
     }
 
     @Override
-    protected Set<StoreTrait> getStoreTraits(final Store store) {
+    protected Set<StoreTrait> getStoreTraits(final Store store, final Context context) {
```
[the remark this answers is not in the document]
GCHQDeveloper314: From what I can tell it could be because when validating a schema all traits should be considered. The way this is implemented is not all that clear. I think it would be better to remove this class and refactor, this probably wouldn't be much more work than documenting this. Something for a new issue.

rest-api/core-rest/src/test/resources/mockito-extensions/org.mockito.plugins.MockMaker
```diff
@@ -0,0 +1 @@
+mock-maker-inline
```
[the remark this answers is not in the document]
GCHQDeveloper314: It enables Mockito's inline mock extension. This is required for using Mockito on final classes. I did this to use add a stub to `Graph` for return the results of an executed operation. There is actually another way to do this by mocking the `Store`. 

It's also possible to enable using an extra dependency, but I think this makes it clearer that it's being used. I could also probably remove this and use a different mocking approach.

store-implementation/map-store/src/test/java/uk/gov/gchq/gaffer/mapstore/MapStoreTest.java
GCHQDeveloper314: Yes this needn't be changed. It's left over from a previous commit that I otherwise undid.

store-implementation/proxy-store/src/test/java/uk/gov/gchq/gaffer/proxystore/integration/ProxyStoreBasicIT.java
GCHQDeveloper314: It could and probably should have done from the start - if it was actually intended to do that. I don't think we should change it in this PR. Might be something to consider if we remove the REST endpoints - I've proposed this in #2913.

rest-api/core-rest/src/test/resources/mockito-extensions/org.mockito.plugins.MockMaker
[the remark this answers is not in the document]
t92549: I don't have an issue with it really, just never seen it

t92549 (approved):

core/store/src/test/java/uk/gov/gchq/gaffer/store/operation/OperationChainValidatorTest.java
```diff
@@ -49,6 +53,18 @@
...
+    private final ViewValidator viewValidator = mock(ViewValidator.class);
+    private final OperationChainValidator validator = new OperationChainValidator(viewValidator);
+    private final Store store = mock(Store.class);
+    private final User user = mock(User.class);
+    private final Schema schema = mock(Schema.class);
+
+    @BeforeEach
+    public void setup() throws OperationException {
+        given(store.getSchema()).willReturn(schema);
+        given(store.execute(any(GetTraits.class), any(Context.class))).willReturn(new HashSet<>());
```
GCHQDev404, line 65: So no traits are returned from mock?

GCHQDev404 (approved):

core/store/src/test/java/uk/gov/gchq/gaffer/store/operation/OperationChainValidatorTest.java
GCHQDeveloper314: Yes it returns an empty set of traits, i.e. no traits. This was what happened before, so far as I can tell. We just have to mock it now as otherwise we get `null` from the operation.

PR #2908, opened 2023-02-24
gh-2903 Return Proxy to being empty

### LINKED ISSUE #2903: Fix FederatedStore cache collision
**Describe the bug**
When using the same cache implementation in a FederatedStore, the caches will collide.

**To Reproduce**
1. Create a FederatedStore
2. Add GraphA with a cache
3. Add GraphB with a cache (same cache implementation)
4. Add a NamedOperation to GraphA (NamedOperations are stored in the cache)
5. List NamedOperations in GraphB - it will be visibile even though it should not be

**Suggested solution**
All graphs require suffix usage, similar to FederatedGraphStroage Cache.

GCHQDev404:
However the return object is not an element but a list of element contents.

# Related Issue

- Resolve #2903


GCHQDev404:
superseded

PR #2921, opened 2023-03-02
Gh 2903 return proxy store to no cache handlers

GCHQDev404:


t92549:
Superseded by https://github.com/gchq/Gaffer/pull/2923

PR #2925, opened 2023-03-06
gh-2865 Cache Exceptions usage and review

### LINKED ISSUE #2865: Cache Exceptions usage and review
**Describe the bug**
Following #2319 The caches now share a parent class `Cache`. However this combined the need for certain Exceptions to be thrown and required by some test. These Exceptions need to be reviewed. One example is a test to for when a Key is not found in the Cache an Exception is thrown. Keys not being found in a cache is not "exceptional behaviour".

GCHQDev404:
# Related Issue

- Resolve #2865

### DIFF
--- a/core/store/src/main/java/uk/gov/gchq/gaffer/store/operation/handler/named/cache/NamedOperationCache.java
+++ b/core/store/src/main/java/uk/gov/gchq/gaffer/store/operation/handler/named/cache/NamedOperationCache.java
@@ -30,8 +30,6 @@ import java.util.HashSet;
 import java.util.Objects;
 import java.util.Set;
 
-import static java.util.Objects.nonNull;
-
 /**
  * Wrapper around the {@link uk.gov.gchq.gaffer.cache.CacheServiceLoader} to provide an interface for handling
  * the {@link uk.gov.gchq.gaffer.named.operation.NamedOperation}s for a Gaffer graph.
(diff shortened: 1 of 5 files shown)


core/store/src/main/java/uk/gov/gchq/gaffer/store/operation/handler/named/cache/NamedOperationCache.java
```diff
@@ -145,36 +138,23 @@ public NamedOperationDetail getFromCache(final String name) throws CacheOperatio
...
+    public void addNamedOperation(final NamedOperationDetail namedOperation, final boolean overwrite, final User user, final String adminAuth) throws CacheOperationException {
+        if (namedOperation == null) {
+            throw new CacheOperationException("NamedOperation cannot be null");
         }
-        if (!overwrite) {
-            addToCache(name, namedOperation, false);
-            return;
+        String name = namedOperation.getOperationName();
+        if (contains(name) && overwrite) {
+            final boolean doesUserHavePermissionToWrite = getFromCache(name).hasWriteAccess(user, adminAuth);
```
t92549, line 147: Shouldn't user need write access to perform normal add also (not overwriting)?

GCHQDev404: So in this situation.

I add a NamedOperation to cache, but I specify you have write permissions. This means you can override my NamedOperation. 

Else where in Gaffer land there will be checks for if you can use the AddNamedOperation, But here is where a 3rd person is stopped from overriding my namedOperation, but who has AddNamedOperation permission.

t92549: Is it possible to add a NamedOperation that I will not be able to access later?

t92549: This shouldn't happen and if it does it is an unrelated preexisting issue

t92549 (approved):

Issue #2914, opened 2023-03-01
Remove redundant constants class


GCHQDeveloper314:
The class `CommonConstants.java` has two static fields. One is unused and the other (`UTF-8`) is used in various places in Gaffer but can be replaced with an equivalent Java builtin.

Remove this class and replace references to its field with the equivalent Java builtin.

ibrijesh:
I am interested in contributing, Could you please assign me this task?

I would like to clarify the following changes that need to be made:

1.Remove the file named **CommonConstants.java**.
2.Replace the line import **uk.gov.gchq.gaffer.commonutil.CommonConstants;** with  **import java.nio.charset.StandardCharsets;**.
3.Replace all instances of **CommonConstants.UTF_8** with **StandardCharsets.UTF_8**.

GCHQDeveloper314:
Yes that is correct. You may also need to ensure the import order is correct after replacing the old import, as we have checkstyle rules for enforcing a clean order. You can run the checkstyle checks without needing to do a full build and test of the codebase by using `mvn checkstyle:check`.

ibrijesh:
When i am runniing this command **mvn checkstyle:check** I am getting below error

![Screenshot from 2023-03-07 14-36-30](https://user-images.githubusercontent.com/41025295/223375408-e048f580-b328-45cf-abb6-f14820f555a8.png)

GCHQDeveloper314:
It looks like you will need to run `mvn clean install -Pquick` first, in order to build the project dependencies.

GCHQDeveloper314:
Java 8 is required for Gaffer 1.22 versions. If you need to use Java 11 then checkout the `v2-alpha` branch and use that instead of the default `develop` branch. The `v2-alpha` branch is where we are doing development for Gaffer 2.0 and it supports Java 11.

ibrijesh:
I have succefully build project dependencies **mvn clean install -Pquick**  but changes in code failing **mvn checkstyle:check** . What rules i have to follow  while making changes ?

GCHQDeveloper314:
When running the checks a list of checkstyle rule failures will be shown for each module which is failing. These are not summarised at the end, so you'll need to scroll up in order to see them. It may be easier to run the checks on the individual modules you've modified by using the `-pl :module-name` flag (e.g. `-pl :store`). You could also open a draft PR which will show us any checkstyle issues as part of our GitHub actions checks.

PR #2927, opened 2023-03-12
#2914 Issue: Remove redundant constants class

ibrijesh:
ISSUE #2914 

Removed redundant constants class


library/hdfs-library/src/main/java/uk/gov/gchq/gaffer/hdfs/operation/handler/job/tool/SampleDataAndCreateSplitsFileTool.java
```diff
@@ -155,7 +155,7 @@ private void writeSplits(final FileSystem fs, final Path resultsFile, final long
         try (final SequenceFile.Reader reader = new SequenceFile.Reader(fs.getConf(), Reader.file(resultsFile));
              final PrintStream splitsWriter = new PrintStream(
                      new BufferedOutputStream(fs.create(new Path(operation.getSplitsFilePath()), true)),
-                     false, CommonConstants.UTF_8)
+                     false, String.valueOf(StandardCharsets.UTF_8))
```
GCHQDeveloper314, line 158: Once we move to using Java 11 [we can use the Charset directly](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/io/PrintStream.html). This is just a note to add this info to issue 2874.

core/operation/src/main/java/uk/gov/gchq/gaffer/named/operation/AddNamedOperation.java
```diff
@@ -71,7 +71,7 @@ public class AddNamedOperation implements Operation, Operations<Operation> {
     private AccessPredicate readAccessPredicate;
     private AccessPredicate writeAccessPredicate;
 
-    private static final String CHARSET_NAME = CommonConstants.UTF_8;
+    private static final String CHARSET_NAME = String.valueOf(StandardCharsets.UTF_8);
```
GCHQDeveloper314, line 74: This could probably be removed and `StandardCharsets.UTF_8` use directly in place of it. I don't think having this a static String adds anything.

GCHQDeveloper314 (commented):
The use of the static `CHARSET_NAME` in some classes could probably be removed. It adds nothing compared to using the charset directly and introduces additional exceptions which have to be handled.

GCHQDeveloper314 (approved):

GCHQDeveloper314:
The codecov failures can be ignored, those checks are broken.

PR #2955, opened 2023-05-17
Gh-2954: Remove mentions of v2-alpha

### LINKED ISSUE #2954: Remove final mentions of v2-alpha and update README
After v2 is merged into develop, we should remove it from the README and pom.
As well as this, it would be good to update the README to match [koryphe's](https://github.com/gchq/koryphe#readme).

t92549:
# Related Issue

- Resolve #2954


t92549:
This needs to happen after merging v2-alpha into develop

GCHQDeveloper314 (approved):

Issue #2972, opened 2023-06-08
Codecov can generate inconsistent coverage reports


GCHQDeveloper314:
**Describe the bug**
There are currently two situations where the coverage reports codecov generates are incorrect:

- When the CI runs coverage reports are generated in [7 different matrix jobs](https://github.com/gchq/Gaffer/blob/f4f40d3ba1d83b8b12a571a72acbf100c8753e7c/.github/workflows/continuous-integration.yaml#L49). These do not complete at the same time which results in an initial coverage report that's incorrect and is then updated as each report is uploaded, eventually becoming correct **only if all jobs pass**. In the interim the report is wrong and this is misleading.
- When there is a problem with the codecov upload for one of the 7 matrix steps this can cause coverage to become wrong if the problem doesn't effect all of the steps. The coverages misses out part of the code and reports wildly higher/lower coverage. Currently there's no warning or error that the coverage didn't upload.

If coverage did not upload then the tests must all be rerun as the coverage upload is directly coupled to this job.

**Desired behavior**
The report should only be generated if all coverage information is available, and if the codecov upload fails due to an intermittent error, we should be able to retry just the upload without a need to rerun any tests.

PR #2973, opened 2023-06-08
Gh-2972: Codecov can generate inconsistent coverage reports

### LINKED ISSUE #2971: Upgrade Spring version
We're not using the latest patch version available. It would also be good to move to a newer minor version if that can be done easily.

GCHQDeveloper314:
Compared to before this avoids making multiple uploads to codecov and allows the codecov upload to be reattempted with rerunning any tests.

# Related Issue

- Resolve #2971

### DIFF
--- a/core/store/pom.xml
+++ b/core/store/pom.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!--
-  ~ Copyright 2016-2020 Crown Copyright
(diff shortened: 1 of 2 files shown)


t92549 (dismissed):

t92549 (approved):

GCHQDeveloper314:
Branch name is wrong. This needs to be reopened as the branch cannot be renamed in-situ.

PR #2974, opened 2023-06-09
Gh-2972: Codecov can generate inconsistent coverage reports 

### LINKED ISSUE #2972: Codecov can generate inconsistent coverage reports

GCHQDeveloper314:
Replaces earlier PR with wrong branch name

# Related Issue

- Resolve #2972


GCHQDeveloper314:
The coverage on the base branch (`develop`) is incorrect. This PR doesn't change the actual coverage.
It fixes the way coverage is uploaded so hopefully we won't face this problem again!

t92549 (approved):

Issue #2977, opened 2023-06-14
GitHub Actions has deprecated actions using Node 12


GCHQDeveloper314:
**Describe the problem**
We are using some older versions of actions, or actions which have not been updated in a while. If these are using Node 12 then as of today they will no longer work. [See announcement: All Actions will run on Node16 instead of Node12 by default](https://github.blog/changelog/2023-06-13-github-actions-all-actions-will-run-on-node16-instead-of-node12-by-default/).

**To Reproduce**
Go to Actions and see previous warnings about this for some of our workflows.

**Desired behavior**
No warnings for any actions workflows.

GCHQDeveloper314:
The following Workflows use actions (`action-bump-semver` & `action-regex-match`) from [actions-ecosystem](https://github.com/actions-ecosystem) which do not have recent releases:

- Create Release Branch
- Create Hotfix Branch

From today these actions will be broken. Hopefully they will soon be updated. If this doesn't happen we'll need to remove these actions from the workflow. It might be better to replace them with use of `sed`.

GCHQDeveloper314:
The [actions-ecosystem](https://github.com/actions-ecosystem) actions appear to be unmaintained and haven't been updated - `action-bump-semver` can be replaced with `sed -r` and `action-regex-match` is easily replaced with `grep -E`. This has been done in the associated PR.

PR #2978, opened 2023-06-14
Gh-2977: GitHub Actions has deprecated actions using Node 12

GCHQDeveloper314:
PR is a draft as this is not a complete fix yet. See issue for why.


.github/workflows/release.yaml
```diff
@@ -43,7 +44,7 @@ jobs:
 
     - name: Output branch name
       id: branch
-      run: echo "::set-output name=branch_name::$BRANCH_NAME"
+      run: echo "{branch_name}={$BRANCH_NAME}" >> $GITHUB_OUTPUT
```
t92549, line 47: Haas this been tested on a fork or another workflow?

.github/workflows/create-release-branch.yaml
```diff
@@ -36,29 +36,12 @@ jobs:
...
-      uses: actions-ecosystem/action-bump-semver@v1
-      id: bump-semver
-      with:
-        current_version: ${{ env.CURRENT_VERSION }}
-        level: minor
-
-    - name: Set version automatically
-      if: ${{ !github.event.inputs.version }}
-      run: echo "RELEASE_VERSION=$(echo ${{ steps.bump-semver.outputs.new_version }} )" >> $GITHUB_ENV
+      run: echo "RELEASE_VERSION=$(echo ${{ env.CURRENT_VERSION }} | sed -r 's/([0-9]+)\.([0-9]+)\.([0-9]+)/echo \1.$((\2+1)).\3/' | sh)" >> $GITHUB_ENV
```
t92549, line 41: Does this sed command bump the minor version or patch version? I noticed it's the same as the hotfix command but the hotfix is meant to bump patch (2.0.0 -> 2.0.1) and release is meant to bump minor (2.0.0 -> 2.1.0)

t92549 (changes requested):

.github/workflows/release.yaml
GCHQDeveloper314: No, but it is taken directly [from the GitHub docs for migrating](https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/) away from `::set-output` which no longer works.

.github/workflows/create-release-branch.yaml
GCHQDeveloper314: Yes this is a mistake, forgot to change to be the patch. I'll add a commit fixing this.

t92549 (approved):

PR #2981, opened 2023-06-16
Gh-2942: Relocate code from main to test in integration-test

### LINKED ISSUE #2942: Relocate code from main to test in integration-test
**Describe the bug**
In a previous PR the core part of the `integration-test` module code was relocated to `main` from `test`. However this code depends on imports which are still part of `test`. This causes problems with code highlighting in some IDEs and prevents the Javadoc from being generated correctly. This is because the javadoc plugin only compiles `main` code and not `test` code (behaviour which cannot be changed), causing some imports to not be found.

**To Reproduce**
Steps to reproduce the bug:
1. Run `javadoc:aggregate`

**Expected behavior**
There should be no conflicts or improper dependencies between code in `main` and `test`.

GCHQDeveloper314:
Due to rename, various file have copyright updated. This difficult to avoid.

# Related Issue

- Resolve #2942


t92549 (approved):
This makes sense. Our extensions to JUnit are still in `main/` so they can be tested, but all the instances of integration tests that should run in Suites (like `GetElementsIT`) are now back in `test/`.

Issue #2997, opened 2023-06-29
NullPointerException occurs in MiniAccumuloStore if Zookeeper not set


GCHQDeveloper314:
**Describe the bug**
If the configuration for a `MiniAccumuloStore` (in `store.properties`) doesn't specify the property `accumulo.zookeepers` then a NullPointerException will occur on line 151. This is because `split()` is called on a `Null`, which causes this unchecked exception to be thrown. The purpose of this code is to get the port for Zookeeper.

https://github.com/gchq/Gaffer/blob/62d83317b8c25a0a6d4bfae1e822352266e7fec1/store-implementation/accumulo-store/src/test/java/uk/gov/gchq/gaffer/accumulostore/MiniAccumuloStore.java#L151

**Expected behavior**
No exception thrown and default Zookeeper port used.

**Additional context**
Adding a null check to the result of `getZookeepers()` will fix this.

oprajapati1:
I would like to be assigned to this and I can add the null check. Thanks

GCHQDeveloper314:
There's no need to be assigned, just go ahead and create a PR for this issue. Sorry for the broken contributors link in the README, [our ways of working are now detailed here](https://gchq.github.io/gaffer-doc/latest/dev/ways-of-working/).

oprajapati1:
Great, I believe I have solved the issue. To resolve this issue, I added a null check for the result of getZookeepers() before calling split(":") on it.

When getZookeepers() returns null, we now set the Zookeeper port to the default value. Otherwise, we extract the port from the getZookeepers() return value like before.

This fix makes sure that a valid Zookeeper port is always set and stops a NullPointerException from being thrown when the accumulo.zookeepers property is not specified.

I do not see any tests for createCluster, should I just commit and push and issue a pull request?

Banji-07:
I have resolved this issue and have created a pull request. 
Could you kindly review and address it when you have the chance?

GCHQDeveloper314:
> I do not see any tests for createCluster, should I just commit and push and issue a pull request?

This is a good point. This class is itself a library which is used exclusively by tests and so it doesn't currently have any tests of it's own. Though a test class could be added.

PR #2999, opened 2023-07-01
Gh-2997: Fix NullPointerException in MiniAccumuloStore

oprajapati1:
Details: Added null check for getZookeepers() to prevent NullPointerException when the accumulo.zookeepers property is not defined in the store.properties file. In case of a null return from getZookeepers(), the default Zookeeper port is now set.


GCHQDeveloper314:
Please run `mvn spotless:apply` to update the copyright header for the modified file, or do this manually. This is otherwise ready to merge.

GCHQDeveloper314:
Issue now resolved

PR #3003, opened 2023-07-04
Gh-3002: Ensure CancelScheduledJobHandler is only added when Jobs are enabled.

Banji-07:
Previously, the CancelScheduledJobHandler would throw an OperationException with the message 'JobTracker not enabled' if used when Jobs were not enabled. This issue arose because the handler was being added even when Jobs were not enabled. To resolve this, changes were made to ensure that the CancelScheduledJobHandler is only added when Jobs are enabled. This ensures that the handler functions correctly and prevents the unnecessary exception from being thrown.


Banji-07:
@GCHQDeveloper314 take a look at this.

Issue #2836, opened 2022-12-02
Create /graph/operations/all endpoint


t92549:
**Describe the new feature you'd like**
A `/graph/operations/all` endpoint should be added to return all operations on the classpath.

**Why do you want this feature?**
The `/graph/operations/details/all` was added to return the details of all operations on the classpath, but sometimes users don't want details, just the name of the operations.

**Additional context**
It would be nice if the endpoint returned them sorted alphabetically (see https://github.com/gchq/Gaffer/issues/2775).
For implementation, see the `/graph/operations/details/all` endpoint.

ibrijesh:
HI , Ian interested in contribution, I have already build the project using  mvn clean install -Pquick command , I wanted to know how to run these api end point.

GCHQDeveloper314:
Note that as with #2690 this is only for spring-rest.

PR #3008, opened 2023-07-16
Gh-2836: New operation names endpoint

Banji-07:
New "/graph/operations/all" endpoint for retrieving sorted operation names.

### DIFF
--- a/rest-api/common-rest/src/main/java/uk/gov/gchq/gaffer/rest/service/v2/AbstractOperationService.java
+++ b/rest-api/common-rest/src/main/java/uk/gov/gchq/gaffer/rest/service/v2/AbstractOperationService.java
@@ -53,17 +53,22 @@ public abstract class AbstractOperationService {
 
     protected abstract GraphFactory getGraphFactory();
 
-    public Set<Class <? extends Operation>> getSupportedOperations() {
-        return getGraphFactory().getGraph().getSupportedOperations();
-    }
-
-    public Set<String> getAllOperationsNames() {
-        Set<Class<? extends Operation>> operationClasses = new HashSet(ReflectionUtil.getSubTypes(Operation.class));
-        return operationClasses.stream().sorted(Comparator.comparing(Class::getSimpleName)).map(Class::getName).collect(Collectors.toCollection(LinkedHashSet::new));
+    public Set<Class<? extends Operation>> getSupportedOperations(final boolean includeUnsupported) {
+        Set<Class<? extends Operation>> operationsClasses;
+        if (includeUnsupported) {
+            operationsClasses = sortOperations(new HashSet(ReflectionUtil.getSubTypes(Operation.class)));
+        } else {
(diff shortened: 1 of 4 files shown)


rest-api/common-rest/src/main/java/uk/gov/gchq/gaffer/rest/service/v2/AbstractOperationService.java
```diff
@@ -54,6 +58,11 @@ public Set<Class <? extends Operation>> getSupportedOperations() {
         return getGraphFactory().getGraph().getSupportedOperations();
     }
 
+    public Set<String> getAllOperationsNames() {
```
GCHQDeveloper314, line 61: This should be the same type as `getSupportedOperations` (`Set<Class <? extends Operation>>`) and should therefore be called `getAllOperations`.

It would be better if instead this was merged into `getSupportedOperations` with a `final boolean includeUnsupported` used to decide if unsupported are included. As done for the Operation details.

rest-api/common-rest/src/main/java/uk/gov/gchq/gaffer/rest/service/v2/AbstractOperationService.java
```diff
@@ -63,10 +77,17 @@ public Set<OperationDetail> getSupportedOperationDetails(final boolean includeUn
         if (includeUnsupported) {
             operationClasses = new HashSet(ReflectionUtil.getSubTypes(Operation.class));
         } else {
-            operationClasses = getSupportedOperations();
+            operationClasses = getSupportedOperations(false);
         }
-        Set<OperationDetail> operationDetails = new HashSet<>();
-
+        Set<OperationDetail> operationDetails = new TreeSet<>((operationDetail1, operationDetail2) -> {
```
GCHQDeveloper314, line 82: This code is the same as what was in a previous version of your other PR. Order of operation details is not part of the ticket related to this PR. Could you revert any changes from this PR which overlap with the other one.

GCHQDeveloper314 (changes requested):
There's some overlap between this PR and another PR. This PR should only contain changes for the related issue and not any others.

I think this can be merged in first. The other PR can then be updated to include the changes from this one and build on them.

rest-api/common-rest/src/main/java/uk/gov/gchq/gaffer/rest/service/v2/AbstractOperationService.java
Banji-07: Is it safe to use an interactive rebase to fix this, or should I cherry pick ?

GCHQDeveloper314: I think you'll need to edit it manually. The commit history doesn't matter so much as it's all squashed prior to merging.

Banji-07: I just push some commits.

PR #3007, opened 2023-07-13
Gh-2775: Sort results for the OperationDetails endpoint

Banji-07:
The operation details endpoints now returns a consistent set, updated the endpoints tests and also the summary explaining each endpoint operations.

PR #3011, opened 2023-07-28
Gh-2836: New operation names endpoint

Banji-07:
New "/graph/operations/all" endpoint for retrieving sorted operation names.


GCHQDeveloper314 (changes requested):
This contains code relating to sorting from the now closed PR #3007. This PR should only be for the issue it is associated with.

GCHQDeveloper314 (approved):
I have made the required changes.

The code coverage for `AbstractOperationService` is zero because the way the test are done means this file is not itself tested, only the implementation are tested. This causes codecov warnings and is something which could only be fixed in a refactor and not in this PR.

Issue #2625, opened 2022-04-07
Improve assertions and use assertJ


t92549:
In many places in Gaffer, JUnit assertions are used and are quite ugly compared to the assertJ equivalent.
See the following example for testing lists:
**JUnit:**
https://github.com/gchq/Gaffer/blob/6a0c8636f71c47096fc3d99401688b8815a32933/core/operation/src/test/java/uk/gov/gchq/gaffer/operation/impl/get/GetElementsTest.java#L144
**assertJ:**
https://github.com/gchq/Gaffer/blob/6a0c8636f71c47096fc3d99401688b8815a32933/core/data/src/test/java/uk/gov/gchq/gaffer/data/graph/function/walk/ExtractWalkEdgesFromHopTest.java#L55

There are also many other occasions that assertJ can provide easier to read and stronger testing, like when using custom Conditions:
https://github.com/gchq/Gaffer/blob/370a7fa6e5fe99a9e5f06fa2a5e5b3d53db94339/store-implementation/accumulo-store/src/test/java/uk/gov/gchq/gaffer/accumulostore/integration/AccumuloMatchedVertexIT.java#L102

There are also places where assertJ assertions are used but could be cleaned up and easier to read, such as directly casting where `.asList()` could be used: 
https://github.com/gchq/Gaffer/blob/370a7fa6e5fe99a9e5f06fa2a5e5b3d53db94339/core/graph/src/test/java/uk/gov/gchq/gaffer/operation/export/resultcache/handler/GetGafferResultCacheExportHandlerTest.java#L133

These usages should be cleaned up, producing less warnings in tests, and making them easier to read.

GCHQDeveloper314:
To help with completing this ticket I've had a look through the codebase for where these improvements can be made.

Simplification of `assertEquals(Lists.newArrayList())`. Using `grep -rFIl "assertEquals(Lists.newArrayList" .` finds:
```
core/operation/src/test/java/uk/gov/gchq/gaffer/operation/impl/output/ToSetTest.java
core/operation/src/test/java/uk/gov/gchq/gaffer/operation/impl/output/ToStreamTest.java
core/operation/src/test/java/uk/gov/gchq/gaffer/operation/impl/output/ToMapTest.java
core/operation/src/test/java/uk/gov/gchq/gaffer/operation/impl/get/GetAdjacentIdsTest.java
core/operation/src/test/java/uk/gov/gchq/gaffer/operation/impl/add/AddElementsTest.java
[...]

PR #3014, opened 2023-08-20
gh-2625: Improve assertions readability with assertJ

AlfonsoEsposito:


core/common-util/src/test/java/uk/gov/gchq/gaffer/commonutil/iterable/CachingIterableTest.java
```diff
@@ -51,8 +50,9 @@ public void shouldCacheSmallIterable(@Mock final Iterable<Integer> mockIterable)
 
         final Iterable<Integer> cachingIterable = new CachingIterable<>(mockIterable, 5);
 
-        assertThat(SMALL_LIST).isEqualTo(Lists.newArrayList(cachingIterable));
-        assertThat(SMALL_LIST).isEqualTo(Lists.newArrayList(cachingIterable));
+        assertThat(cachingIterable)
+                .containsExactlyElementsOf(SMALL_LIST)
+                .containsExactlyElementsOf(SMALL_LIST);
```
GCHQDeveloper314, line 55: I'm not sure why this was duplicated previously, but I can't see a reason for it to remain. Same for the `LARGE_LIST ` assertion below.

GCHQDeveloper314 (changes requested):
This doesn't address everything required by the issue. The original issue comment only gave indicative examples of the work required and not which files need to be changed.

[I've added a new comment to the issue](https://github.com/gchq/Gaffer/issues/2625#issuecomment-1687723803) with further information and some hints on where changes need to be made.

core/common-util/src/test/java/uk/gov/gchq/gaffer/commonutil/iterable/CachingIterableTest.java
t92549: I think it is to test that the iterable doesn't get exhausted after being evaluated

t92549:
@GCHQDeveloper314 Although this will not close the issue, it does make progress towards and so I think can be merged

core/common-util/src/test/java/uk/gov/gchq/gaffer/commonutil/iterable/CachingIterableTest.java
GCHQDeveloper314: Ok, in that case how sure are we sure that these assertions can test for that?

GCHQDeveloper314 (approved):

t92549 (approved):

PR #3028, opened 2023-10-09
Gh-3025: Fix ApplyViewToElements Serialisation

### LINKED ISSUE #3025: The federated merge function 'ApplyViewToElementsFunction' does not serialise
**Describe the bug**
When running a `GetAllElements` or `GetElements` operation on a federated store using the REST API a server 500 error occurs. As these now use the `ApplyViewToElementsFunction` by default it appears to be the issue as a google guava type cannot be serialised.

**To Reproduce**
Run a `GetAllElements` operation on a federated graph via the REST API.

**Expected behavior**
The operation returns a federated list of elements

**Stack trace and errors**
Error output:
```json
{
    "statusCode": 500,
    "status": "Internal Server Error",
    "simpleMessage": "uk.gov.gchq.gaffer.exception.SerialisationException: (was java.lang.UnsupportedOperationException) (through reference chain: uk.gov.gchq.gaffer.federatedstore.operation.FederatedOperation[\"mergeFunction\"]->uk.gov.gchq.gaffer.federatedstore.util.ApplyViewToElementsFunction[\"requiredContextValues\"]->com.google.common.collect.RegularImmutableSet[3])"
}
```

**Additional context**
Likely due to missing Jackson module to help serialise the guava types. Adding the [Jackson datatype for guava](https://github.com/FasterXML/jackson-datatypes-collections/tree/master/guava) should hopefully fix it.

May also need to add the tag to the relevant method:
```java
@JsonDeserialize(as = ImmutableSet.class)
```

tb06904:
Fixes the serialisation issues the merge function `ApplyViewToElements` had due to using the google guava types. As we are using Java 8+ these can be swapped out for standard types from the `Collections` library (essentially just serialises as a Map now).

Could be some further work to add some unit testing for serialisation as this seems to be a bit of an hole in testing where only the Java API is tested a lot of the time, I will leave that to a future ticket to keep this a quick fix.

# Related issue

- Resolve #3025


GCHQDeveloper314 (commented):
Out of interest, are the new Jackson annotations actually required?

t92549 (approved):

tb06904:
> Out of interest, are the new Jackson annotations actually required?

They do appear to be, without them it hits the same error. Serialising the `Collections` classes appears to be a bit of an edge case in Jackson the annotations allow Jackson to fallback and use a standard Map to serialise with (which is what the context is initialised in the class as) the alternative is to write custom deserialiser which is a bit overkill.

n3101:
@tb06904 " I will leave that to a future ticket to keep this a quick fix."

Fine. Please submit that ticket before it is forgotten

tb06904:
> @tb06904 " I will leave that to a future ticket to keep this a quick fix."
> 
> Fine. Please submit that ticket before it is forgotten

@n3101 #3032 has been raise to cover

GCHQDeveloper314 (approved):

PR #3035, opened 2023-10-16
Gh-2794: Make Kerberos support more configurable

### LINKED ISSUE #2794: Make Kerberos support more configurable
I want to have more control over Kerberos configurations when using Gaffer. This includes being able to set all minor configuration parameters that Accumulo supports. It also includes overriding an existing Kerberos login which is otherwise automatically picked up by Hadoop UGI in the code.

If the system which Gaffer is running on has already logged into Kerberos, then the keytab I supply is not used and the existing credentials are reused instead. If I want to use different credentials from the ones already in use on the system, there's no way to configure Gaffer for this. This is the relevant section of the code:
https://github.com/gchq/Gaffer/blob/e892bbe5b6b3e1b3c4fcb97421de33839c1d6c0e/store-implementation/accumulo-store/src/main/java/uk/gov/gchq/gaffer/accumulostore/utils/TableUtils.java#L249-L250
I also need to set other parameters between the Accumulo client and cluster before I can establish a connection.

We've considered changing the system config but this is not possible because of our environment.

GCHQDeveloper314:
# Related issue

- Resolve #2794


store-implementation/accumulo-store/src/main/java/uk/gov/gchq/gaffer/accumulostore/utils/TableUtils.java
```diff
@@ -244,13 +245,14 @@ public static Connector getConnectorKerberos(final String instanceName, final St
                 conf.set("hadoop.security.authorization", "true");
                 UserGroupInformation.setConfiguration(conf);
             }
-            // If already logged in with Keytab, then check if ticket needs renewal, else do initial login
-            if (UserGroupInformation.isLoginKeytabBased()) {
+            // If initial login is complete and logged in using Keytab, then check if ticket needs renewal, else do initial login
+            if (initialKerberosLoginComplete && UserGroupInformation.isLoginKeytabBased()) {
```
tb06904, line 249: More out of curiosity but would [`isFromKeytab()`](https://hadoop.apache.org/docs/stable/api/org/apache/hadoop/security/UserGroupInformation.html#isFromKeytab--) or [`hasKerberosCredentials()`](https://hadoop.apache.org/docs/stable/api/org/apache/hadoop/security/UserGroupInformation.html#hasKerberosCredentials--) be relevant here/replace this boolean?

GCHQDeveloper314: They wouldn't. It's explained some more in the issue, but briefly the requirement is to ignore any existing login which may have occurred and ensure that a login is done with the credentials supplied in the Gaffer config.

The methods you reference are for instances of UGI, whereas Accumulo uses the static UGI methods. The [`isLoginKeytabBased()`](https://hadoop.apache.org/docs/stable/api/org/apache/hadoop/security/UserGroupInformation.html#isLoginKeytabBased--) is an equivalent.

If a separate keytab login has been done on the system using different credentials prior to starting Gaffer, then the UGI method will return true and login would not happen with the credentials given to Gaffer. As Gaffer is given creds they should always be used for establishing the connection, that's what this new boolean ensures.

tb06904: Nice that makes sense to ignore any existing login

tb06904 (approved):

t92549 (commented):
CodeCov not happy with coverage, but presumably this cannot be tested in unit tests and needs testing in gaffer-docker? Has this been tested manually with gaffer-docker?

GCHQDeveloper314:
> CodeCov not happy with coverage, but presumably this cannot be tested in unit tests and needs testing in gaffer-docker? Has this been tested manually with gaffer-docker?

Yes, this has been confirmed as working as expected. It's not possible to test the Kerberos support with unit/integration tests unfortunately - at least not without substantial changes to Accumulo.

t92549 (approved):

PR #3049, opened 2023-10-23
Gh-3045: Merge schema ignoring exceptions

### LINKED ISSUE #3045: MergeSchema is ignoring SchemaExceptions
**Describe the bug**
When the `MergeSchema` class applies the `Schema.BaseBuilder.merge` function, it catches `SchemaException`s. This is because it will retry while ignoring vertex serialisers if the error is to do with vertex serialisers:
https://github.com/gchq/Gaffer/blob/e90fba0cb33ca23ab60cd1e2c175dfd29801fde1/store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/util/MergeSchema.java#L56-L65
However, if the `SchemaException` is unrelated to vertex serialisers, I would expect the exception to still be thrown:
```java
if (e.getMessage().contains(Schema.UNABLE_TO_MERGE_SCHEMAS_CONFLICT_WITH_VERTEX_SERIALISER_OPTIONS_ARE)) {
    // handle the vertex serialiser issue...
} else {
    // throw other types of SchemaException
    throw e
}
```
This is not the case.

**Additional context**
If you simply add this else block in, tests will start to fail. These tests are [FederatedStoreVisibilityTest](https://github.com/gchq/Gaffer/blob/e90fba0cb33ca23ab60cd1e2c175dfd29801fde1/store-implementation/federated-store/src/test/java/uk/gov/gchq/gaffer/federatedstore/FederatedStoreVisibilityTest.java) (issue: https://github.com/gchq/Gaffer/issues/1834) and [FederatedStoreSchemaOverlapTest](https://github.com/gchq/Gaffer/blob/e90fba0cb33ca23ab60cd1e2c175dfd29801fde1/store-implementation/federated-store/src/test/java/uk/gov/gchq/gaffer/federatedstore/FederatedStoreSchemaOverlapTest.java) (issue: https://github.com/gchq/Gaffer/issues/2767). These tests were added as part of Gaffer 2 development to codify the new FederatedStore behaviour, however, we did not realise this behaviour was happening as a result of a bug.

**Suggested fix**
`MergeSchema` should be fixed and extended to manage all types of SchemaException.
When a SchemaException happens due to conflicting visibility property names, this can probably be intentionally ignored with a warning message.
When a SchemaException happens due to conflicting types for the same property in the same group across two subgraphs, things are more complicated. Perhaps, if the type definitions are the same the error can be ignored, and if not they can be de-duplicated. 
This will likely need work on `Schema.BaseBuilder.validateSharedGroupsAreCompatible`

GCHQDev404:
# Related issue

- Resolve #3045


store-implementation/federated-store/src/test/java/uk/gov/gchq/gaffer/federatedstore/FederatedStoreSchemaOverlapTest.java
```diff
@@ -170,17 +181,12 @@ public void shouldOverwriteTypeDefinitionWhenConflictingSchemasMerge() throws Op
...
-        assertThat(schema.validate().isValid())
-                .withFailMessage(schema.validate().getErrorString())
-                .isTrue();
-        assertThat(schema.getType(INTEGER)).isEqualTo(STRING_TYPE);
-        assertThat(schema.toString())
-                .isNotEqualTo(GRAPH_A_SCHEMA.toString())
-                .isEqualTo(graphBSchema.toString());
+        // When then
+        assertThatException()
+                .isThrownBy(() -> federatedStore.execute(new GetSchema.Builder().build(), testContext))
```
GCHQDeveloper314, line 186: With these, it might be an idea to include a check on the type of exception thrown. To ensure these tests are revisited if how the exception is thrown gets changed.

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/FederatedStore.java
```diff
@@ -413,7 +413,7 @@ public List<GraphSerialisable> getGraphs(final User user, final List<String> gra
                 rtn = getDefaultGraphs(user, operation);
             } else {
                 if (graphIds.isEmpty()) {
-                    LOGGER.info("A get graphs request was made with empty graphIds");
+                    LOGGER.debug("A get graphs request was made with empty graphIds");
```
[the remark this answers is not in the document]
GCHQDev404: Yeah, there was complaint about seeing it so much, so I dropped it down a level. It was useful when building the Federation, but not so much every day.

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/operation/handler/impl/FederatedOperationHandler.java
```diff
@@ -105,7 +108,8 @@ private Object mergeResults(final Iterable resultsFromAllGraphs, final Federated
 
             return rtn;
         } catch (final Exception e) {
-            throw new OperationException(String.format("Error while merging results. %s", Objects.toString(e.getMessage(), "")), e);
+            final Object graphs = context.getVariable(MERGE_FUNCTION_GET_GRAPHS);
+            throw new OperationException(String.format("Error while merging results from graphs: %s due to: %s", graphs.toString(), Objects.toString(e.getMessage(), "")), e);
```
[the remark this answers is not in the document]
GCHQDev404: Yeah, I have no memory of using it, but it was there before. seems maybe better that having a "ternary" or "if null"
🤷‍♂️

core/store/src/main/java/uk/gov/gchq/gaffer/store/schema/Schema.java
```diff
@@ -72,7 +72,14 @@
...
+    public static final String FORMAT_EXCEPTION = "%s, options are: %s and %s";
+    public static final String FORMAT_UNABLE_TO_MERGE_SCHEMAS_CONFLICT_WITH_S = "Unable to merge schemas because of conflict with the %s";
+    public static final String VERTEX_SERIALISER = "vertex serialiser";
+    public static final String SCHEMAS_CONFLICT_WITH_VERTEX_SERIALISER = String.format(FORMAT_UNABLE_TO_MERGE_SCHEMAS_CONFLICT_WITH_S, VERTEX_SERIALISER);
+    public static final String VISIBILITY_PROPERTY = "visibility property";
+    public static final String SCHEMAS_CONFLICT_WITH_VISIBILITY_PROPERTY = String.format(FORMAT_UNABLE_TO_MERGE_SCHEMAS_CONFLICT_WITH_S, VISIBILITY_PROPERTY);
+    public static final String ELEMENT_GROUP_MUST_ALL_BE_DEFINED_IN_A_SINGLE_SCHEMA = "Element group properties cannot be defined in different schema parts, they must all be defined in a single schema part. Please fix this group: ";
+    public static final String FORMAT_ERROR_WITH_THE_SCHEMA_TYPE_NAMED_S_DUE_TO_S = "Error with the schema type named:%s due to: %s";
```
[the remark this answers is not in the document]
GCHQDev404: They get used in another class.

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/util/MergeSchema.java
```diff
@@ -48,20 +60,52 @@ public Schema apply(final Schema update, final Schema state) {
...
+                    context.put(WIPE_VISIBILITY_PROPERTY, true);
+                    mergeSchema.merge(new Schema.Builder(update)
+                            .visibilityProperty(null)
+                            .build());
+                    mergeSchema.visibilityProperty(null);
+                } else if (e.getMessage().contains(SCHEMAS_CONFLICT_WITH_TYPE_CLASS)
+                        || e.getMessage().contains(SCHEMAS_CONFLICT_WITH_TYPE_SERIALISER)
+                        || e.getMessage().contains(SCHEMAS_CONFLICT_WITH_AGGREGATE_FUNCTION)) {
+                    // Not possible to resolve from these collisions.
+                    throw new SchemaException(String.format(FORMAT_MERGE_FUNCTION_UNABLE_TO_RECOVER_FROM_ERROR_DUE_TO_S, e.getMessage()), e);
```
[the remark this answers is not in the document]
GCHQDev404: The error message is more informative for the error on line 105.
Yeah there was a hope to be able to fix this, but it's currently not possible without new feature and behaviour requests. Altho verbose I thought it was good for this code to remain, to demonstrate that is specific state that can't be recovered.

maybe I change error message on line 108 to something like that does NOT say "unable to recover" because it is a catch all, the exception isn't as expected as the others?

🤷‍♂️

store-implementation/federated-store/src/test/java/uk/gov/gchq/gaffer/federatedstore/FederatedStoreSchemaOverlapTest.java
GCHQDev404: I was going to, but then I didn't think it matters if it gets caught and re-thrown. The value is in the stack trace messages. You think is valuable? I did consider it.

GCHQDeveloper314: It depends. I was thinking if someone edits the exception type then it would probably be a good idea to revisit the tests at the same time. Checking the exception type will ensure that happens. This might not be something we want to do as it's not strictly important what the type is.

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/util/MergeSchema.java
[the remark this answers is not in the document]
GCHQDeveloper314: Yes, a different error message would be good. This would not be duplicating anything then.

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/operation/handler/impl/FederatedOperationHandler.java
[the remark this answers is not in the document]
GCHQDeveloper314: I don't think this is required at all. `String.format` is null safe. If `e.getMessage()` is `null`, it will automatically replace it with empty string, we don't need to do this.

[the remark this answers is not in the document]
GCHQDev404: I guess in a situation when the object "e" could throw a null pointer this is safe. but in a Catch e will never be null. 😆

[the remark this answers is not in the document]
GCHQDev404: I think it's for tidyness. Some exceptions might not have message like NullPointer.

store-implementation/federated-store/src/test/java/uk/gov/gchq/gaffer/federatedstore/FederatedStoreSchemaOverlapTest.java
GCHQDev404: It throws SchemaExceptions and by the time the test catches it I think it becomes an Operation Exception. Due to the many rethrows and adding more info to exceptions.

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/operation/handler/impl/FederatedOperationHandler.java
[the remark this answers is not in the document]
GCHQDev404: done

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/util/MergeSchema.java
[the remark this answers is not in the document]
GCHQDev404: done

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/FederatedStore.java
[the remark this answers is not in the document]
t92549: As an aside, this happens more than 20 times when starting the federated store, I suppose this is normal but seems really high

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/operation/handler/impl/FederatedOperationHandler.java
```diff
@@ -61,6 +62,7 @@ private Iterable getAllGraphResults(final FederatedOperation<INPUT, OUTPUT> oper
         try {
             List<Object> results;
             final Collection<GraphSerialisable> graphs = getGraphs(operation, context, store);
+            context.setVariable(MERGE_FUNCTION_GET_GRAPHS, graphs.stream().map(GraphSerialisable::getGraphId).collect(Collectors.toList()));
```
t92549, line 65: I am a little confused by this change

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/util/MergeSchema.java
```diff
@@ -30,10 +30,22 @@
...
+import static uk.gov.gchq.gaffer.store.schema.Schema.SCHEMAS_CONFLICT_WITH_VISIBILITY_PROPERTY;
+import static uk.gov.gchq.gaffer.store.schema.Schema.VERTEX_SERIALISER;
+import static uk.gov.gchq.gaffer.store.schema.Schema.VISIBILITY_PROPERTY;
+import static uk.gov.gchq.gaffer.store.schema.TypeDefinition.SCHEMAS_CONFLICT_WITH_AGGREGATE_FUNCTION;
+import static uk.gov.gchq.gaffer.store.schema.TypeDefinition.SCHEMAS_CONFLICT_WITH_TYPE_CLASS;
+import static uk.gov.gchq.gaffer.store.schema.TypeDefinition.SCHEMAS_CONFLICT_WITH_TYPE_SERIALISER;
 
 public class MergeSchema implements BiFunction<Schema, Schema, Schema>, ContextSpecificMergeFunction<Schema, Schema, Schema> {
     public static final String WIPE_VERTEX_SERIALISERS = "wipe_vertex_serialisers";
+    public static final String WIPE_VISIBILITY_PROPERTY = "wipe_visibility_property";
```
t92549, line 44: Shouldn't these be camelCase rather than snake_case?

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/util/MergeSchema.java
```diff
@@ -48,20 +60,52 @@ public Schema apply(final Schema update, final Schema state) {
...
-            } catch (final SchemaException e) {
-                if (e.getMessage().contains(Schema.UNABLE_TO_MERGE_SCHEMAS_CONFLICT_WITH_VERTEX_SERIALISER_OPTIONS_ARE)) {
-                    LOGGER.error(String.format("Caught SchemaException, attempting to re-merge but without vertex serialisers. Error message:%s", e.getMessage()));
+                mergeSchema.merge(new Schema.Builder(update)
+                        .vertexSerialiser(vertexSerialiser)
+                        .visibilityProperty(visibilityProperty)
+                        .build());
+
+            } catch (final Exception e) {
+                if (e.getMessage().contains(SCHEMAS_CONFLICT_WITH_VERTEX_SERIALISER)) {
```
t92549, line 82: Maybe rather than all these `if (e.getMessage().contains(CONSTANT)`, there could be different types of schema exception that are caught?

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/operation/handler/impl/FederatedOperationHandler.java
GCHQDev404: I'm retaining this info into context, so I can smuggle the graphs info for a more meaningful error message if required later in the flow.

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/util/MergeSchema.java
GCHQDev404: I'm don't it matters for values. easy to read 🤷

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/util/MergeSchema.java
GCHQDev404: I'm not sure I agree with this. I don't want to create 4-6 new Exception classes just for Info regarding a particular type of SchemaException.

tb06904: Having multiple exception classes is pretty widely accepted as the 'correct' way to do exception handling in java over things like return codes as you can then catch and handle each one as needed with the `catch` block to give a clear program flow e.g.
```java
} catch (final CustomExceptionOne e1) {
...
} catch (final CustomExceptionTwo e2) {
...
} catch (final Exception e3) {
...
}
```
You could also enforce custom formatting or params for your exceptions to reduce some of the extra string formatting you are doing when throwing them.

store-implementation/federated-store/src/test/java/uk/gov/gchq/gaffer/federatedstore/FederatedStoreSchemaOverlapTest.java
```diff
@@ -154,14 +160,19 @@ public void shouldOverwriteVisibilityPropertyWhenConflictingSchemasMerge() throw
...
                 .withFailMessage(schema.validate().getErrorString())
                 .isTrue();
-        assertThat(schema.getVisibilityProperty()).isEqualTo(PROPERTY_1);
+        assertThat(schema.getVisibilityProperty())
+                .withFailMessage("After a recovery from a collision with Visibility Property, the property should be set to Null")
+                .isNull();
         assertThat(schema.toString())
-                .isNotEqualTo(GRAPH_A_SCHEMA.toString())
-                .isEqualTo(graphBSchema.toString());
+                .isEqualTo(new Schema.Builder(GRAPH_A_SCHEMA)
```
tb06904, line 167: Use the built in to string comparison for `assertThat`
```java
assertThat(schema).hasToString(...);
```

store-implementation/federated-store/src/test/java/uk/gov/gchq/gaffer/federatedstore/FederatedStoreSchemaTest.java
```diff
@@ -324,11 +325,36 @@ public void shouldChangeSchemaWhenAddingGraphB() throws OperationException {
...
+        // Given
+        addGraphWith(GRAPH_ID_B, STRING_REQUIRED_TYPE, PROPERTY_1, PROPERTY_2);
+
         // When
         final Schema schemaAB = federatedStore.getSchema(testContext, false);
 
         // Then
         assertThat(schemaAB).isNotEqualTo(schemaA);
+        assertThat(schemaAB.getEdge(GROUP_BASIC_EDGE).getProperties()).contains(PROPERTY_1);
         assertThat(schemaAB.getEdge(GROUP_BASIC_EDGE).getProperties()).contains(PROPERTY_2);
```
tb06904, line 358:
```suggestion
        assertThat(schemaAB.getEdge(GROUP_BASIC_EDGE).getProperties())
            .contains(PROPERTY_1)
            .contains(PROPERTY_2);
```

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/util/MergeSchema.java
t92549: It seems Gaffer does have a convention for these variables. `They_Look_Like_This`:
https://github.com/gchq/Gaffer/blob/7ee3762b3ccaf77004eace8dcc999b7d3c094f03/store-implementation/accumulo-store/src/main/java/uk/gov/gchq/gaffer/accumulostore/utils/AccumuloStoreConstants.java#L23-L30

core/store/src/main/java/uk/gov/gchq/gaffer/store/schema/Schema.java
```diff
@@ -331,21 +338,22 @@ public CHILD_CLASS merge(final Schema schema) {
                 Schema thatSchema;
```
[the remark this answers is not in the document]
t92549: I agree, I find guard clauses way more readable

core/store/src/main/java/uk/gov/gchq/gaffer/store/schema/Schema.java
```diff
@@ -480,39 +492,43 @@ private Schema getThisSchema() {
             return getElementDefs();
         }
 
-        private void validateSharedGroupsAreCompatible(final Map<String, ? extends SchemaElementDefinition> elements1, final Map<String, ? extends SchemaElementDefinition> elements2) {
-            final Set<String> sharedGroups = new HashSet<>(elements1.keySet());
-            sharedGroups.retainAll(elements2.keySet());
+        private void validateSharedGroupsAreCompatible(final Map<String, ? extends SchemaElementDefinition> thisElements, final Map<String, ? extends SchemaElementDefinition> thatElements, final String thisVisibilityProperty, final String thatVisibilityProperty) {
```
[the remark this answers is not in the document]
t92549: We don't currently have a check like this in the CI pipeline but it would be nice

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/operation/handler/impl/FederatedOperationHandler.java
GCHQDev404: Ah yeah. Originally this exception was getting thrown else where, but it's been pulled up and up so. So now that it is actually back in the same class, so I probably could just be set to a field variable now :D

GCHQDev404:
``` java
    @JsonIgnore
    private List<String> errorMessageGraphIds; // large scope and exclusion from JSON just to improve an error message. 
```

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/util/MergeSchema.java
GCHQDev404: I'll sprinkle upper case into the snake then.

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/util/MergeSchema.java
GCHQDev404: Adding the 4-6 new Exception Classes I don't believe will help because of re-throwing. 
 I don't receive the same type of class that was thrown. I would have to create new classes and then navigate down the `e.getCause()` path.

See: 
[Schema ](https://github.com/gchq/Gaffer/blob/f63eca2053be7ecf592a7899c16db1fa1c21d239/core/store/src/main/java/uk/gov/gchq/gaffer/store/schema/Schema.java#L430)

https://github.com/gchq/Gaffer/blob/f63eca2053be7ecf592a7899c16db1fa1c21d239/core/store/src/main/java/uk/gov/gchq/gaffer/store/schema/Schema.java#L355

GCHQDev404: at [TypeDef L170](https://github.com/gchq/Gaffer/blob/f63eca2053be7ecf592a7899c16db1fa1c21d239/core/store/src/main/java/uk/gov/gchq/gaffer/store/schema/TypeDefinition.java#L170) I am currently throwing a SchemaException even though I'm only merging TypeDefinitions.

maybe I could remove Schema L355 and Schema L430 and provide more information at TypeDef 170.

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/util/MergeSchema.java
GCHQDev404: please examine https://github.com/gchq/Gaffer/pull/3051 as to why I don't think it will work.

I loose adding information about types, and I am unable to catch it in the MergeSchema.

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/util/MergeSchema.java
GCHQDev404: please examine https://github.com/gchq/Gaffer/pull/3051 as to why I don't think it will work.

I loose adding information about types, and I am unable to catch it in the MergeSchema.

core/store/src/main/java/uk/gov/gchq/gaffer/store/schema/Schema.java
```diff
@@ -418,7 +424,11 @@ private void mergeTypes(final Schema thatSchema) {
                     if (null == typeDef) {
                         getThisSchema().types.put(newType, newTypeDef);
                     } else {
-                        typeDef.merge(newTypeDef);
+                        try {
+                            typeDef.merge(newTypeDef);
+                        } catch (final Exception e) {
+                            throw new GafferRuntimeException(String.format(FORMAT_ERROR_WITH_THE_SCHEMA_TYPE_NAMED_S_DUE_TO_S, entry.getKey(), e.getMessage()), e);
```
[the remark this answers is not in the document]
GCHQDev404: I don't think so, this will be a catch all, so if someone next week introduces a NullPointer. You'll know which schema type threw it.

core/store/src/main/java/uk/gov/gchq/gaffer/store/schema/Schema.java
```diff
@@ -480,39 +492,43 @@ private Schema getThisSchema() {
...
 
                     // Check if just one group has the properties and groupBy fields set.
-                    final SchemaElementDefinition elementDef1 = elements1.get(sharedGroup);
-                    if ((null == elementDef1.properties || elementDef1.properties.isEmpty())
-                            && elementDef1.groupBy.isEmpty()) {
+                    final SchemaElementDefinition thisElementDef = thisElements.get(sharedGroup);
+
+                    if ((null == thisElementDef.properties || thisElementDef.properties.isEmpty())
+                            && thisElementDef.groupBy.isEmpty()) {
                         continue;
```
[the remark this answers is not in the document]
GCHQDev404: I agree and I HATE multiple exits. This was pre-existing to this PR.

core/store/src/main/java/uk/gov/gchq/gaffer/store/schema/Schema.java
```diff
@@ -72,7 +72,14 @@
 @JsonDeserialize(builder = Schema.Builder.class)
 @JsonPropertyOrder(value = {"class", "edges", "entities", "types"}, alphabetic = true)
 public class Schema extends ElementDefinitions<SchemaEntityDefinition, SchemaEdgeDefinition> implements Cloneable {
-    public static final String UNABLE_TO_MERGE_SCHEMAS_CONFLICT_WITH_VERTEX_SERIALISER_OPTIONS_ARE = "Unable to merge schemas. Conflict with vertex serialiser, options are: ";
+    public static final String FORMAT_EXCEPTION = "%s, options are: %s and %s";
```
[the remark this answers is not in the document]
GCHQDev404: Not sure, only 5 out of 13 thrown exceptions use this wording style.

Please see https://github.com/gchq/Gaffer/pull/3051 why having new exception classes I don't think is optimal.

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/operation/handler/impl/FederatedOperationHandler.java
GCHQDev404: Spoken about and resolved

store-implementation/federated-store/src/test/java/uk/gov/gchq/gaffer/federatedstore/FederatedStoreSchemaOverlapTest.java
GCHQDev404: ooooh nice! 👌

store-implementation/federated-store/src/test/java/uk/gov/gchq/gaffer/federatedstore/FederatedStoreSchemaTest.java
GCHQDev404: lol

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/util/MergeSchema.java
GCHQDev404: https://github.com/gchq/Gaffer/pull/3051

has now been improved and satisfies the concerns I had about untidy code and re-throwing many similar exceptions.

core/store/src/main/java/uk/gov/gchq/gaffer/store/schema/Schema.java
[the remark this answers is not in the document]
GCHQDev404: #3051 has been merged.

store-implementation/federated-store/src/test/java/uk/gov/gchq/gaffer/federatedstore/FederatedStoreSchemaOverlapTest.java
```diff
@@ -154,14 +160,19 @@ public void shouldOverwriteVisibilityPropertyWhenConflictingSchemasMerge() throw
...
                 .isTrue();
-        assertThat(schema.getVisibilityProperty()).isEqualTo(PROPERTY_1);
-        assertThat(schema.toString())
-                .isNotEqualTo(GRAPH_A_SCHEMA.toString())
-                .isEqualTo(graphBSchema.toString());
+        assertThat(schema.getVisibilityProperty())
+                .withFailMessage("After a recovery from a collision with Visibility Property, the property should be set to Null")
+                .isNull();
+        assertThat(schema)
+                .hasToString(new Schema.Builder(GRAPH_A_SCHEMA)
```
t92549, line 167: Rather than checking the `toStrings` would it make more sense to assert they are the same?

GCHQDev404: yes, but I think when this errors it shows you the strings, which is nice. I think that's why it's there.

core/data/src/main/java/uk/gov/gchq/gaffer/data/elementdefinition/exception/SchemaException.java
```diff
@@ -27,6 +27,8 @@
  */
 public class SchemaException extends GafferRuntimeException {
     private static final long serialVersionUID = 3150434301320173603L;
+    @SuppressWarnings("PMD.AvoidStringBufferField")
+    private StringBuilder prependToMessage;
```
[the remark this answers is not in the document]
GCHQDev404: string1+string2+string3+string4 is really bad in Java. <-- you end up with 8 strings in memory, I think.

I think the warning is for if this field persisted for a long time. Exceptions normally die quickly.

store-implementation/federated-store/src/test/java/uk/gov/gchq/gaffer/federatedstore/FederatedStoreSchemaOverlapTest.java
```diff
@@ -138,12 +139,17 @@ public void shouldSetVertexSerialiserToNullWhenConflictingSchemasMerge() throws
     }
 
     @Test
-    public void shouldOverwriteVisibilityPropertyWhenConflictingSchemasMerge() throws OperationException {
+    public void shouldMergeSchemaWithoutTheClashingVisibilityProperty() throws OperationException {
         // Given
         addGraphWith(GRAPH_ID_A, GRAPH_A_SCHEMA);
+        final String visibility1 = "visibility1";
```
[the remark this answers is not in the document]
GCHQDev404: but 😆 that's the wrong number 🤡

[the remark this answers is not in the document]
t92549: The actual number doesn't matter, we just need 2 properties that are different

t92549 (approved):

tb06904 (approved):

PR #3050, opened 2023-10-23
Gh-3018: Migrate GraphConfig code out of Graph

### LINKED ISSUE #3018: Migrate GraphConfig code out of Graph
Refactor `Graph.java` to migrate methods and functionality relating `GraphConfig` into `GraphConfig.java`.

For example, the `getGraphHooks()` method creates a list of hook classes used in a instance of `GraphConfig`. This could easily be moved to `GraphConfig` itself.

If this was moved, then `hasHook` could also be moved. There's also other code which can be moved to `GraphConfig`, such as the `updateStoreProperties` method which can also be simplified.

Refactoring in this way will reduce the amount of code in `Graph.java` which is a very large class of approx 1000 lines. It would put all utility functions relating to `GraphConfig` in the `GraphConfig` class, making the codebase easier to understand and use.

Keeping everything as it is now. Not preferable because this isn't best practice.

**Additional context**
[PR comment where this theme was originally identified](https://github.com/gchq/Gaffer/pull/3017/files#r1308667392). PR #3017 also introduces code which interacts with `GraphConfig`. This ticket shouldn't be closed until that PR is merged and any changes taken into account.

tb06904:
Fair bit of refactoring to the Graph and GraphConfig classes this lead to some tweaks to testing as it highlighted some bugs.

As an overview the following methods have been altered:

*Graph.java*
* `updateSchema()` -> broken into `applyParentSchemas()` and `loadSchemaFromJson()` and refactored.
* `updateStore()` -> broken into `applyParentStoreProperties()` and `initStore()` and refactored.
* `updateView()` -> migrated to `GraphConfig.java` and renamed `initView()`.
* `validateAndUpdateGetFromCacheHook()` -> migrated to `GraphConfig.java` and heavily refactored. During refactoring a slight oddity was found that I couldn't get to the bottom of where graph hook seems to always have to be added to index 0 of the list of hooks else tests fail. See line 229 in `GraphConfig.java`.

*GraphConfig.java*
* Added logging.
* General code smells addressed and minor refactoring.

*Other*
* Added custom exception classes for graph hook errors and refactored tests to look for them.
* Use assertJ where possible.
* Javadocs

# Related issue

- Resolve #3018


core/graph/src/test/java/uk/gov/gchq/gaffer/operation/export/resultcache/handler/ExportToGafferResultCacheHandlerTest.java
```diff
@@ -132,7 +132,7 @@ public void shouldHandleOperationByDelegatingToAnNewExporter() throws OperationE
         // Then
         assertSame(handlerResult, results);
         final ArgumentCaptor<OperationChain> opChain = ArgumentCaptor.forClass(OperationChain.class);
-        verify(cacheStore, times(3)).execute(opChain.capture(), Mockito.any(Context.class));
+        verify(cacheStore, atLeast(1)).execute(opChain.capture(), Mockito.any(Context.class));
```
GCHQDeveloper314, line 135: What's the significance of this change?

core/graph/src/main/java/uk/gov/gchq/gaffer/graph/GraphConfig.java
```diff
@@ -188,21 +324,25 @@ public Builder json(final byte[] bytes) {
...
-                }
-                if (null != config.getLibrary() && !(config.getLibrary() instanceof NoGraphLibrary)) {
-                    this.config.setLibrary(config.getLibrary());
-                }
-                if (null != config.getDescription()) {
-                    this.config.setDescription(config.getDescription());
-                }
-                this.config.getHooks().addAll(config.getHooks());
+            if (config == null) {
+                LOGGER.warn("Unable to merge with a null config ignoring call");
```
GCHQDeveloper314, line 328:
```suggestion
                LOGGER.warn("Unable to merge GraphConfig with a null config, ignoring call");
```

core/graph/src/main/java/uk/gov/gchq/gaffer/graph/Graph.java
```diff
@@ -854,199 +862,151 @@ public Graph build() {
...
+        /**
+         * Initialises the {@link Store} class instance with a given graph configuration.
+         * <p>
+         * Will also attempt to extract some configuration from the current store if
+         * not currently configured.
+         *
+         * @param config The graph config
+         */
+        private void initStore(final GraphConfig config) {
+            // Init the store if needed
```
GCHQDeveloper314, line 971: I don't think this comment is required, the Javadoc covers this.

core/graph/src/main/java/uk/gov/gchq/gaffer/graph/Graph.java
```diff
@@ -854,199 +862,151 @@ public Graph build() {
...
+            if (store == null) {
+                LOGGER.debug("Store currently null initialising with Id: {} and existing schema/properties", config.getGraphId());
                 store = Store.createStore(config.getGraphId(), cloneSchema(schema), properties);
-            } else if ((null != config.getGraphId() && !config.getGraphId().equals(store.getGraphId()))
-                    || (null != schema)
-                    || (null != properties && !properties.equals(store.getProperties()))) {
```
GCHQDeveloper314, line 1013: This logic has changed, so that these `if`s are now run unconditionally?

core/graph/src/main/java/uk/gov/gchq/gaffer/graph/GraphConfig.java
```diff
@@ -116,25 +133,144 @@ public void setHooks(final List<GraphHook> hooks) {
...
+            if (!file.exists()) {
+                throw new IllegalArgumentException("Unable to find graph hook file: " + file.toString());
             }
+            try {
+                hooks.add(JSONSerialiser.deserialise(FileUtils.readFileToByteArray(file), GraphHook.class));
+            } catch (final IOException e) {
+                throw new IllegalArgumentException("Unable to deserialise graph hook from file: " + file.toString(), e);
+            }
+        } else {
+            LOGGER.debug("Adding Hook {}", hook.getClass().getSimpleName());
```
GCHQDeveloper314, line 159: Why log here but not when added from a path? I'd suggest also adding something there.

GCHQDeveloper314 (commented):
I think there are a few bits of logic which have changed, could that be explained? It's difficult to tell because of how GitHub has formatted the diff.

core/graph/src/main/java/uk/gov/gchq/gaffer/graph/Graph.java
tb06904: The original logic didn't really make much sense technically not much has really changed though. 

Before to get into this block at least one of those conditions must've been met but then you don't do anything unless the value is `null`. Take for example the `schema`, to get into this block it either needs to be set or one of the other values are set at which point you don't do anything with it anyways as the next check is to make sure it is `null`. Therefore, conditionally the outer block is basically making no difference as to what gets called internally so may as well remove it.

core/graph/src/test/java/uk/gov/gchq/gaffer/operation/export/resultcache/handler/ExportToGafferResultCacheHandlerTest.java
tb06904: This was something I couldn't quite get to the bottom of, originally this test was implemented to just check the `execute()` method was called at least once. Then under [this commit](https://github.com/gchq/Gaffer/commit/dd18afb4367c1700da21ebc9ca69f59c9c14967a) it was changed to exactly 3 with no explanation, with the current changes it now seems to run 4 times so changed to original implementation.

core/graph/src/main/java/uk/gov/gchq/gaffer/graph/GraphConfig.java
tb06904: Ahh this was something left from debugging the tests I'll apply to other one too

core/graph/src/test/java/uk/gov/gchq/gaffer/operation/export/resultcache/handler/ExportToGafferResultCacheHandlerTest.java
GCHQDeveloper314: Interesting, @t92549 - any ideas?

core/graph/src/main/java/uk/gov/gchq/gaffer/graph/Graph.java
GCHQDeveloper314: Ah ok, just wanted to be sure.

core/graph/src/test/java/uk/gov/gchq/gaffer/operation/export/resultcache/handler/ExportToGafferResultCacheHandlerTest.java
t92549: In that old PR I probably just made the test case more specific/precise, however, I am pretty sure there is no reason it has to be exactly 3, we just want to make sure it was called once, so I think this change is good.

core/graph/src/main/java/uk/gov/gchq/gaffer/graph/GraphConfig.java
```diff
@@ -116,25 +133,143 @@ public void setHooks(final List<GraphHook> hooks) {
...
+
+        // Get Handler for the operation to try extract the cache suffix if applicable class
+        final OperationHandler<Operation> addToCacheHandler = store.getOperationHandler(operationClass);
+        if (AddToCacheHandler.class.isAssignableFrom(addToCacheHandler.getClass())) {
+            // Extract the Suffix
+            suffix = ((AddToCacheHandler<?>) addToCacheHandler).getSuffixCacheName();
+        } else {
+            // Otherwise log warning and continue with the suffix from the properties
+            LOGGER.warn(
+                "Handler for: {} was not expected type: {} cant get suffixCache using value from property: {}",
```
t92549, line 210:
```suggestion
                "Handler for: {} was not expected type: {}. Can't get suffixCache using value from property: {}",
```

core/graph/src/main/java/uk/gov/gchq/gaffer/graph/GraphConfig.java
```diff
@@ -116,25 +133,143 @@ public void setHooks(final List<GraphHook> hooks) {
...
+                .filter(hook -> hookClass.isAssignableFrom(hook.getClass()))
+                .findAny()
+                .orElseThrow(() -> new GraphHookException(
+                        String.format("Unable to find matching hook in graph config for class %s", hookClass.getSimpleName())));
+
+            // Validate the suffix for a mismatch
+            final String nvrSuffix = nvrHook.getSuffixCacheName();
+            if (!suffix.equals(nvrSuffix)) {
+                //Error
+                throw new HookSuffixException(
```
t92549, line 244: Maybe this can be called `GraphHookSuffixException`?

core/graph/src/main/java/uk/gov/gchq/gaffer/graph/GraphConfig.java
```diff
@@ -116,25 +133,143 @@ public void setHooks(final List<GraphHook> hooks) {
...
+                // Provide info about the graph not having the required hook
+                LOGGER.info(
+                    "For GraphID: {} a handler was supplied for Operation: {}, but without a {}, adding {} with suffix: {}",
+                    getGraphId(),
+                    operationClass,
+                    hookClass.getSimpleName(),
+                    hookClass.getSimpleName(),
+                    suffix);
+                // Try add the hook
+                hooks.add(0, hookClass.getDeclaredConstructor(String.class).newInstance(suffix));
```
t92549, line 228: Strange this needs to get added at index 0, do any of the tests assert order?

t92549:
@tb06904 Really nice changes and refactoring! A lot easier to follow what is happening now and the logging is improved

core/graph/src/main/java/uk/gov/gchq/gaffer/graph/Graph.java
```diff
@@ -809,13 +816,19 @@ public Graph build() {
                 config.setLibrary(new NoGraphLibrary());
             }
 
-            if (null == config.getGraphId() && null != store) {
-                config.setGraphId(store.getGraphId());
+            // Make sure parent store properties are applied
+            if (parentStorePropertiesId != null) {
+                StoreProperties parentStoreProperties = config.getLibrary().getProperties(parentStorePropertiesId);
```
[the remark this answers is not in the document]
tb06904: Good point technically the var it scoped to the if statement anyways and only used once but guess there's no harm in adding it

t92549 (approved):

GCHQDev404 (dismissed):
So In the past I've lightly tried to sort out Graph.Builder() but it wasn't a good choice while also adding extra functionality so I revert it out I think twice. This seems like a good refactor. The logic which I care about seems to be retained and function equivelently. I just have a couple points I'd like you to answer.

core/graph/src/test/java/uk/gov/gchq/gaffer/graph/hook/UpdateGraphHookTest.java
```diff
@@ -0,0 +1,136 @@
...
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatExceptionOfType;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+class UpdateGraphHookTest {
+
+    @Test
+    void cantAddBadGraphHooks(@TempDir Path tmpDir) throws IOException {
```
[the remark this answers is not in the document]
t92549: Yeah I missed this, the tests all follow "shouldX". I guess in this case: `shouldNotBeAbleToAddBadGraphHooks`.

core/graph/src/main/java/uk/gov/gchq/gaffer/graph/Graph.java
```diff
@@ -777,16 +782,18 @@ public Builder addSchemas(final URI... schemaURI) {
...
         public Builder addSchema(final Path schemaPath) {
-            if (schemaPath != null) {
-                try (DirectoryStream<Path> stream = Files.isDirectory(schemaPath) ? Files.newDirectoryStream(schemaPath) : null) {
-                    if (stream != null) {
-                        stream.forEach(this::addSchema);
-                    } else {
-                        addSchema(Files.readAllBytes(schemaPath));
+            try {
+                // If directory load all schemas from it
+                if (Files.isDirectory(schemaPath)) {
```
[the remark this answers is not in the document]
tb06904: I'm not quite sure what you mean the stream is not created until the next line where its in a try with resources block? The `readAllBytes()`method handles the close automatically see [Javadocs](https://docs.oracle.com/javase/7/docs/api/java/nio/file/Files.html#readAllBytes(java.nio.file.Path))

core/graph/src/main/java/uk/gov/gchq/gaffer/graph/Graph.java
```diff
@@ -854,199 +864,156 @@ public Graph build() {
             return new Graph(config, store);
         }
 
-        private void updateGraphHooks(final GraphConfig config) {
-            updateNamedViewResolverHook(config);
+        private void validateAndUpdateHooks(final GraphConfig config) {
+            config.validateAndUpdateGetFromCacheHook(
```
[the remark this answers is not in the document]
tb06904: I would argue having multiple methods made it harder to follow as they just called another method or were just one if statement

core/graph/src/main/java/uk/gov/gchq/gaffer/graph/Graph.java
```diff
@@ -854,199 +864,156 @@ public Graph build() {
...
+                // If the merged result is still null init with the current schema else merge
+                if (mergedSchema == null) {
+                    mergedSchema = parentSchema;
                 } else {
-                    schema = new Schema.Builder()
-                            .merge(mergedParentSchema)
-                            .merge(schema)
-                            .build();
+                    mergedSchema = new Schema.Builder()
+                        .merge(mergedSchema)
```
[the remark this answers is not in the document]
tb06904: I might be wrong but I think the logic is different as if `addSchema()` was used then the order that the shemas are merged may be wrong e.g. currently the parent schemas will always be added first then any existing ones merged later. If I used `addSchema()` the parent ones could be merged after, this might not be an issue but I'm unsure.

core/graph/src/main/java/uk/gov/gchq/gaffer/graph/Graph.java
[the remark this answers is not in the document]
GCHQDeveloper314: I don't see any issue with that. While this refactor is better than before, I think it may be possible to improve further and not have a nested `try`.

core/graph/src/main/java/uk/gov/gchq/gaffer/graph/Graph.java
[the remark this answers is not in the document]
GCHQDeveloper314: It's unfortunate that these methods were only just added quite recently. I had thought there was some other reason for splitting this out - such as to be called from elsewhere - but if not, it seems better without them; though I'm not a big fan of splitting the arguments across lots of lines.

core/data/src/test/java/uk/gov/gchq/gaffer/data/util/ElementUtil.java
GCHQDeveloper314, line 1: These changes are not in scope for this issue. I presume the code removed is redundant. Preferably this would be done in a separate PR.

store-implementation/federated-store/src/test/java/uk/gov/gchq/gaffer/federatedstore/FederatedStoreCacheSuffixTest.java
```diff
@@ -38,158 +38,169 @@
...
+            .config(new GraphConfig.Builder()
+                    .graphId(GRAPH_ID_TEST_FEDERATED_STORE)
+                    .addHook(new NamedOperationResolver((String) null))
+                    .build())
+            .storeProperties(new FederatedStoreProperties());
+
+        // Ensure exception thrown with relevant info
+        assertThatExceptionOfType(GraphHookSuffixException.class)
+            .isThrownBy(() -> builder.build())
+            .withMessageContaining(NamedOperationResolver.class.getSimpleName())
```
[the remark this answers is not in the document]
GCHQDeveloper314: A case in point is the dependency uplift for Gaffer 2.0.0 - when doing this I found various error strings which had changed due to the dependency providing the string switching to a new format. For almost all of these I changed to using `withMessageContaining`. Though for a longer error like this there's more of an argument to have an exact match to ensure order, regex could also ensure order while allowing for some variation.

GCHQDeveloper314 (commented):
Changes to these files appear to be out of scope of the original issue:
- `GraphFactoryTest.java`
- `ElementUtil.java`
- `ExportToGafferResultCacheHandlerTest.java`
- `GetGafferResultCacheExportHandlerTest.java`

store-implementation/federated-store/src/test/java/uk/gov/gchq/gaffer/federatedstore/FederatedStoreCacheSuffixTest.java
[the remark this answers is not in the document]
GCHQDeveloper314: A case in point is the dependency uplift for Gaffer 2.0.0 - when doing this I found various error strings which had changed due to the dependency providing the string switching to a new format. For almost all of these I changed to using `withMessageContaining`. Though for a longer error like this there's more of an argument to have an exact match to ensure order, regex could also ensure order while allowing for some variation.

core/data/src/test/java/uk/gov/gchq/gaffer/data/util/ElementUtil.java
tb06904: These changes were made as these tests are using aspects of the changes in this PR:
`GraphFactoryTest.java` - Changes made as it used one of the methods affected by the refactor moved it to new asserts and the ordering of the graph hooks should not be important.
`ElementUtil.java` - Changes were made here to help with debugging as a test was failing here so the changes to use better assertions helped find the problem.
`ExportToGafferResultCacheHandlerTest.java` and `GetGafferResultCacheExportHandlerTest.java` - Already a comment about this as the assertion was being too specific.

core/graph/src/main/java/uk/gov/gchq/gaffer/graph/Graph.java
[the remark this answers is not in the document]
tb06904: The only way to not have the try block and still use a directory stream would be to have a finally block in the outer try, unsure which is more readable?

t92549 (approved):

GCHQDeveloper314 (approved):

PR #3051, opened 2023-10-27
PR requests with issues

### LINKED ISSUE #3045: MergeSchema is ignoring SchemaExceptions

GCHQDev404:
# Related issue

- Resolve #3045


core/data/src/main/java/uk/gov/gchq/gaffer/data/elementdefinition/exception/SchemaException.java
```diff
@@ -35,4 +36,19 @@ public SchemaException(final String message) {
     public SchemaException(final String message, final Throwable e) {
         super(message, e, BAD_REQUEST);
     }
+
+    public SchemaException preAppendToMessage(final String appendToMessage) {
+        if (null == this.appendToMessage) {
+            this.appendToMessage = new StringBuilder();
+        }
+        //preAppend
```
t92549, line 44: This is called prepend

core/store/src/main/java/uk/gov/gchq/gaffer/store/schema/Schema.java
```diff
@@ -350,8 +348,10 @@ public CHILD_CLASS merge(final Schema schema) {
 
                     mergeConfig(thatSchema);
 
+                } catch (final SchemaException e) {
+                    throw e.preAppendToMessage(ERROR_MERGING_SCHEMA_DUE_TO);
```
t92549, line 352: A little confused by this method when straight afterwards `throw new SchemaException(ERROR_MERGING_SCHEMA_DUE_TO + e.getMessage(), e);` is used

tb06904: This is used to add some additional information to any exceptions thrown that extend the `SchemaException` type without loosing the sub exception type which is caught and handled in the `MergeSchema` class.

Technically this is kind of the same as doing:

```java
throw e.getCause().getClass().getDeclaredConstructor().newInstance("something" + e.getMessage());`
```

t92549: Okay, this does look less verbose than what you showed, so makes sense. Maybe naming could be better though

tb06904: I agree, to be honest it might be simpler now to just catch the exact exceptions now there is only three to eliminate the need for custom internal prepend method e.g.
```java
catch (SplitElementGroupDefException e) {
    throw new SplitElementGroupDefException("extra info" + e.getMessage(), e);
} catch (VertexSerialiserSchemaException e) {
    throw new VertexSerialiserSchemaException("extra info" + e.getMessage(), e);
} catch (VisibilityPropertySchemaException e) {
    throw new VisibilityPropertySchemaException("extra info" + e.getMessage(), e);
}
```

core/data/src/main/java/uk/gov/gchq/gaffer/data/elementdefinition/exception/SchemaException.java
GCHQDev404: lol

core/store/src/main/java/uk/gov/gchq/gaffer/store/schema/Schema.java
t92549: But they add different information depending on the type

GCHQDev404: This line is adding the same information.

PR #3055, opened 2023-11-02
Gh-2891: Aggregate First and Last properties consistently in Accumulo

### LINKED ISSUE #2891: First properties are inconsistently aggregated in Accumulo
**Describe the bug**
Properties with the aggregationFunction of First seem to return inconsistent results when updated multiple times in an Accumulo store. 

**To Reproduce**
1. Have an AccumuloStore with a schema where a property is aggregated with First
2. AddElements and you should have the value returned by GetElements
3. Perform multiple AddElements and keep track of the order that you added them in
4. Perform GetElements and note that the resulting property will change inconsistently

**Expected behavior**
What should happen for First is that the most recent value is returned. 

**Additional context**
This leads to occasions where the First function provides inconsistent results, which should never happen. Other pure functions like Max are deterministic and will always return the max value immediately. I worry that due to how the scanning is working and applying iterators, First will produce non deterministic results and that sometimes data will be lost. 
The same is also maybe true of Last which should always retain the oldest value.

An experiment/test should be made in order to try and reliably recreate this issue

tb06904:
Following some investigation the bug appears to be due to the Accumulo timestamp.
 
Looking at the Accumulo docs it is stated that if you insert a key with same rowID, column and timestamp then the behaviour is non-deterministic, this is what was happening when aggregating as the timestamp is being hardcoded to a default value and rowID and column would be the same ([see](https://accumulo.apache.org/1.8/accumulo_user_manual.html#_versioning_iterators_and_timestamps))

The solution here will apply a timestamp to elements that are aggregated via time sensitive aggregation functions such as, First and Last. This should fix the original issue and ensure no unnecessary timestamps are added to data that doesn't need them.

Integration test has been added to ensure this is fixed.

# Related issue

- Resolve #2891

### DIFF
--- a/store-implementation/accumulo-store/src/main/java/uk/gov/gchq/gaffer/accumulostore/key/core/AbstractCoreKeyAccumuloElementConverter.java
+++ b/store-implementation/accumulo-store/src/main/java/uk/gov/gchq/gaffer/accumulostore/key/core/AbstractCoreKeyAccumuloElementConverter.java
@@ -353,7 +353,7 @@ public abstract class AbstractCoreKeyAccumuloElementConverter implements Accumul
 
     @Override
     public long buildTimestamp(final String group, final Properties properties) {
-        Long timestamp = null;
+        Long timestamp;
         if (timestampProperty != null) {
             timestamp = (Long) properties.get(timestampProperty);
             return timestamp;
@@ -375,7 +375,6 @@ public abstract class AbstractCoreKeyAccumuloElementConverter implements Accumul
         } else {
             timestamp = LongUtil.getTimeBasedRandom();
         }
-        
 
         return timestamp;
     }
--- a/store-implementation/accumulo-store/src/main/java/uk/gov/gchq/gaffer/accumulostore/key/core/AbstractCoreKeyAccumuloElementConverter.java
+++ b/store-implementation/accumulo-store/src/main/java/uk/gov/gchq/gaffer/accumulostore/key/core/AbstractCoreKeyAccumuloElementConverter.java
@@ -387,12 +387,11 @@ public abstract class AbstractCoreKeyAccumuloElementConverter implements Accumul
 
     @Override
     public long buildTimestamp(final String group, final Properties properties) {
-        Long timestamp = LongUtil.getTimeBasedRandom();
+        Long timestamp;
 
         // Allow override via property
         if (timestampProperty != null) {
-            timestamp = (Long) properties.get(timestampProperty);
-            return timestamp;
+            return (Long) properties.get(timestampProperty);
         }
 
         // Check if aggregating to see what timestamp we should apply
@@ -402,7 +401,10 @@ public abstract class AbstractCoreKeyAccumuloElementConverter implements Accumul
             if (timeSensitiveAggregatedGroups.contains(group)) {
                 timestamp = System.currentTimeMillis();
             }
+        } else {
+            timestamp = LongUtil.getTimeBasedRandom();
         }
+
         return timestamp;
     }
 
--- a/store-implementation/accumulo-store/src/main/java/uk/gov/gchq/gaffer/accumulostore/key/core/AbstractCoreKeyAccumuloElementConverter.java
+++ b/store-implementation/accumulo-store/src/main/java/uk/gov/gchq/gaffer/accumulostore/key/core/AbstractCoreKeyAccumuloElementConverter.java
@@ -50,6 +50,8 @@ import java.nio.charset.StandardCharsets;
 import java.util.Collections;
 import java.util.Iterator;
 import java.util.Set;
+import java.util.stream.Stream;
+import java.util.stream.StreamSupport;
 
 @SuppressWarnings("unchecked")
 public abstract class AbstractCoreKeyAccumuloElementConverter implements AccumuloElementConverter {
@@ -352,17 +354,28 @@ public abstract class AbstractCoreKeyAccumuloElementConverter implements Accumul
     @Override
     public long buildTimestamp(final String group, final Properties properties) {
         Long timestamp = null;
-        if (null != timestampProperty) {
+        if (timestampProperty != null) {
             timestamp = (Long) properties.get(timestampProperty);
+            return timestamp;
         }
 
-        if (null == timestamp) {
-            if (aggregatedGroups.contains(group)) {
-                timestamp = AccumuloStoreConstants.DEFAULT_TIMESTAMP;
+        // Check if aggregating to see what timestamp we should apply
+        if (aggregatedGroups.contains(group)) {
+            Stream<TypeDefinition> typeStream = StreamSupport.stream(
+                schema.getElement(group).getPropertyTypeDefs().spliterator(), true);
+
+            // If any types used by the element aggregate using a time sensitive function, add a timestamp
+            if (typeStream.anyMatch(td -> AccumuloStoreConstants.TIME_SENSITIVE_AGGREGATORS.contains(
+                    td.getAggregateFunction().getClass().getSimpleName()))) {
+                // Add timestamp
+                timestamp = System.currentTimeMillis();
             } else {
-                timestamp = LongUtil.getTimeBasedRandom();
+                timestamp = AccumuloStoreConstants.DEFAULT_TIMESTAMP;
             }
+        } else {
+            timestamp = LongUtil.getTimeBasedRandom();
         }
+        
 
         return timestamp;
     }
--- a/store-implementation/accumulo-store/src/main/java/uk/gov/gchq/gaffer/accumulostore/key/core/AbstractCoreKeyAccumuloElementConverter.java
+++ b/store-implementation/accumulo-store/src/main/java/uk/gov/gchq/gaffer/accumulostore/key/core/AbstractCoreKeyAccumuloElementConverter.java
@@ -47,7 +47,6 @@ import uk.gov.gchq.gaffer.store.schema.TypeDefinition;
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.nio.charset.StandardCharsets;
-import java.util.Arrays;
 import java.util.Collections;
 import java.util.Iterator;
 import java.util.Set;
--- a/store-implementation/accumulo-store/src/main/java/uk/gov/gchq/gaffer/accumulostore/key/core/AbstractCoreKeyAccumuloElementConverter.java
+++ b/store-implementation/accumulo-store/src/main/java/uk/gov/gchq/gaffer/accumulostore/key/core/AbstractCoreKeyAccumuloElementConverter.java
@@ -396,10 +396,11 @@ public abstract class AbstractCoreKeyAccumuloElementConverter implements Accumul
 
         // Check if aggregating to see what timestamp we should apply
         if (aggregatedGroups.contains(group)) {
-            timestamp = AccumuloStoreConstants.DEFAULT_TIMESTAMP;
             // If any types used by the element aggregate using a time sensitive function, add a timestamp
             if (timeSensitiveAggregatedGroups.contains(group)) {
                 timestamp = System.currentTimeMillis();
+            } else {
+                timestamp = AccumuloStoreConstants.DEFAULT_TIMESTAMP;
             }
         } else {
             timestamp = LongUtil.getTimeBasedRandom();
--- a/store-implementation/accumulo-store/src/main/java/uk/gov/gchq/gaffer/accumulostore/key/core/AbstractCoreKeyAccumuloElementConverter.java
+++ b/store-implementation/accumulo-store/src/main/java/uk/gov/gchq/gaffer/accumulostore/key/core/AbstractCoreKeyAccumuloElementConverter.java
@@ -366,7 +366,7 @@ public abstract class AbstractCoreKeyAccumuloElementConverter implements Accumul
                 schema.getElement(group).getPropertyTypeDefs().spliterator(), true);
 
             // If any types used by the element aggregate using a time sensitive function, add a timestamp
-            if (typeStream.anyMatch(td -> Arrays.asList(AccumuloStoreConstants.TIME_SENSITIVE_AGGREGATORS)
+            if (typeStream.anyMatch(td -> AccumuloStoreConstants.TIME_SENSITIVE_AGGREGATORS
                     .contains(td.getAggregateFunction().getClass().getSimpleName()))) {
                 // Add timestamp
                 timestamp = System.currentTimeMillis();
--- a/store-implementation/accumulo-store/src/main/java/uk/gov/gchq/gaffer/accumulostore/key/core/AbstractCoreKeyAccumuloElementConverter.java
+++ b/store-implementation/accumulo-store/src/main/java/uk/gov/gchq/gaffer/accumulostore/key/core/AbstractCoreKeyAccumuloElementConverter.java
@@ -280,15 +280,12 @@ public abstract class AbstractCoreKeyAccumuloElementConverter implements Accumul
         for (final String group : aggregatedGroups) {
             // Check all type definitions for the group
             Stream<TypeDefinition> typeStream = StreamSupport.stream(
-                schema.getElement(group).getPropertyTypeDefs().spliterator(), true);
+                schema.getElement(group).getPropertyTypeDefs().spliterator(), false);
+
             // Check if a time sensitive function is used anywhere
-            boolean timeSensitiveGroup = typeStream.anyMatch(td -> {
-                if (td.getAggregateFunction() == null) {
-                    return false;
-                }
-                return AccumuloStoreConstants.TIME_SENSITIVE_AGGREGATORS.contains(
-                    td.getAggregateFunction().getClass().getSimpleName());
-            });
+            boolean timeSensitiveGroup = typeStream.anyMatch(td ->
+                td.getAggregateFunction() != null &&
+                AccumuloStoreConstants.TIME_SENSITIVE_AGGREGATORS.contains(td.getAggregateFunction().getClass().getName()));
 
             if (timeSensitiveGroup) {
                 timeSensitiveAggregatedGroups.add(group);
--- a/store-implementation/accumulo-store/src/main/java/uk/gov/gchq/gaffer/accumulostore/key/core/AbstractCoreKeyAccumuloElementConverter.java
+++ b/store-implementation/accumulo-store/src/main/java/uk/gov/gchq/gaffer/accumulostore/key/core/AbstractCoreKeyAccumuloElementConverter.java
@@ -47,6 +47,7 @@ import uk.gov.gchq.gaffer.store.schema.TypeDefinition;
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.nio.charset.StandardCharsets;
+import java.util.Arrays;
 import java.util.Collections;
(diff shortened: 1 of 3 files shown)


store-implementation/accumulo-store/src/main/java/uk/gov/gchq/gaffer/accumulostore/key/core/AbstractCoreKeyAccumuloElementConverter.java
```diff
@@ -351,17 +353,27 @@ protected boolean isColumnQualifierBytesValid(final byte[] bytes, final int numP
...
+            return timestamp;
         }
 
-        if (null == timestamp) {
-            if (aggregatedGroups.contains(group)) {
-                timestamp = AccumuloStoreConstants.DEFAULT_TIMESTAMP;
+        // Check if aggregating to see what timestamp we should apply
+        if (aggregatedGroups.contains(group)) {
+            Stream<TypeDefinition> typeStream = StreamSupport.stream(
+                schema.getElement(group).getPropertyTypeDefs().spliterator(), true);
```
t92549, line 365: Perhaps these checks can be done once when the ElementConverter is created and the results cached in a Map? I just fear that running these checks on every AddElement would be inefficient.
I suppose this would only work if the ElementConverter is created once and reused, which I am assuming it is but uncertain.

tb06904: That was my concern too, yeah I wasn't sure if the ElementConverter was created then reused, I guess if it isn't then it probably should be anyway so caching a list of all the groups that use this type of aggregator should cut some of the logic out and be more efficient.

t92549: Having a little dig around, it seems the ElementConverter is created in a bunch of `init` functions and reused:
https://github.com/gchq/Gaffer/blob/dde023fb2f4a2c585612f3a382f4f76b1933aada/store-implementation/accumulo-store/src/main/java/uk/gov/gchq/gaffer/accumulostore/key/core/impl/CoreKeyGroupByCombiner.java#L365-L373
Would also make sense as to why `timestampProperty` and `aggregatedGroups` are cached:
https://github.com/gchq/Gaffer/blob/dde023fb2f4a2c585612f3a382f4f76b1933aada/store-implementation/accumulo-store/src/main/java/uk/gov/gchq/gaffer/accumulostore/key/core/AbstractCoreKeyAccumuloElementConverter.java#L60-L64

tb06904: Good to know I'll move the logic to the constructor then

store-implementation/accumulo-store/src/main/java/uk/gov/gchq/gaffer/accumulostore/key/core/AbstractCoreKeyAccumuloElementConverter.java
```diff
@@ -353,29 +387,22 @@ protected boolean isColumnQualifierBytesValid(final byte[] bytes, final int numP
 
     @Override
     public long buildTimestamp(final String group, final Properties properties) {
-        Long timestamp;
+        Long timestamp = LongUtil.getTimeBasedRandom();
```
t92549, line 390: I think the ordering can be changed so this doesn't happen first, given it will be a tiny bit expensive. Currently, `timeSensitiveAggregatedGroups` will have to call `System.currentTimeMillis()` twice.

store-implementation/accumulo-store/src/main/java/uk/gov/gchq/gaffer/accumulostore/key/core/AbstractCoreKeyAccumuloElementConverter.java
```diff
@@ -58,11 +58,20 @@ public abstract class AbstractCoreKeyAccumuloElementConverter implements Accumul
     protected final Schema schema;
     private final String timestampProperty;
     private final Set<String> aggregatedGroups;
+    private final Set<String> timeSensitiveAggregatedGroups = new HashSet<>();
 
-    public AbstractCoreKeyAccumuloElementConverter(final Schema schema) {
```
t92549, line 60: Although this seems sensible, I would maybe leave it public as there might be users using this class directly.

tb06904: The class is abstract users cant instantiate it directly anyways so it would be a code smell to leave it public.

store-implementation/accumulo-store/src/main/java/uk/gov/gchq/gaffer/accumulostore/key/core/AbstractCoreKeyAccumuloElementConverter.java
tb06904: True good point.

store-implementation/accumulo-store/src/main/java/uk/gov/gchq/gaffer/accumulostore/key/core/AbstractCoreKeyAccumuloElementConverter.java
t92549: Oops good point didn't catch that...

store-implementation/accumulo-store/src/main/java/uk/gov/gchq/gaffer/accumulostore/key/core/AbstractCoreKeyAccumuloElementConverter.java
```diff
@@ -351,17 +387,22 @@ protected boolean isColumnQualifierBytesValid(final byte[] bytes, final int numP
...
         }
 
-        if (null == timestamp) {
-            if (aggregatedGroups.contains(group)) {
-                timestamp = AccumuloStoreConstants.DEFAULT_TIMESTAMP;
-            } else {
-                timestamp = LongUtil.getTimeBasedRandom();
+        // Check if aggregating to see what timestamp we should apply
+        if (aggregatedGroups.contains(group)) {
+            timestamp = AccumuloStoreConstants.DEFAULT_TIMESTAMP;
```
t92549, line 399: Perhaps this could be in an else after the `if (timeSensitiveAggregatedGroups.contains(group)) {` just so it isn't set twice. I promise I am not usually so nitpicky! Just really conscious about performance / making the logic flow as small as possible as this code will be run a lot.

tb06904: I think counter intuitively it can be more expensive to have more branches due to CPU branch prediction, if this assignment was running a method it would be bit more clear cut but it is a static variable which means this is very cheap to do. I think it will depend more on what hardware its ran on but ultimately the time complexity is the same either way so differences between would be negligible.

t92549: My thinking was branch prediction wouldn't come into this as the if happens no matter what, and the else doesn't add anything on top of that. Might be completely wrong though I am no expert on this.

t92549 (approved):

store-implementation/accumulo-store/src/test/java/uk/gov/gchq/gaffer/accumulostore/integration/FirstLastIT.java
```diff
@@ -0,0 +1,199 @@
...
+import static org.assertj.core.api.Assertions.assertThat;
+import static uk.gov.gchq.gaffer.commonutil.TestGroups.ENTITY;
+import static uk.gov.gchq.gaffer.commonutil.TestPropertyNames.PROP_1;
+import static uk.gov.gchq.gaffer.commonutil.TestPropertyNames.PROP_2;
+import static uk.gov.gchq.gaffer.store.TestTypes.ID_STRING;
+import static uk.gov.gchq.gaffer.store.TestTypes.PROP_INTEGER;
+import static uk.gov.gchq.gaffer.store.TestTypes.PROP_INTEGER_2;
+import static uk.gov.gchq.gaffer.store.TestTypes.STRING_TYPE;
+
+class FirstLastIT extends StandaloneIT {
```
t92549, line 59: Thinking about this again, perhaps this should go into core Gaffer ITs rather than a standalone AccumuloIT

t92549 (changes requested):

store-implementation/accumulo-store/src/test/java/uk/gov/gchq/gaffer/accumulostore/integration/FirstLastIT.java
t92549: Due to how this interacts specifically with Accumulo compactions, I don't see any reason this should affect any other store.

t92549 (dismissed):

l46978 (approved):

store-implementation/accumulo-store/src/main/java/uk/gov/gchq/gaffer/accumulostore/key/core/AbstractCoreKeyAccumuloElementConverter.java
```diff
@@ -260,6 +271,31 @@ public byte[] buildColumnQualifier(final String group, final Properties properti
...
+    /**
+     * Populates the set of groups that aggregate with a time sensitive
+     * aggregation function.
+     */
+    private void populateTimeSensitiveGroups() {
+        // Find the groups that use a time sensitive aggregation function
+        for (final String group : aggregatedGroups) {
+            // Check all type definitions for the group
+            Stream<TypeDefinition> typeStream = StreamSupport.stream(
+                schema.getElement(group).getPropertyTypeDefs().spliterator(), true);
```
GCHQDev404, line 283: > true

groups aren't that big, is there any cost to setting up parellel ?

store-implementation/accumulo-store/src/main/java/uk/gov/gchq/gaffer/accumulostore/key/core/AbstractCoreKeyAccumuloElementConverter.java
```diff
@@ -260,6 +271,31 @@ public byte[] buildColumnQualifier(final String group, final Properties properti
...
+            // Check if a time sensitive function is used anywhere
+            boolean timeSensitiveGroup = typeStream.anyMatch(td -> {
+                if (td.getAggregateFunction() == null) {
+                    return false;
+                }
+                return AccumuloStoreConstants.TIME_SENSITIVE_AGGREGATORS.contains(
+                    td.getAggregateFunction().getClass().getSimpleName());
+            });
```
GCHQDev404, line 291: Single return condition. please

```suggestion
            boolean timeSensitiveGroup = typeStream.anyMatch(td ->
                    td.getAggregateFunction() != null
                            && AccumuloStoreConstants.TIME_SENSITIVE_AGGREGATORS
                            .contains(td.getAggregateFunction().getClass().getSimpleName()));
```

store-implementation/accumulo-store/src/main/java/uk/gov/gchq/gaffer/accumulostore/utils/AccumuloStoreConstants.java
```diff
@@ -45,6 +52,12 @@ public final class AccumuloStoreConstants {
      */
     public static final long DEFAULT_TIMESTAMP = 1L;
 
+    // Time sensitive aggregation functions where an actual timestamp must applied to the key
+    public static final List<String> TIME_SENSITIVE_AGGREGATORS = Collections.unmodifiableList(
+        Arrays.asList(
+            First.class.getSimpleName(),
```
GCHQDev404, line 58: I don't think you want to use simple name.

store-implementation/accumulo-store/src/main/java/uk/gov/gchq/gaffer/accumulostore/key/core/AbstractCoreKeyAccumuloElementConverter.java
GCHQDev404: Also I don't think you want to use simple name.

GCHQDev404 (dismissed):
1 question
1 request
1 suggestion

store-implementation/accumulo-store/src/main/java/uk/gov/gchq/gaffer/accumulostore/key/core/AbstractCoreKeyAccumuloElementConverter.java
tb06904: It should be fairly cheap to split the data but if there's not going to be that many as you say then probably not worth it

store-implementation/accumulo-store/src/main/java/uk/gov/gchq/gaffer/accumulostore/key/core/AbstractCoreKeyAccumuloElementConverter.java
tb06904: Using simple name as allows for easy comparison rather than messing around with instances of classes etc. I could use the full name if needed though.

store-implementation/accumulo-store/src/main/java/uk/gov/gchq/gaffer/accumulostore/key/core/AbstractCoreKeyAccumuloElementConverter.java
GCHQDev404: groups you'll have 10's maybe a 100.

store-implementation/accumulo-store/src/main/java/uk/gov/gchq/gaffer/accumulostore/key/core/AbstractCoreKeyAccumuloElementConverter.java
GCHQDev404: Else where in Gaffer we use the full names.

t92549 (dismissed):

GCHQDev404 (approved):

GCHQDeveloper314 (changes requested):

store-implementation/accumulo-store/src/test/java/uk/gov/gchq/gaffer/accumulostore/integration/FirstLastIT.java
```diff
@@ -0,0 +1,199 @@
...
+                Arguments.of(new GetAllElements()),
+                Arguments.of(new GetElements.Builder().input(new EntitySeed(VERTEX)).build()));
+    }
+
+    @ParameterizedTest
+    @MethodSource("getOperations")
+    void shouldReturnCorrectResultsAfterCompaction(final Output<Iterable<? extends Element>> getOperation) throws OperationException, InterruptedException, StoreException, TableNotFoundException, AccumuloException, AccumuloSecurityException {
+        final Graph graph = createGraph();
+        final AccumuloStore accumuloStore = new AccumuloStore();
+        accumuloStore.initialise(GRAPH_ID, createSchema(), createStoreProperties());
```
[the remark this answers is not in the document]
tb06904: One for @t92549 but I assume its to be consistent with the other tests as `createStoreProperties()` is an override for the `StandAloneIT`

store-implementation/accumulo-store/src/main/java/uk/gov/gchq/gaffer/accumulostore/key/core/AbstractCoreKeyAccumuloElementConverter.java
```diff
@@ -351,17 +384,23 @@ protected boolean isColumnQualifierBytesValid(final byte[] bytes, final int numP
...
     @Override
     public long buildTimestamp(final String group, final Properties properties) {
-        Long timestamp = null;
-        if (null != timestampProperty) {
-            timestamp = (Long) properties.get(timestampProperty);
+        Long timestamp;
+
+        // Allow override via property
+        if (timestampProperty != null) {
+            return (Long) properties.get(timestampProperty);
```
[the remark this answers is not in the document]
tb06904: I didn't really see much point in assigning to the variable as this way it will be left uninitialised too saving memory, a variable doesn't have to be used on all possible branches, I'd argue this is just as clear to follow.

store-implementation/accumulo-store/src/test/java/uk/gov/gchq/gaffer/accumulostore/integration/FirstLastIT.java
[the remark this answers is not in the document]
GCHQDeveloper314: If it's an override then that's fine.

store-implementation/accumulo-store/src/main/java/uk/gov/gchq/gaffer/accumulostore/utils/AccumuloStoreConstants.java
```diff
@@ -45,6 +52,12 @@ public final class AccumuloStoreConstants {
      */
     public static final long DEFAULT_TIMESTAMP = 1L;
 
+    // Time sensitive aggregation functions where an actual timestamp must applied to the key
```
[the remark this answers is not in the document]
tb06904: Switched to class comparison but was reluctant to enforce sub type just in case but you are right in practise usually always going to be binary operators.

GCHQDeveloper314 (commented):
Minor comment on generics, otherwise ready

store-implementation/accumulo-store/src/main/java/uk/gov/gchq/gaffer/accumulostore/utils/AccumuloStoreConstants.java
```diff
@@ -45,6 +52,12 @@ public final class AccumuloStoreConstants {
      */
     public static final long DEFAULT_TIMESTAMP = 1L;
 
+    /**
+     * Time sensitive aggregation functions where an actual timestamp must be applied to the key.
+     */
+    public static final List<Class<?>> TIME_SENSITIVE_AGGREGATORS = Collections.unmodifiableList(
```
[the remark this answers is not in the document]
tb06904: I think there is a wider issue with the generics as, as you said `getAggregateFunction()` is returning a raw type anyway so any enforcement here would require a cast on that side. Probably should be fixed under tech debt that fixes generics.

GCHQDeveloper314 (approved):

store-implementation/accumulo-store/src/main/java/uk/gov/gchq/gaffer/accumulostore/utils/AccumuloStoreConstants.java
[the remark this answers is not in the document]
t92549: See https://github.com/gchq/Gaffer/issues/3074

t92549 (approved):

PR #3060, opened 2023-11-03
Gh-3059: ApplyViewToElementsFunction apply Schema validation

### LINKED ISSUE #3059: ApplyViewToElementsFunction does not apply Schema validation
**Describe the bug**
ApplyViewToElementsFunction will apply the View of an operation to the results from multiple subgraphs. This is meant to ensure that the results are correctly aggregated and validated after a query to a FederatedStore.

However, despite the View validation being applied, **the Schema validation is not applied**. This could be intended, but I suspect it is not, as the intention of this merge function is to ensure all aggregation and validation happens as if the results came from one graph. If this is indeed unintended, I would suggest the merge function should be renamed to something like ValidateAndAggregateElements.

**Additional context**
The reason this happens is that the results are currently merged inside a MapStore and the MapStore does not currently support schema validation: https://github.com/gchq/Gaffer/issues/2613.

Some potential fixes are:
- Add MapStore validation as suggested in https://github.com/gchq/Gaffer/issues/2613.
- Change the ApplyViewToElementsFunction to create an ephemeral AccumuloStore rather than MapStore

GCHQDev404:
# Related issue

- Resolve #3059


store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/util/ApplyViewToElementsFunction.java
```diff
@@ -120,6 +145,55 @@ private static void validate(final Map<String, Object> context) {
...
+    private static void updateEdgesViewFilterWithSchemaValidation(final Schema schema, final View view, final View.Builder updatedView) {
+        final Map<String, SchemaEdgeDefinition> elements = schema.getEdges();
+        for (final Map.Entry<String, ? extends SchemaElementDefinition> schemaElement : elements.entrySet()) {
+
+            final String elementKey = schemaElement.getKey();
+            final ViewElementDefinition updatedViewElementDef = getUpdatedViewElementDef(elementKey, schemaElement.getValue(), view);
+
+            updatedView.edge(elementKey, updatedViewElementDef);
+        }
+    }
```
GCHQDev404, line 168: If anyone let me know how I can squash these two into 1 method I'd love that.
`schema.getEdges();` and ` updatedView.edge(elementKey, updatedViewElementDef)`

The only way I can thing is return crappy map.

store-implementation/federated-store/src/test/java/uk/gov/gchq/gaffer/federatedstore/FederatedStoreViewAggregationTest.java
```diff
@@ -132,6 +135,37 @@ public void shouldOnlyReturn1EntitySmallerThanViewFilter() throws Exception {
                 .hasSize(1);
     }
 
+    @Test
+    public void shouldOnlyReturn1EntitySmallerThanSchemaValidationLimit() throws Exception {
```
t92549, line 139: I think this test doesn't make sense in the FederatedStoreViewAggregationTest, maybe should be moved or the test renamed.

t92549: Ideally, there would be a whole suite of tests that check how different things are merged when schemas are slightly different. These exists elsewhere, but anything to do with ApplyViewToElementsFunction should probably be in one place.

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/util/ApplyViewToElementsFunction.java
```diff
@@ -77,13 +84,31 @@ public ApplyViewToElementsFunction(final Map<String, Object> context) throws Gaf
...
+            updateViewWithValidationFromSchema(context);
+
             this.context = Collections.unmodifiableMap(context);
         } catch (final Exception e) {
             throw new GafferCheckedException("Unable to create TemporaryResultsGraph", e);
         }
 
     }
 
+    private static void updateViewWithValidationFromSchema(final Map<String, Object> context) {
```
t92549, line 97: Should the function `ApplyViewToElementsFunction` be renamed?

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/util/ApplyViewToElementsFunction.java
```diff
@@ -77,13 +84,31 @@ public ApplyViewToElementsFunction(final Map<String, Object> context) throws Gaf
...
+        //Only do this for MapStore, not required for other stores.
+        if (MapStore.class.getName().equals(((Graph) context.get(TEMP_RESULTS_GRAPH)).getStoreProperties().getStoreClass())) {
```
t92549, line 99: This seems like a lot of manual tweaking of the View. I think adding validation to the MapStore would be a cleaner solution. Is there a reason to this over that?

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/util/ApplyViewToElementsFunction.java
tb06904: As a thought this might be a good place to use inline lambdas instead you could replace both these methods with just:

```java
schema.getEntities().entrySet().forEach(entity -> {
    if (entity.getValue().hasValidation()) {
        updatedView.entity(entity.getKey(), getUpdatedViewElementDef(entity.getValue(), view, entity.getKey()));
    }
});

schema.getEdges().entrySet().forEach(edge -> {
    if (edge.getValue().hasValidation()) {
        updatedView.edge(edge.getKey(), getUpdatedViewElementDef(edge.getValue(), view, edge.getKey()));
    }
});
```

GCHQDev404: Thank you. I tried lambdas but I couldn't get it play nicely on Friday. but I was trying to pass the a lambda function in to execute within the loop. Completely forgot I could have just stream a foreach and not bothered with the for loop anymore. Silly me. a tired Friday I guess. I just knew it could be tidier. I went with returning a crappy map which you probably have seen. I'll whack this in.

store-implementation/federated-store/src/test/java/uk/gov/gchq/gaffer/federatedstore/FederatedStoreViewAggregationTest.java
GCHQDev404: "doesn't make sense in the FederatedStoreViewAggregationTest"

Can you explain?

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/util/ApplyViewToElementsFunction.java
GCHQDev404: To what?

It still applies view to elements. This additional new Validation suff is because of MapStore failings, this would be done for free otherwise. It should get ripped out when Mapstore is fixed.

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/util/ApplyViewToElementsFunction.java
GCHQDev404: Yes, Mapstore should be fixed and this code should be ripped out.

store-implementation/federated-store/src/test/java/uk/gov/gchq/gaffer/federatedstore/FederatedStoreViewAggregationTest.java
t92549: The test is called FederatedStoreViewAggregationTest but shouldOnlyReturn1EntitySmallerThanSchemaValidationLimit tests that the ApplyViewToElementsFunction applies schema validation, which has nothing to do with FederatedStoreViewAggregation.

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/util/ApplyViewToElementsFunction.java
t92549: It doesn't just apply the view anymore, it also reapplies the schema now. On this topic, does your fix apply schema aggregation too? Or just schema validation?

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/util/ApplyViewToElementsFunction.java
t92549: In that case, could it be written in such a way that it is obviously temporary? Maybe within one block that can just be deleted when the time comes?

store-implementation/federated-store/src/test/java/uk/gov/gchq/gaffer/federatedstore/FederatedStoreViewAggregationTest.java
GCHQDev404: > Ideally, there would be a whole suite of tests that check how different things are merged when schemas are slightly different. These exists elsewhere, but anything to do with ApplyViewToElementsFunction should probably be in one place.


asking for tests is always valid for all PR 😆 
You already wrote the tests that addGraphs and getSchema for slight different schema test? a ticket can be opened to pop some elements in those test? or modify them or expand them. `FederatedStoreSchemaTest.java`

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/util/ApplyViewToElementsFunction.java
t92549: "does your fix apply schema aggregation too?" - I think this is not necessary as the MapStore does this.

store-implementation/federated-store/src/test/java/uk/gov/gchq/gaffer/federatedstore/FederatedStoreViewAggregationTest.java
GCHQDev404: yes I agree 3058 should resolve that.

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/util/ApplyViewToElementsFunction.java
GCHQDev404: yep correct, mapstore does it. This is all due to Mapstore having issues that need completing.

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/util/ApplyViewToElementsFunction.java
GCHQDev404: It's all One method `updateViewWithValidationFromSchema()` the with sub methods that exist is for general Java readability and such. They will become highlighted and unused when you delete `updateViewWithValidationFromSchema`

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/util/ApplyViewToElementsFunction.java
GCHQDev404: I must have been super tired coz I didn't push my map changes on Friday either.

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/util/ApplyViewToElementsFunction.java
t92549: Perhaps this could actually be a different function called `ApplyViewAndSchemaToElementsFunction` and be an alternative?

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/util/MergeElementFunction.java
GCHQDeveloper314, line 1: The rename of this class is breaking.

GCHQDeveloper314 (changes requested):

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/util/MergeElementFunction.java
GCHQDev404: a breaking change?
I think it's low risk, it was the default merge, and would be overwritten to something else. unlikely someone would specify the default function.

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/util/MergeElementFunction.java
```diff
@@ -152,4 +220,12 @@ public Iterable<Object> apply(final Object update, final Iterable<Object> state)
             throw new GafferRuntimeException("Error getting all elements from temporary graph, due to:" + e.getMessage(), e);
         }
     }
+
+    private static Graph getGraph(final Map<String, Object> context) {
```
rj77259, line 224: Do we need static on this method considering it is private?

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/util/MergeElementFunction.java
```diff
@@ -152,4 +220,12 @@ public Iterable<Object> apply(final Object update, final Iterable<Object> state)
...
+        final Object g = context.get(TEMP_RESULTS_GRAPH);
```
rj77259, line 225: What is 'g'?

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/util/MergeElementFunction.java
```diff
@@ -56,16 +70,16 @@ public class ApplyViewToElementsFunction implements ContextSpecificMergeFunction
     @JsonProperty("context")
     private Map<String, Object> context;
 
-    public ApplyViewToElementsFunction() {
+    public MergeElementFunction() {
     }
 
-    public ApplyViewToElementsFunction(final Map<String, Object> context) throws GafferCheckedException {
+    public MergeElementFunction(final Map<String, Object> context) throws GafferCheckedException {
```
rj77259, line 76: Can this be refactored? The method is doing more than one thing. Can we have separate methods for each step of the process?

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/util/MergeElementFunction.java
```diff
@@ -77,16 +91,37 @@ public ApplyViewToElementsFunction(final Map<String, Object> context) throws Gaf
             }
             // Validate the supplied context before using
             validate(context);
+
+            updateViewWithValidationFromSchema(context);
+
             this.context = Collections.unmodifiableMap(context);
         } catch (final Exception e) {
             throw new GafferCheckedException("Unable to create TemporaryResultsGraph", e);
```
rj77259, line 99: Do we need to throw an exception here? Can we not just log the issue?

rj77259 (changes requested):

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/util/ApplyViewToElementsFunction.java
GCHQDev404: I think we landed on "MergeElementFunction"

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/util/MergeElementFunction.java
```diff
@@ -77,16 +91,37 @@ public ApplyViewToElementsFunction(final Map<String, Object> context) throws Gaf
...
         }
 
     }
 
+    private static void updateViewWithValidationFromSchema(final Map<String, Object> context) {
```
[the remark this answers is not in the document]
GCHQDev404: The method is not used by other classes so it's private. and the method is not tied to the instance of this class so it can be static.

It only changes the values of a context object. After that object has been modified it is then assigned to a persistent field.

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/util/MergeElementFunction.java
GCHQDev404: This is only used by this class so it is private.
It is not tied to the instance of the class, it a utility method for fetching graphs out of context which has been configured by this class. so it can be static.

Is there a reason for this to not be private static?

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/util/MergeElementFunction.java
GCHQDev404: Yes, the constructor is doing far too much here. But this code pre-dated this PR and branch.
I will be happy to for us to go through a refactor session together over this.

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/util/MergeElementFunction.java
GCHQDev404: If the result graph is not able to be created for merging, then the currently execution cannot continue.
Except get thrown and more info is added further up the stack.
There are places higher in the Stack that will log this error, but with more information.

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/util/MergeElementFunction.java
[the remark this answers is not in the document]
GCHQDev404: please examine https://github.com/gchq/Gaffer/pull/3165

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/util/MergeElementFunction.java
GCHQDev404: https://github.com/gchq/Gaffer/pull/3165

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/util/MergeElementFunction.java
GCHQDev404: please examine https://github.com/gchq/Gaffer/pull/3165

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/util/MergeElementFunction.java
GCHQDeveloper314: I don't think this matter so much anymore, so this is ok now

store-implementation/federated-store/src/test/java/uk/gov/gchq/gaffer/federatedstore/MergeElementFunctionAggregationViewFilterTest.java
```diff
@@ -110,17 +113,14 @@ public void shouldOnlyReturn1EntitySmallerThanViewFilter() throws Exception {
         //then
         assertThat(elementsWithPropertyLessThan2)
                 .isNotNull()
-                .withFailMessage("should return entity with property 1 which is less than view filter 2")
-                .contains(entity1)
                 .withFailMessage("should not return entity with property 99 which is more than view filter 2")
                 .doesNotContain(entity99)
                 .withFailMessage("should contain only 1")
-                .hasSize(1);
```
GCHQDeveloper314, line 118: The fail message should also have been removed.

store-implementation/federated-store/src/test/java/uk/gov/gchq/gaffer/federatedstore/MergeElementFunctionAggregationViewFilterTest.java
```diff
@@ -110,17 +113,14 @@ public void shouldOnlyReturn1EntitySmallerThanViewFilter() throws Exception {
         //then
         assertThat(elementsWithPropertyLessThan2)
                 .isNotNull()
-                .withFailMessage("should return entity with property 1 which is less than view filter 2")
-                .contains(entity1)
                 .withFailMessage("should not return entity with property 99 which is more than view filter 2")
                 .doesNotContain(entity99)
                 .withFailMessage("should contain only 1")
```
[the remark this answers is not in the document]
GCHQDev404: done

store-implementation/federated-store/src/test/java/uk/gov/gchq/gaffer/federatedstore/MergeElementFunctionAggregationViewFilterTest.java
GCHQDev404: done

GCHQDeveloper314:
This is a breaking change and is not a current priority. It can be revisited later.

PR #3062, opened 2023-11-03
Gh-3061: Ensure ToHllSketch functions copy

### LINKED ISSUE #3061: ToHllSketch functions do not copy from other HllSketch
**Describe the bug**
`ToHllSketch` and `IterableToHllSketch` can both be intitialised with another HllSketch. It should be copied but is not.
Additionally, these fields should be JsonIgnored as they won't be used over the rest api and wouldn't work anyway.

t92549:
# Related issue

- Resolve #3061

### DIFF
--- a/library/sketches-library/src/main/java/uk/gov/gchq/gaffer/sketches/datasketches/cardinality/function/IterableToHllSketch.java
+++ b/library/sketches-library/src/main/java/uk/gov/gchq/gaffer/sketches/datasketches/cardinality/function/IterableToHllSketch.java
@@ -37,27 +37,26 @@ import static uk.gov.gchq.gaffer.sketches.datasketches.cardinality.serialisation
 public class IterableToHllSketch extends KorypheFunction<Iterable<Object>, HllSketch> {
     private int logK = DEFAULT_LOG_K;
     @JsonIgnore
-    private HllSketch initHllSketch;
+    private HllSketch hllSketch;
 
     public IterableToHllSketch() {
+        if (hllSketch == null) {
+            hllSketch = new HllSketch(logK);
+        }
     }
 
     public IterableToHllSketch(final int logK) {
-        this.logK = logK;
+        setLogK(logK);
+        this();
     }
 
-    public IterableToHllSketch(final HllSketch initHllSketch) {
-        this.initHllSketch = initHllSketch;
+    public IterableToHllSketch(final HllSketch hllSketch) {
+        setHllSketch(hllSketch);
+        this();
     }
 
     @Override
     public HllSketch apply(final Iterable<Object> iterable) {
-        HllSketch hllSketch;
-        if (initHllSketch == null) {
-            hllSketch = new HllSketch(logK);
-        } else {
-            hllSketch = initHllSketch.copy();
-        }
         if (nonNull(iterable)) {
             for (final Object o : iterable) {
                 if (nonNull(o)) {
@@ -94,11 +93,13 @@ public class IterableToHllSketch extends KorypheFunction<Iterable<Object>, HllSk
         this.logK = logK;
     }
 
-    public HllSketch getInitHllSketch() {
-        return initHllSketch;
+    public HllSketch getHllSketch() {
+        return hllSketch;
     }
 
-    public void setInitHllSketch(final HllSketch initHllSketch) {
-        this.initHllSketch = initHllSketch;
+    public void setHllSketch(final HllSketch hllSketch) {
+        if (hllSketch != null) {
+            this.hllSketch = hllSketch.copy();
+        }
     }
 }
--- a/library/sketches-library/src/main/java/uk/gov/gchq/gaffer/sketches/datasketches/cardinality/function/IterableToHllSketch.java
+++ b/library/sketches-library/src/main/java/uk/gov/gchq/gaffer/sketches/datasketches/cardinality/function/IterableToHllSketch.java
@@ -37,7 +37,7 @@ import static uk.gov.gchq.gaffer.sketches.datasketches.cardinality.serialisation
 public class IterableToHllSketch extends KorypheFunction<Iterable<Object>, HllSketch> {
     private int logK = DEFAULT_LOG_K;
     @JsonIgnore
-    private HllSketch initHllSketch;
+    private HllSketch hllSketch;
 
     public IterableToHllSketch() {
     }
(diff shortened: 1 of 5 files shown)


library/sketches-library/src/main/java/uk/gov/gchq/gaffer/sketches/datasketches/cardinality/function/ToHllSketch.java
```diff
@@ -44,14 +46,17 @@ public ToHllSketch(final int logK) {
         this.logK = logK;
     }
 
-    public ToHllSketch(final HllSketch hllSketch) {
-        this.hllSketch = hllSketch;
+    public ToHllSketch(final HllSketch initHllSketch) {
+        this.initHllSketch = initHllSketch;
```
tb06904, line 50: Out of curiosity could the copy be taken here instead if not null? e.g.
```java
if (initHllSketch != null) {
    this.hllSketch = initHllSketch.copy();
}
```

library/sketches-library/src/test/java/uk/gov/gchq/gaffer/sketches/datasketches/cardinality/function/IterableToHllSketchTest.java
```diff
@@ -112,6 +112,24 @@ public void shouldCorrectlyCreateFromAnotherHllSketch() {
...
 
+    @Test
+    public void shouldCorrectlyCopyAnotherHllSketch() {
+        // Given
+        final HllSketch anotherSketch = new HllSketch();
+        anotherSketch.update("second");
+        anotherSketch.update("third");
+
+        IterableToHllSketch iterableToHllSketch = new IterableToHllSketch(anotherSketch);
+        final List<Object> input = Arrays.asList("one", "one", "two", "two", "three");
```
tb06904, line 123: Pretty minor but could be `List<String>` instead?

library/sketches-library/src/main/java/uk/gov/gchq/gaffer/sketches/datasketches/cardinality/function/ToHllSketch.java
t92549: I guess my thinking was someone could use the setter to overwrite it as null but that seems very unlikely. What do you think?

library/sketches-library/src/test/java/uk/gov/gchq/gaffer/sketches/datasketches/cardinality/function/IterableToHllSketchTest.java
t92549: When I do that `iterableToHllSketch.apply` is unhappy as it specifies `Iterable<Object>`.

tb06904: Fair makes sense

library/sketches-library/src/main/java/uk/gov/gchq/gaffer/sketches/datasketches/cardinality/function/ToHllSketch.java
t92549: I think it gets messy if another constructor is used

tb06904: Not quite sure I follow, do you mean by subclasses of this class? All I was thinking is it sounds like you want to allow an existing HllSketch to be used to init the one associated with this class but then just make sure not modify that original object so make a copy of it before using? This could be done in the constructor then the set method just does the same.

t92549: It could be done there, but you would still need this null check in the apply in case someone uses one of the other constructors.

tb06904: You can call the other constructors from one of the overloaded ones with `this()` (I edited my comment to add an example)

library/sketches-library/src/main/java/uk/gov/gchq/gaffer/sketches/datasketches/cardinality/function/IterableToHllSketch.java
```diff
@@ -94,6 +99,8 @@ public HllSketch getHllSketch() {
     }
 
     public void setHllSketch(final HllSketch hllSketch) {
-        this.hllSketch = hllSketch;
+        if (hllSketch != null) {
+            this.hllSketch = hllSketch.copy();
+        }
```
GCHQDeveloper314, line 104: Previously this had no usages. I'm not sure that changing it to ignore if the sketch is already set is the best approach.

library/sketches-library/src/main/java/uk/gov/gchq/gaffer/sketches/datasketches/cardinality/function/IterableToHllSketch.java
```diff
@@ -35,24 +36,28 @@
...
     public IterableToHllSketch() {
+        if (hllSketch == null) {
+            hllSketch = new HllSketch(logK);
+        }
     }
 
     public IterableToHllSketch(final int logK) {
-        this.logK = logK;
+        setLogK(logK);
+        if (hllSketch == null) {
```
GCHQDeveloper314, line 50: As this is the constructor, won't `hllSketch` always be null?

library/sketches-library/src/main/java/uk/gov/gchq/gaffer/sketches/datasketches/cardinality/function/IterableToHllSketch.java
```diff
@@ -35,24 +36,28 @@
 @Summary("Creates a new HllSketch instance and initialises it from the given iterable")
 public class IterableToHllSketch extends KorypheFunction<Iterable<Object>, HllSketch> {
     private int logK = DEFAULT_LOG_K;
+    @JsonIgnore
     private HllSketch hllSketch;
 
     public IterableToHllSketch() {
+        if (hllSketch == null) {
```
GCHQDeveloper314, line 43: As this is the constructor, won't `hllSketch` always be null?

library/sketches-library/src/main/java/uk/gov/gchq/gaffer/sketches/datasketches/cardinality/function/ToHllSketch.java
GCHQDeveloper314, line 1: See comment in other file relating to constructors.

library/sketches-library/src/main/java/uk/gov/gchq/gaffer/sketches/datasketches/cardinality/function/IterableToHllSketch.java
tb06904: I don't think that is what this change is doing, its checking if the newly supplied sketch is not null.

GCHQDeveloper314: You're right. This makes more sense, but I still question doing it. There's not much reason you'd want to set to `null`, but now if someone does want to do that it won't work and they will not know why unless they look at the code.

tb06904: I'm guessing its to avoid the null pointer exception as you can't call `copy()` on null as, a thought there could be an else statement for setting to null?

tb06904 (dismissed):

GCHQDeveloper314 (dismissed):

tb06904 (approved):

PR #3071, opened 2023-11-07
Gh-3069: Improve test coverage for GafferPopEdge.java

### LINKED ISSUE #3069: Improve Test Coverage for GafferPopEdge
Adding testing to Tinkerpop work to improve coverage - works to help complete #2990.

l46978:
# Related issue

- Resolve #3069

### DIFF
--- a/library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/GafferPopEdge.java
+++ b/library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/GafferPopEdge.java
@@ -130,11 +130,7 @@ public final class GafferPopEdge extends GafferPopElement implements Edge {
     private static GafferPopVertex checkVertex(final Object vertexId, final GafferPopGraph graph) {
         final GafferPopVertex gafferPopVertex;
         if (vertexId instanceof Vertex) {
-            if (vertexId instanceof GafferPopVertex) {
-                gafferPopVertex = checkVertex(((GafferPopVertex) vertexId), graph);
-            } else {
-                gafferPopVertex = new GafferPopVertex(GafferPopGraph.ID_LABEL, ((Vertex) vertexId).id(), graph);
-            }
+            gafferPopVertex = new GafferPopVertex(GafferPopGraph.ID_LABEL, ((Vertex) vertexId).id(), graph);
         } else {
             gafferPopVertex = new GafferPopVertex(GafferPopGraph.ID_LABEL, vertexId, graph);
         }
--- a/library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/GafferPopEdge.java
+++ b/library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/GafferPopEdge.java
@@ -130,7 +130,11 @@ public final class GafferPopEdge extends GafferPopElement implements Edge {
     private static GafferPopVertex checkVertex(final Object vertexId, final GafferPopGraph graph) {
         final GafferPopVertex gafferPopVertex;
         if (vertexId instanceof Vertex) {
-            gafferPopVertex = new GafferPopVertex(GafferPopGraph.ID_LABEL, ((Vertex) vertexId).id(), graph);
+            if (vertexId instanceof GafferPopVertex) {
+                gafferPopVertex = checkVertex(((GafferPopVertex) vertexId), graph);
+            } else {
+                gafferPopVertex = new GafferPopVertex(GafferPopGraph.ID_LABEL, ((Vertex) vertexId).id(), graph);
+            }
         } else {
             gafferPopVertex = new GafferPopVertex(GafferPopGraph.ID_LABEL, vertexId, graph);
         }
--- a/library/tinkerpop/src/test/java/uk/gov/gchq/gaffer/tinkerpop/GafferPopEdgeTest.java
+++ b/library/tinkerpop/src/test/java/uk/gov/gchq/gaffer/tinkerpop/GafferPopEdgeTest.java
@@ -35,6 +35,7 @@ import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertSame;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
 
 public class GafferPopEdgeTest {
 
@@ -180,5 +181,22 @@ public class GafferPopEdgeTest {
             .isThrownBy(() -> edge.property(TestPropertyNames.STRING, propValue1))
             .withMessageMatching("Updates are not supported");
 
+    }
+
+    @Test
+    public void shouldCreateNewGafferPopVertexWithVertexId() {
+        // Given
+        final GafferPopGraph graph = mock(GafferPopGraph.class);
+        final Vertex outVertex = mock(Vertex.class);
+        final Vertex inVertex = mock(Vertex.class);
+        when(inVertex.id()).thenReturn("inVertextId");
+        when(outVertex.id()).thenReturn("outVertextId");
+
+        // When
+        final GafferPopEdge edge = new GafferPopEdge(TestGroups.EDGE, outVertex, inVertex, graph);
+
+        // Then
+        assertSame(outVertex.id(), edge.outVertex().id());
+        assertSame(inVertex.id(), edge.inVertex().id());
     }
 }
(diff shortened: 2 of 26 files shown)


library/tinkerpop/src/test/java/uk/gov/gchq/gaffer/tinkerpop/GafferPopEdgeTest.java
```diff
@@ -112,4 +156,60 @@ public void shouldCreateReadableToString() {
...
+        final GafferPopVertex outVertex = new GafferPopVertex(GafferPopGraph.ID_LABEL, source, graph);
+        final GafferPopVertex inVertex = new GafferPopVertex(GafferPopGraph.ID_LABEL, dest, graph);
+        final List<Vertex> result = new ArrayList<>();
+
+        // When
+        final GafferPopEdge edge = new GafferPopEdge(TestGroups.EDGE, outVertex, inVertex, graph);
+
+        // Then
+        edge.vertices(Direction.OUT).forEachRemaining(result::add);
+        assertThat(result).contains(outVertex);
```
GCHQDeveloper314, line 175: For iterables we usually use the Assertj assertions directly with `containsExactly` and similar assertions.
```suggestion
        final GafferPopVertex inVertex = new GafferPopVertex(GafferPopGraph.ID_LABEL, dest, graph);

        // When
        final GafferPopEdge edge = new GafferPopEdge(TestGroups.EDGE, outVertex, inVertex, graph);

        // Then
        assertThat(edge.vertices(Direction.OUT)).toIterable().containsExactly(outVertex);
```

pom.xml
GCHQDeveloper314, line 1: I think this might have been included by mistake?

tb06904: The old variable that was being used `${project.reporting.outputDirectory}` doesn't exist which meant you couldn't check the coverage locally, new one is just standard location and lets the coverage site from jacoco to be viewed.

GCHQDeveloper314: You mean the HTML coverage pages? I've not had a problem with that before.
This change breaks the CI coverage - this PR has 0.00% coverage due to this. Changing this should not be done as part of this PR as other changes are required.

GCHQDev404 (commented):
What are the mocked graphs... mocking?
Are they not queried for things within the Generator?

pom.xml
tb06904: Where do the pages go by default if this variable is used? The CI still works as it uses the .exec file not the site, the PR genuinely has 0% coverage as only one line has changed and its not covered by tests.

library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/GafferPopEdge.java
```diff
@@ -130,11 +130,7 @@ public Vertex inVertex() {
     private static GafferPopVertex checkVertex(final Object vertexId, final GafferPopGraph graph) {
         final GafferPopVertex gafferPopVertex;
         if (vertexId instanceof Vertex) {
-            if (vertexId instanceof GafferPopVertex) {
```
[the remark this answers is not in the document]
l46978: This check is unnecessary as if the vertexId was an instance of GafferPopVertex the method below would execute not the current method.

l46978:
> What are the mocked graphs... mocking? Are they not queried for things within the Generator?

Not quite sure what you mean. The generator needs a graph to be instantiated but its not used in the scope of the test..

l46978:
> What are the mocked graphs... mocking? Are they not queried for things within the Generator?

Not quite sure what you mean. We need to mock the graphs as this is needed to create the edges and vertices.

library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/GafferPopEdge.java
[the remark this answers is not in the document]
t92549: It does look slightly different. Beforehand, it would not need to create a new object if the label was id

pom.xml
```diff
@@ -1544,7 +1544,7 @@
                                         ${project.build.directory}/coverage-reports/jacoco-ut.exec
                                     </dataFile>
                                     <outputDirectory>
-                                        ${project.reporting.outputDirectory}/jacoco-ut
+                                        ${project.build.directory}/coverage-reports/jacoco-ut
```
t92549, line 1547: Is the coverage going to the wrong place?

library/tinkerpop/src/test/java/uk/gov/gchq/gaffer/tinkerpop/GafferPopEdgeTest.java
```diff
@@ -100,6 +121,28 @@ public void shouldGetIterableOfEdgeProperties() {
         );
     }
 
+
+    @Test
+    public void shouldGetIterableOfSingleEdgeProperty() {
+        // Given
+        final String source = "source";
```
t92549, line 128: These could be put into static class variables and reused across the test.

library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/GafferPopEdge.java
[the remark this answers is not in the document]
t92549: It does look slightly different. Beforehand, it would not need to create a new object if the label was id

pom.xml
l46978: I have added tests to cover the line which was changed. The CI pipeline has now passed.

library/tinkerpop/src/test/java/uk/gov/gchq/gaffer/tinkerpop/GafferPopEdgeTest.java
```diff
@@ -28,18 +30,24 @@
...
+import static org.assertj.core.api.Assertions.assertThatExceptionOfType;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertSame;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
 
 public class GafferPopEdgeTest {
+
+    public static String source = "source";
```
t92549, line 42: Sorry, I should have mentioned, these are usually formatted like so:
```java
public static final String SOURCE = "source";
```
See GafferPopGraphTest:
https://github.com/gchq/Gaffer/blob/ecc9f4487b1106f3ceffde3031e2018f776f96b7/library/tinkerpop/src/test/java/uk/gov/gchq/gaffer/tinkerpop/GafferPopGraphTest.java#L47-L55

pom.xml
GCHQDeveloper314: If I build and run tests on the main branch with this module (`mvn clean install -Pcoverage -pl :tinkerpop`) then the coverage pages are placed in `target/site/jacoco-ut`.
The variable `${project.reporting.outputDirectory}` [is defined by the site plugin](https://maven.apache.org/plugins/maven-site-plugin/site-mojo.html#outputDirectory) and is [used in the default value for this property by jacoco](https://www.eclemma.org/jacoco/trunk/doc/report-mojo.html#outputDirectory). This config changes the default jacoco output directory to have different directories for UTs and ITs.
I see no reason to change the existing config for one which diverges from this approach, it has potential to cause issues where the path is hardcoded. If the site directory is missing for you then there's something else causing that problem.

GCHQDeveloper314 (changes requested):
See existing POM comment

GCHQDeveloper314 (approved):

tb06904 (dismissed):

library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/GafferPopEdge.java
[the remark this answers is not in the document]
t92549: Although I think the original code is messy, this change technically changes behaviour as a new object is always created, whereas previously sometimes it is just assigned:
https://github.com/gchq/Gaffer/blob/f03f6aea98a52ccbc530099d72301efa4fc428b0/library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/GafferPopEdge.java#L141-L144
Maybe this can be cleaned up or just returned to how it was before?

t92549 (changes requested):

library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/GafferPopEdge.java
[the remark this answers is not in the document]
t92549: On second look, I think it is fine as the other function would be called instead

t92549 (approved):

t92549 (approved):

Issue #3058, opened 2023-11-02
Add ApplyViewToElementsFunction ITs


t92549:
Currently, the Gaffer FederatedStoreITs use the `ConcatenateMergeFunction`, so they test the behaviour as it was in Gaffer 1.X:
https://github.com/gchq/Gaffer/blob/dde023fb2f4a2c585612f3a382f4f76b1933aada/store-implementation/federated-store/src/test/java/uk/gov/gchq/gaffer/federatedstore/integration/FederatedStoreITs.java#L42

This means that the default MergeFunction, `ApplyViewToElementsFunction`, is currently untested with Gaffer's suite of integration tests.

These tests should be enabled and any resulting bugs should be fixed.

**Note:** Due to the difference in behaviour of the merge functions, some tests will fail as they are expecting different behaviour, not because the merge function is broken. That being said, there will almost certainly be some bugs, for example: https://github.com/gchq/Gaffer/issues/3052

Issue #2819, opened 2022-11-14
FederatedStore DiscardOutput should not be wrapped 


GCHQDev404:
**Describe the bug**
In `FederatedStore.addAdditionalOperationHandlers` there is forEach to have `!Output` operations handled by `FederatedNoOutputHandler`. It is very likely `DiscardOutput` should be added to the filter. 

**Expected behavior**
`DiscardOutput` happens locally on the FederatedStore otherwise it can be sent to sub-graphs inside a `FederatedOperation`.

t92549:
Closed by https://github.com/gchq/Gaffer/pull/2820

PR #3072, opened 2023-11-08
Gh-3053: ApplyViewToElementsFunction optimisations

### LINKED ISSUE #3053: ApplyViewToElementsFunction unnecessary with one subgraph
When you only have one subgraph in your FederatedStore, and you use the default merge function `ApplyViewToElementsFunction`, it will still create an internal MapStore and collate the results inside it. This is unnecessary as all the aggregation and validation from the View has already happened at the subgraph level, and so does not need to happen again at the FederatedStore level.

This logic could be extended further. If two subgraphs do not contain any shared groups, then the internal MapStore is also unnecessary. I propose the `ApplyViewToElementsFunction` merge function is reworked so that only groups shared across multiple subgraphs go through the internal MapStore and have the View reapplied. This would be much more efficient.

tb06904:
Adds in logic to skip the default merge function if only one graph or if no common groups between graph schemas, instead it will just flatten the current list of results and return.

Unsure if there are existing tests for this functionality or the integration tests need fixing first under #3058 

# Related issue

- Resolve #3053

### DIFF
--- a/store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/operation/handler/impl/FederatedOperationHandler.java
+++ b/store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/operation/handler/impl/FederatedOperationHandler.java
@@ -23,6 +23,7 @@ import uk.gov.gchq.gaffer.core.exception.GafferCheckedException;
 import uk.gov.gchq.gaffer.federatedstore.FederatedStore;
 import uk.gov.gchq.gaffer.federatedstore.operation.FederatedOperation;
 import uk.gov.gchq.gaffer.federatedstore.util.ApplyViewToElementsFunction;
+import uk.gov.gchq.gaffer.federatedstore.util.ConcatenateMergeFunction;
 import uk.gov.gchq.gaffer.federatedstore.util.FederatedStoreUtil;
 import uk.gov.gchq.gaffer.graph.Graph;
 import uk.gov.gchq.gaffer.graph.GraphSerialisable;
@@ -109,19 +110,16 @@ public class FederatedOperationHandler<INPUT, OUTPUT> implements OperationHandle
         try {
             Object rtn = null;
 
-            final BiFunction mergeFunction = getMergeFunction(operation, store, context, isEmpty(resultsFromAllGraphs));
+            BiFunction mergeFunction = getMergeFunction(operation, store, context, isEmpty(resultsFromAllGraphs));
             final List<GraphSerialisable> graphs = getGraphs(operation, context, store);
 
             // If default merging and only have one graph or no common groups then just return the current results
             if (!graphs.isEmpty()
                     && mergeFunction instanceof ApplyViewToElementsFunction
                     && (graphs.size() == 1 || !graphsHaveCommonSchemaGroups(graphs))) {
-                LOGGER.info("Skipping merge function as not required when only one graph or no common groups");
-                // Just flatten current results to a single set to return
-                Iterable<?> iterableStream = () -> StreamSupport.stream(resultsFromAllGraphs.spliterator(), false)
-                    .flatMap(result -> StreamSupport.stream(((Iterable<?>) result).spliterator(), false))
-                    .iterator();
-                return iterableStream;
+                LOGGER.info("Returning flat list of results as complex merging not required when only one graph or no common groups");
+                // Just use the concatenate merge to flatten the results
+                mergeFunction = new ConcatenateMergeFunction();
             }
 
             // Reduce
--- a/store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/operation/handler/impl/FederatedOperationHandler.java
+++ b/store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/operation/handler/impl/FederatedOperationHandler.java
@@ -43,7 +43,9 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
 import java.util.function.BiFunction;
+import java.util.stream.Collector;
 import java.util.stream.Collectors;
+import java.util.stream.StreamSupport;
 
 import static com.google.common.collect.Iterables.isEmpty;
 import static java.util.Objects.nonNull;
@@ -118,10 +120,10 @@ public class FederatedOperationHandler<INPUT, OUTPUT> implements OperationHandle
                     && mergeFunction instanceof ApplyViewToElementsFunction
                     && (graphs.size() == 1 || !graphsHaveCommonSchemaGroups(graphs))) {
                 LOGGER.info("Skipping merge function as not required when only one graph or no common groups");
-                // Just flatten current results that are essentially a list of lists into one list and return
-                HashSet<Element> mergedResults = new HashSet<>();
-                resultsFromAllGraphs.forEach(result -> ((Iterable<Element>) result).iterator().forEachRemaining(mergedResults::add));
-                return mergedResults;
+                // Just flatten current results to a single set to return
+                return StreamSupport.stream(resultsFromAllGraphs.spliterator(), false)
+                    .flatMap(result -> StreamSupport.stream(((Iterable<?>) result).spliterator(), false))
+                    .collect(Collectors.toSet());
             }
 
             // Reduce
--- a/store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/operation/handler/impl/FederatedOperationHandler.java
+++ b/store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/operation/handler/impl/FederatedOperationHandler.java
@@ -121,9 +121,10 @@ public class FederatedOperationHandler<INPUT, OUTPUT> implements OperationHandle
                     && (graphs.size() == 1 || !graphsHaveCommonSchemaGroups(graphs))) {
                 LOGGER.info("Skipping merge function as not required when only one graph or no common groups");
                 // Just flatten current results to a single set to return
-                return StreamSupport.stream(resultsFromAllGraphs.spliterator(), false)
+                Iterable<?> iterableStream = () -> StreamSupport.stream(resultsFromAllGraphs.spliterator(), false)
                     .flatMap(result -> StreamSupport.stream(((Iterable<?>) result).spliterator(), false))
-                    .collect(Collectors.toSet());
+                    .iterator();
+                return iterableStream;
             }
 
             // Reduce
--- a/store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/operation/handler/impl/FederatedOperationHandler.java
+++ b/store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/operation/handler/impl/FederatedOperationHandler.java
@@ -179,7 +179,7 @@ public class FederatedOperationHandler<INPUT, OUTPUT> implements OperationHandle
                 Set<String> firstGroupSet = graphSchemas.get(i).getGroups();
                 Set<String> secondGroupSet = graphSchemas.get(j).getGroups();
                 if (!Collections.disjoint(firstGroupSet, secondGroupSet)) {
-                     LOGGER.debug("Found common schema groups between requested graphs");
+                    LOGGER.debug("Found common schema groups between requested graphs");
                     return true;
                 }
             }
--- a/store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/operation/handler/impl/FederatedOperationHandler.java
+++ b/store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/operation/handler/impl/FederatedOperationHandler.java
@@ -16,6 +16,9 @@
 
 package uk.gov.gchq.gaffer.federatedstore.operation.handler.impl;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 import uk.gov.gchq.gaffer.core.exception.GafferCheckedException;
 import uk.gov.gchq.gaffer.data.element.Element;
 import uk.gov.gchq.gaffer.federatedstore.FederatedStore;
@@ -53,6 +56,7 @@ import static uk.gov.gchq.gaffer.federatedstore.util.FederatedStoreUtil.updateOp
  */
 @Since("2.0.0")
 public class FederatedOperationHandler<INPUT, OUTPUT> implements OperationHandler<FederatedOperation<INPUT, OUTPUT>> {
+    private static final Logger LOGGER = LoggerFactory.getLogger(FederatedOperationHandler.class);
 
     public static final String ERROR_WHILE_RUNNING_OPERATION_ON_GRAPHS_FORMAT = "Error while running operation on graphs, due to: %s";
     public static final String MERGE_FUNCTION_GET_GRAPHS = FederatedOperationHandler.class.getCanonicalName() + "merge.function.getGraphs";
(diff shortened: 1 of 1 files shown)


store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/operation/handler/impl/FederatedOperationHandler.java
```diff
@@ -140,4 +161,31 @@ private List<GraphSerialisable> getGraphs(final FederatedOperation<INPUT, OUTPUT
...
+     * @return True if any schema groups are shared.
+     */
+    private boolean graphsHaveCommonSchemaGroups(final List<GraphSerialisable> graphs) {
+        // Check if any of the graphs have common groups in their schemas
+        List<Schema> graphSchemas = graphs.stream()
+            .map(GraphSerialisable::getSchema)
+            .collect(Collectors.toList());
+
+        // Compare all schemas against each other
+        for (int i = 0; i < graphSchemas.size() - 1; i++) {
```
t92549, line 178: Might be nicer to use [cartesianProduct](https://guava.dev/releases/snapshot/api/docs/com/google/common/collect/Lists.html#cartesianProduct(java.util.List...))

tb06904: I was trying not to rely on an external library with cartesian product I'd still have to do a double loop as you'd end up with a list of lists that have just have two elements.

t92549: I was thinking `for pair in cartesianProduct(schemas, schemas)`, but you would have to check the schemas aren't the same so a tad awkward

tb06904: Yeah I'm not sure it would change too much of the logic it would end up something like:
```java
List<List<Schema>> product = Lists.cartesianProduct(graphSchemas, graphSchemas);

for (List<Schema> pair : product) {
    Set<String> firstGroupSet = pair.get(0).getGroups();
    Set<String> secondGroupSet = pair.get(1).getGroups();
    if (firstGroupSet.stream().anyMatch(secondGroupSet::contains)) {
        LOGGER.debug("Found common schema groups between requested graphs");
        return true;
    }
}
```
I'd rather not use an external library if its not required

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/operation/handler/impl/FederatedOperationHandler.java
```diff
@@ -97,13 +106,25 @@ private Iterable getAllGraphResults(final FederatedOperation<INPUT, OUTPUT> oper
...
-    private Object mergeResults(final Iterable resultsFromAllGraphs, final FederatedOperation operation, final FederatedStore store) throws OperationException {
+    private Object mergeResults(final Iterable resultsFromAllGraphs, final FederatedOperation<INPUT, OUTPUT> operation, final FederatedStore store) throws OperationException {
         try {
             Object rtn = null;
 
             final BiFunction mergeFunction = getMergeFunction(operation, store, context, isEmpty(resultsFromAllGraphs));
+            final List<GraphSerialisable> graphs = getGraphs(operation, context, store);
+
+            // If default merging and only have one graph or no common groups then just return the current results
+            if (!graphs.isEmpty()
```
GCHQDeveloper314, line 119: Do we really need to check if `graphs` is empty? I _believe_ that if there are no graph then this method doesn't get run.

t92549 (dismissed):

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/operation/handler/impl/FederatedOperationHandler.java
```diff
@@ -97,13 +106,25 @@ private Iterable getAllGraphResults(final FederatedOperation<INPUT, OUTPUT> oper
...
+                    && mergeFunction instanceof ApplyViewToElementsFunction
+                    && (graphs.size() == 1 || !graphsHaveCommonSchemaGroups(graphs))) {
+                LOGGER.info("Skipping merge function as not required when only one graph or no common groups");
+                // Just flatten current results that are essentially a list of lists into one list and return
+                HashSet<Element> mergedResults = new HashSet<>();
+                resultsFromAllGraphs.forEach(result -> ((Iterable<Element>) result).iterator().forEachRemaining(mergedResults::add));
```
GCHQDeveloper314: We should instead use the concat merge function which exists for this. This is also inefficient as it iterates through everything which does not preserve lazy behaviour (see #2819). Would be easiest to swap out the `mergeFunction` and use the existing code below.

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/operation/handler/impl/FederatedOperationHandler.java
GCHQDeveloper314: I'd agree regarding the use of external libraries. I've added a separate suggestion regarding the loop.

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/operation/handler/impl/FederatedOperationHandler.java
```diff
@@ -140,4 +161,31 @@ private List<GraphSerialisable> getGraphs(final FederatedOperation<INPUT, OUTPUT
...
+            .map(GraphSerialisable::getSchema)
+            .collect(Collectors.toList());
+
+        // Compare all schemas against each other
+        for (int i = 0; i < graphSchemas.size() - 1; i++) {
+            for (int j = i + 1; j < graphSchemas.size(); j++) {
+                // Compare each schema against the others to see if common groups
+                Set<String> firstGroupSet = graphSchemas.get(i).getGroups();
+                Set<String> secondGroupSet = graphSchemas.get(j).getGroups();
+                if (firstGroupSet.stream().anyMatch(secondGroupSet::contains)) {
```
GCHQDeveloper314, line 185: There's no reason to use C style looping here.
```suggestion
        for (Schema schemaLeft : graphSchemas) {
            for (Schema schemaRight : graphSchemas) {
                // Compare each schema against the others to see if common groups
                Set<String> groupSetLeft = schemaLeft.getGroups();
                Set<String> groupSetRight = schemaRight.getGroups();
                if (groupSetLeft.stream().anyMatch(groupSetRight::contains)) {
```

GCHQDeveloper314 (changes requested):
The comment on Lazy iterables needs to be fixed.

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/operation/handler/impl/FederatedOperationHandler.java
tb06904: This will do duplicate comparisons e.g. if you have `[1,2,3]` and loop like this you will end up comparing both directions:
```
[1,1]
[1,2]
[1,3]
```
and
```
[1,1]
[2,1]
[3,1]
``

File: store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/operation/handler/impl/FederatedOperationHandler.java (hunk 2 above)
Comment (tb06904) [reply store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/operation/handler/impl/FederatedOperationHandler.java:119]: I thought the same but lots of the federation unit tests fail without this check in as it seems lots of the tests don't create a graph but still run the merge function.

File: store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/operation/handler/impl/FederatedOperationHandler.java (hunk 4 above)
Comment (GCHQDeveloper314) [reply store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/operation/handler/impl/FederatedOperationHandler.java:185]: Ah yes, had not looked at the loop properly. I think there's no easy way to do this in Java 8, which is a shame. Could use `Collections.disjoint` instead of `stream`?

File: store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/operation/handler/impl/FederatedOperationHandler.java (hunk 2 above)
Comment (GCHQDeveloper314) [reply store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/operation/handler/impl/FederatedOperationHandler.java:119]: Ok, when used normally I think it doesn't get run. Not ideal to have this to work with tests, could raise an issue for that. This can stay for this PR.

File: store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/operation/handler/impl/FederatedOperationHandler.java (hunk 3 above)
Comment (tb06904) [reply store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/operation/handler/impl/FederatedOperationHandler.java]: This is something I noticed after opening, I believe moving to `flatMap()` rather than concat is preferable for this use case as there might be more than two graphs. Updated to now return a lazy iterable from the merged streams.

File: store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/operation/handler/impl/FederatedOperationHandler.java (hunk 2 above)
Comment (tb06904) [reply store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/operation/handler/impl/FederatedOperationHandler.java:119]: Agreed I think its more an edge case with the testing.

File: store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/operation/handler/impl/FederatedOperationHandler.java (hunk 4 above)
Comment (tb06904) [reply store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/operation/handler/impl/FederatedOperationHandler.java:186]: Yeah it is unfotunate, good point `disjoint` might be bit cleaner to use here.

File: store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/operation/handler/impl/FederatedOperationHandler.java (hunk 3 above)
Comment (GCHQDeveloper314) [reply store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/operation/handler/impl/FederatedOperationHandler.java]: That is better, but functionally the same as using the concat iterator for what I can tell.

File: store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/operation/handler/impl/FederatedOperationHandler.java (hunk 2 above)
Comment (t92549) [reply store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/operation/handler/impl/FederatedOperationHandler.java:116]: I think if you have an empty FederatedStore this might get run?

Comment (GCHQDeveloper314) [reply store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/operation/handler/impl/FederatedOperationHandler.java:116]: I thought that with an empty store there was some logic to stop. Possibly in a similar place to code which logs the message about empty schemas. In any case this check is required so I'll resolve this.

File: store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/operation/handler/impl/FederatedOperationHandler.java (hunk 3 above)
Comment (GCHQDeveloper314) [reply store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/operation/handler/impl/FederatedOperationHandler.java]: I don't know of any reason. The concat merge function is simple as the merging is done with other classes which are also used elsewhere in Gaffer. Potentially this could all be refactored to use streams, but I don't see many benefits. `new ChainedIterable<>(item1, item2)` is easier to read than a multiline stream statement.

Comment (tb06904) [reply store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/operation/handler/impl/FederatedOperationHandler.java]: Streams do have a lot of advantages though as spliterators have a much lower per element cost than iterators and reduced memory bus activity to implement for the merge function its as easy as something like:

```java
Stream<Object> stateStream = StreamSupport.stream(state.spliterator(), false);
Stream<Object> updateStream = StreamSupport.stream(updateSafe.spliterator(), false);
Iterable<Object> merged = () -> Stream.concat(stateStream, updateStream).iterator();
return merged;
```

GCHQDeveloper314: Could be worth raising a tech debt ticket to refactor all of this (total rework/remove of these different classes) with some notes about streams. For this issue/PR we should use the preexisting merge functions and not introduce a mix of two different approaches.

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/operation/handler/impl/FederatedOperationHandler.java
```diff
@@ -180,8 +180,8 @@ private boolean graphsHaveCommonSchemaGroups(final List<GraphSerialisable> graph
                 // Compare each schema against the others to see if common groups
                 Set<String> firstGroupSet = graphSchemas.get(i).getGroups();
                 Set<String> secondGroupSet = graphSchemas.get(j).getGroups();
-                if (firstGroupSet.stream().anyMatch(secondGroupSet::contains)) {
-                    LOGGER.debug("Found common schema groups between requested graphs");
+                if (!Collections.disjoint(firstGroupSet, secondGroupSet)) {
+                     LOGGER.debug("Found common schema groups between requested graphs");
```
t92549, line 184:
```suggestion
                    LOGGER.debug("Found common schema groups between requested graphs");
```

GCHQDeveloper314 (approved):

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/operation/handler/impl/FederatedOperationHandler.java
```diff
@@ -97,13 +105,23 @@ private Iterable getAllGraphResults(final FederatedOperation<INPUT, OUTPUT> oper
...
+            BiFunction mergeFunction = getMergeFunction(operation, store, context, isEmpty(resultsFromAllGraphs));
+            final List<GraphSerialisable> graphs = getGraphs(operation, context, store);
+
+            // If default merging and only have one graph or no common groups then just return the current results
+            if (!graphs.isEmpty()
+                    && mergeFunction instanceof ApplyViewToElementsFunction
+                    && (graphs.size() == 1 || !graphsHaveCommonSchemaGroups(graphs))) {
+                LOGGER.info("Returning flat list of results as complex merging not required when only one graph or no common groups");
+                // Just use the concatenate merge to flatten the results
+                mergeFunction = new ConcatenateMergeFunction();
```
t92549, line 121: This could maybe be tested to improve coverage?

tb06904: Coverage should hopefully be sorted now, there was an issue with how it was getting the graphs which is now fixed so existing tests appear to cover this functionality.

t92549 (approved):

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/operation/handler/impl/FederatedOperationHandler.java
```diff
@@ -48,25 +54,26 @@
  */
 @Since("2.0.0")
 public class FederatedOperationHandler<INPUT, OUTPUT> implements OperationHandler<FederatedOperation<INPUT, OUTPUT>> {
+    private static final Logger LOGGER = LoggerFactory.getLogger(FederatedOperationHandler.class);
 
     public static final String ERROR_WHILE_RUNNING_OPERATION_ON_GRAPHS_FORMAT = "Error while running operation on graphs, due to: %s";
-    public static final String MERGE_FUNCTION_GET_GRAPHS = FederatedOperationHandler.class.getCanonicalName() + "merge.function.getGraphs";
+    private List<GraphSerialisable> graphs;
```
GCHQDev404, line 60: So these handlers now retain the state of graphs from the previous Operation?

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/operation/handler/impl/FederatedOperationHandler.java
```diff
@@ -48,25 +54,26 @@
...
 
         return mergeResults(allGraphResults, operation, (FederatedStore) store);
     }
 
-    private Iterable getAllGraphResults(final FederatedOperation<INPUT, OUTPUT> operation, final FederatedStore store) throws OperationException {
+    private Iterable getAllGraphResults(final FederatedOperation<INPUT, OUTPUT> operation) throws OperationException {
         try {
             List<Object> results;
-            final Collection<GraphSerialisable> graphs = getGraphs(operation, context, store);
-            context.setVariable(MERGE_FUNCTION_GET_GRAPHS, graphs.stream().map(GraphSerialisable::getGraphId).collect(Collectors.toList()));
```
GCHQDev404, line 68: Information of graphs already existed and was retained here. This one line could have been modified to store the whole GraphSerialisable.

PR #3099, opened 2023-11-16
Gh-3098: Improve Testing of GafferEntityGenerator

### LINKED ISSUE #3098: Improve the testing of the GafferEntityGenerator class
Improve the testing of the `GafferEntityGenerator`so that it has adequate coverage e.g. > 80%

Could also look at improving the test quality using mutation testing to check for areas of improvement.

Minor changes to the class may be permitted to improve tests.

tb06904:
Improves tests for the `GafferEntityGenerator` coverage now > 80%

Small tweak to the main class so it checks for null values and to use easier to read lambdas.

# Related issue

- Resolve #3098

### DIFF
--- a/library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/generator/GafferEntityGenerator.java
+++ b/library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/generator/GafferEntityGenerator.java
@@ -16,33 +16,29 @@
 
 package uk.gov.gchq.gaffer.tinkerpop.generator;
 
-import org.apache.tinkerpop.gremlin.structure.Property;
-import org.apache.tinkerpop.gremlin.structure.VertexProperty;
-
 import uk.gov.gchq.gaffer.data.element.Entity;
 import uk.gov.gchq.gaffer.data.generator.OneToOneElementGenerator;
 import uk.gov.gchq.gaffer.tinkerpop.GafferPopVertex;
 
-import java.util.Iterator;
-
 public class GafferEntityGenerator implements OneToOneElementGenerator<GafferPopVertex> {
     @Override
     public Entity _apply(final GafferPopVertex vertex) {
+        if (vertex == null) {
+            throw new IllegalArgumentException("Unable to convert a null GafferPopVertex Object");
+        }
+
         final Entity entity = new Entity(vertex.label(), vertex.id());
-        final Iterator<VertexProperty<Object>> vertPropItr = vertex.properties();
-        while (vertPropItr.hasNext()) {
-            final VertexProperty<Object> vertProp = vertPropItr.next();
-            if (null != vertProp.key()) {
+        // Tinkerpop allows nested properties under a key for Gaffer we need to flatten these so only one property per key
+        vertex.properties().forEachRemaining(vertProp -> {
+            if (vertProp.key() != null) {
                 entity.putProperty(vertProp.key(), vertProp.value());
             }
-            final Iterator<Property<Object>> propItr = vertProp.properties();
-            while (propItr.hasNext()) {
-                final Property<Object> prop = propItr.next();
-                if (null != prop.key()) {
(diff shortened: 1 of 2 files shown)


library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/generator/GafferEntityGenerator.java
```diff
@@ -16,33 +16,29 @@
...
                 entity.putProperty(vertProp.key(), vertProp.value());
             }
-            final Iterator<Property<Object>> propItr = vertProp.properties();
-            while (propItr.hasNext()) {
-                final Property<Object> prop = propItr.next();
-                if (null != prop.key()) {
+            vertProp.properties().forEachRemaining(prop -> {
+                if (prop.key() != null) {
                     entity.putProperty(prop.key(), prop.value());
                 }
```
GCHQDeveloper314, line 39: This is essentially repeated code. It could be reduced by defining a lambda earlier on. Though I'm not sure if this is worth it as it would only save a few lines of code. Might be more readable though?

GCHQDeveloper314 (commented):
You've added null checks for the vertex properties, but this is showing as untested.

tb06904:
> You've added null checks for the vertex properties, but this is showing as untested.

I haven't added them they were in the existing logic although looking at the code a bit more is impossible to actually make a GafferPop Edge or Vertex with a property that has a null key as it is validated when its added and also the properties are technically stored in a HashMap which also fundamentally does not allow it. Therefore these checks likely can just be removed as they are adding nothing.

library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/generator/GafferEntityGenerator.java
tb06904: Not sure its possible with one lambda as you need to loop over the first iterable before you have access to the nested iterable open to suggestions but simplest I can think of is remove the null check as its not adding anything (see other comment) and do something like:

```java
vertex.properties().forEachRemaining(vertProp -> {
    entity.putProperty(vertProp.key(), vertProp.value());
    vertProp.properties().forEachRemaining(prop -> entity.putProperty(prop.key(), prop.value()));
});
```

GCHQDeveloper314: I was thinking of a function as a variable which could be reused inside the existing lambdas. But this is a moot point if the null check is removed as that will condense these as you've shown.

GCHQDeveloper314 (dismissed):

library/tinkerpop/src/test/java/uk/gov/gchq/gaffer/tinkerpop/generator/GafferEntityGeneratorTest.java
```diff
@@ -46,21 +55,67 @@ public void shouldConvertGafferPopVertexToGafferEntity() {
...
+        vertexWithProps.property(Cardinality.list, TestPropertyNames.STRING, propValue);
 
         final GafferEntityGenerator generator = new GafferEntityGenerator();
+        // When
+        final Entity entityNoProps = generator._apply(vertexNoProps);
+        final Entity entityWithProps = generator._apply(vertexWithProps);
+
+        // Then
+        assertThat(entityNoProps.getGroup()).isEqualTo(TestGroups.ENTITY);
+        assertThat(entityNoProps.getVertex()).isEqualTo(vertexNoPropsId);
```
t92549, line 70: Is it worth checking this has empty properties?

tb06904: Good point wouldn't harm to

t92549 (approved):

GCHQDeveloper314 (approved):

PR #3102, opened 2023-11-17
Gh-3091: Improve testing of GafferPop NamedOperationService

### LINKED ISSUE #3091: Improve testing of GafferPopNamedOperationService
Improved testing and test coverage of `GafferPopNamedOperationService.java` and `GafferPopNamedOperationServiceFactory.java`.

Improved reliability of the codebase.

GCHQDeveloper314:
Includes a fix for missing Generics in the class itself

# Related issue

- Resolve #3091

### DIFF
--- a/library/tinkerpop/src/test/java/uk/gov/gchq/gaffer/tinkerpop/service/GafferPopNamedOperationServiceIT.java
+++ b/library/tinkerpop/src/test/java/uk/gov/gchq/gaffer/tinkerpop/service/GafferPopNamedOperationServiceIT.java
@@ -1,0 +1,170 @@
+package uk.gov.gchq.gaffer.tinkerpop.service;
+
+import org.apache.commons.configuration2.BaseConfiguration;
+import org.apache.commons.configuration2.Configuration;
+import org.apache.tinkerpop.gremlin.process.traversal.dsl.graph.GraphTraversal;
+import org.apache.tinkerpop.gremlin.process.traversal.dsl.graph.GraphTraversalSource;
+import org.junit.jupiter.api.Test;
+
+import uk.gov.gchq.gaffer.commonutil.StreamUtil;
+import uk.gov.gchq.gaffer.graph.Graph;
+import uk.gov.gchq.gaffer.graph.GraphConfig;
+import uk.gov.gchq.gaffer.mapstore.MapStoreProperties;
+import uk.gov.gchq.gaffer.named.operation.AddNamedOperation;
+import uk.gov.gchq.gaffer.operation.OperationChain;
+import uk.gov.gchq.gaffer.operation.OperationException;
+import uk.gov.gchq.gaffer.operation.impl.Count;
+import uk.gov.gchq.gaffer.operation.impl.get.GetAllElements;
+import uk.gov.gchq.gaffer.tinkerpop.GafferPopGraph;
+import uk.gov.gchq.gaffer.tinkerpop.GafferPopGraphTest;
+import uk.gov.gchq.gaffer.user.User;
+
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatNoException;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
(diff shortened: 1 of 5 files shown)


library/tinkerpop/src/test/java/uk/gov/gchq/gaffer/tinkerpop/service/GafferPopNamedOperationServiceIT.java
```diff
@@ -0,0 +1,186 @@
...
+        return new Graph.Builder()
+                .config(new GraphConfig.Builder()
+                        .graphId("graph1")
+                        .build())
+                .storeProperties(PROPERTIES)
+                .addSchemas(StreamUtil.openStreams(this.getClass(), "/gaffer/schema"))
+                .build();
+    }
+
+    private AddNamedOperation getOpChainElementCount(String name) {
```
t92549, line 168: This could maybe be called something like `getAddNamedOpElementCount`

library/tinkerpop/src/test/java/uk/gov/gchq/gaffer/tinkerpop/service/GafferPopNamedOperationServiceIT.java
```diff
@@ -0,0 +1,186 @@
...
+    @Test
+    void shouldAddNamedOperation() {
+        // Given
+        final GafferPopGraph gafferPopGraph = GafferPopGraph.open(TEST_CONFIGURATION, getGafferGraph());
+        GraphTraversalSource g = gafferPopGraph.traversal();
+
+        // When
+        Map<String, Object> addParams = new HashMap<>();
+        addParams.put("name", "testNamedOp");
+        addParams.put("opChain", getOpChainElementCount(null).getOperationChainAsString());
```
t92549, line 116: This adds a named op with name `null`, which I didn't know would work

library/tinkerpop/src/test/java/uk/gov/gchq/gaffer/tinkerpop/service/GafferPopNamedOperationServiceTest.java
```diff
@@ -0,0 +1,54 @@
...
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+import static org.mockito.Mockito.mock;
+
+public class GafferPopNamedOperationServiceTest {
+    private final GafferPopGraph graph = mock(GafferPopGraph.class);
+
+    @Test
+    void shouldGetType() {
+        // Given
+        GafferPopNamedOperationService<String, String> namedOpService = new GafferPopNamedOperationService<>(graph);
```
t92549, line 36: None of the parameters have `final`. I don't think they need it, just seems to be everywhere else in Gaffer

GCHQDeveloper314: I don't think so either, but happy to add this.

library/tinkerpop/src/test/java/uk/gov/gchq/gaffer/tinkerpop/service/GafferPopNamedOperationServiceIT.java
GCHQDeveloper314: Not quite. It creates a named op which has a null name but this name is never used. The operation is created and `getOperationChainAsString()` is then used to extract only the chain from it as JSON. This chain is then used with the name "testNamedOp" to create and add a new named op.

The creation of the named operation object is only a helper to easily get the operation chain as a string, instead of having a difficult to read hardcoded JSON sting.

library/tinkerpop/src/test/java/uk/gov/gchq/gaffer/tinkerpop/service/GafferPopNamedOperationServiceIT.java
GCHQDeveloper314: I think previously I was using an op chain, but there was no way to directly get the JSON representation of the chain with that. So I switched to the full operation, but forgot to change the method name. I'll fix this.

t92549 (approved):

t92549 (approved):

tb06904 (approved):

PR #3111, opened 2023-11-22
Gh-3106: Refactor GafferPopNamedOperationService

### LINKED ISSUE #3106: Refactor and improve GafferPopNamedOperationService
During work on #3091, it was noticed that `GafferPopNamedOperationService` would benefit from some refactoring and addition of Javadoc.

Currently there's a single method with all the logic and this is nested with various `if`s. Splitting this out to separate methods and using a case statement would improve readability.

GCHQDeveloper314:
Some Generics handing is not ideal. This is due to the implementation of generators, how this class is used and TinkerPop itself (rawtype `Map`).

# Related issue

- Resolve #3106

### DIFF
--- a/library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/service/GafferPopNamedOperationService.java
+++ b/library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/service/GafferPopNamedOperationService.java
@@ -37,12 +37,13 @@ import java.util.stream.StreamSupport;
  * Service for running Gaffer Named Operations
  *
  * <p>This service should be called at the start
- * of a traversal.
- * @see <a href="https://tinkerpop.apache.org/docs/3.7.0/reference/#call-step">
- * Tinkerpop Call Step Documentation</a></p>
+ * of a traversal.</p>
  *
  * @param <I> Ignored
  * @param <R> Ignored
+ *
+ * @see <a href="https://tinkerpop.apache.org/docs/3.7.0/reference/#call-step">
+ * Tinkerpop Call Step Documentation</a>
  */
 public class GafferPopNamedOperationService<I, R> implements Service<I, R> {
     private final GafferPopGraph graph;
--- a/library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/service/GafferPopNamedOperationService.java
+++ b/library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/service/GafferPopNamedOperationService.java
@@ -74,7 +74,7 @@ public class GafferPopNamedOperationService<I, R> implements Service<I, R> {
         } else if (params.containsKey("add")) {
             return addNamedOperation((Map<String, String>) params.get("add"));
         } else {
-            throw new IllegalStateException("Missing parameter");
+            throw new IllegalStateException("Missing parameter, either 'execute' or 'add' expected");
         }
     }
 
--- a/library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/service/GafferPopNamedOperationService.java
+++ b/library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/service/GafferPopNamedOperationService.java
@@ -78,11 +78,21 @@ public class GafferPopNamedOperationService<I, R> implements Service<I, R> {
         }
     }
 
+    /**
+     * Find the requested Named Operation and depending on its output class
+     * either execute and then convert all {@link uk.gov.gchq.gaffer.data.element.Element}
+     * to {@link GafferPopElement}, to just execute and return the result as
+     * an Iterable.
+     *
+     * @param name Name of the Named Operation to execute
+     * @return results of the operation execution
+     */
     protected CloseableIterator<R> executeNamedOperation(final String name) {
         // Fetch details for the requested Named Operation or throw an exception if it's not found
         Iterable<NamedOperationDetail> allNamedOps = graph.execute(new OperationChain.Builder().first(new GetAllNamedOperations()).build());
         NamedOperationDetail namedOperation = StreamSupport.stream(allNamedOps.spliterator(), false)
-                .filter((NamedOperationDetail nd) -> nd.getOperationName().equals(name)).findFirst()
+                .filter(nd -> nd.getOperationName().equals(name))
+                .findFirst()
                 .orElseThrow(() -> new IllegalStateException("Named Operation not found"));
 
         // If the Named Operation outputs an Iterable, then execute and convert Elements to GafferPopElement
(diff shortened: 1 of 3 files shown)


library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/service/GafferPopNamedOperationService.java
```diff
@@ -44,44 +57,61 @@ public Type getType() {
...
         } else if (params.containsKey("add")) {
-            final Map<String, Object> addParams = (Map) params.get("add");
-            graph.execute(new OperationChain.Builder()
-                    .first(new AddNamedOperation.Builder()
-                            .name((String) addParams.get("name"))
-                            .operationChain((String) addParams.get("opChain"))
-                            .build())
-                    .build());
-            return CloseableIterator.empty();
+            return addNamedOperation((Map<String, String>) params.get("add"));
```
tb06904, line 75: This is technically an unchecked cast is it work wrapping in a try?

library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/service/GafferPopNamedOperationService.java
```diff
@@ -44,44 +57,61 @@ public Type getType() {
...
         } else {
             throw new IllegalStateException("Missing parameter");
```
tb06904, line 77: Maybe could have a bit more information in this exception

library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/service/GafferPopNamedOperationService.java
```diff
@@ -44,44 +57,61 @@ public Type getType() {
...
         }
     }
 
+    protected CloseableIterator<R> executeNamedOperation(final String name) {
```
tb06904, line 81: Should have javadocs on all methods not just public ones

library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/service/GafferPopNamedOperationService.java
```diff
@@ -44,44 +57,61 @@ public Type getType() {
...
+        // Fetch details for the requested Named Operation or throw an exception if it's not found
+        Iterable<NamedOperationDetail> allNamedOps = graph.execute(new OperationChain.Builder().first(new GetAllNamedOperations()).build());
+        NamedOperationDetail namedOperation = StreamSupport.stream(allNamedOps.spliterator(), false)
+                .filter((NamedOperationDetail nd) -> nd.getOperationName().equals(name)).findFirst()
```
tb06904, line 85: Think this is an unnecessary cast to `NamedOperationDetail` as the `Iterable` has already been declared as this type.

Also `.findFirst()` should be on the next line to be consistent with formatting.

library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/service/GafferPopNamedOperationService.java
GCHQDeveloper314: I don't think so. It's always the case that these will be strings. Potentially not if the service isn't being called correctly, but I'm not sure about that.

library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/service/GafferPopNamedOperationService.java
GCHQDeveloper314: It's not great but I didn't want to change the code around too much. @t92549 - you wrote this, what do you think?

library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/service/GafferPopNamedOperationService.java
GCHQDeveloper314: This isn't a cast. It just provides the full type for the lambda parameter. However, you're correct that this isn't required. I'll change the formatting too.

library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/service/GafferPopNamedOperationService.java
```diff
@@ -44,44 +57,61 @@ public Type getType() {
...
+                .orElseThrow(() -> new IllegalStateException("Named Operation not found"));
+
+        // If the Named Operation outputs an Iterable, then execute and convert Elements to GafferPopElement
+        // Else execute and wrap the output as an iterator
+        if (namedOperation.getOperationChainWithDefaultParams().getOutputClass().equals(Iterable.class)) {
```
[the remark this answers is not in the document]
GCHQDeveloper314: This doesn't confirm that the Iterable return will be of Elements, only that it is an Iterable.

The key is how to tell **before** the results are returned. This is the closest we can do. I don't think the Named Operation stores the type of Object that is inside the Iterable anyway, so we can't get this information.

With the results, it's easy to tell if the results of an Operation are an Iterable. To tell if the Iterable has Elements is more difficult, but can be done. However, we need to know before we get any results in order to run the GafferPop Element generator. We can't get the results and somehow put them back in to run the generator. It might be possible to conditionally run the generator within the Gaffer Operation? Or we could create a wrapped Iterable with the generator to be able to do this after getting the results.

library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/service/GafferPopNamedOperationService.java
t92549: Yeah could be improved. I think this will be displayed in the gremlin console so worth keeping it brief but informative

library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/service/GafferPopNamedOperationService.java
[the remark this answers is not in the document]
t92549: If it can't be determined before the operation is executed then I agree this is the best we can do, as I think we want the generator to run inside Gaffer.

library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/service/GafferPopNamedOperationService.java
GCHQDeveloper314: Done

tb06904 (approved):

library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/service/GafferPopNamedOperationService.java
```diff
@@ -44,44 +57,79 @@ public Type getType() {
...
+     * either execute and then convert all {@link uk.gov.gchq.gaffer.data.element.Element}
+     * to {@link GafferPopElement}, to just execute and return the result as
+     * an Iterable.
+     *
+     * @param name Name of the Named Operation to execute
+     * @return results of the operation execution
+     */
+    protected CloseableIterator<R> executeNamedOperation(final String name) {
+        // Fetch details for the requested Named Operation or throw an exception if it's not found
+        Iterable<NamedOperationDetail> allNamedOps = graph.execute(new OperationChain.Builder().first(new GetAllNamedOperations()).build());
```
t92549, line 92: Eventually it would be nice if Gaffer had an endpoint to get named operations by name, this would be a lot more effecient

t92549 (approved):

PR #3116, opened 2023-11-23
Gh-3113: Fix regex in create release branch workflow

### LINKED ISSUE #3113: Fix create release branch workflow regex
In the "Bump latest tag variable version" step, there is an incorrect regex:
https://github.com/gchq/Gaffer/blob/1ef3621323064ab6a15c84fa554d35fe1dea003d/.github/workflows/create-release-branch.yaml#L41
It should set the patch version to 0 but does not

t92549:
# Related issue

- Resolve #3113

### DIFF
--- a/.github/workflows/create-release-branch.yaml
+++ b/.github/workflows/create-release-branch.yaml
@@ -38,7 +38,7 @@ jobs:
 
     - name: Bump latest tag variable version
       if: ${{ !github.event.inputs.version }}
-      run: echo "RELEASE_VERSION=$(echo ${{ env.CURRENT_VERSION }} | sed -r 's/([0-9]+)\.([0-9]+)\.([0-9]+)/echo \1.$((\2+1)).\3/' | sh)" >> $GITHUB_ENV
+      run: echo "RELEASE_VERSION=$(echo ${{ env.CURRENT_VERSION }} | sed -r 's/([0-9]+)\.([0-9]+)\.([0-9]+)/echo \1.$((\2+1)).0/' | sh)" >> $GITHUB_ENV
 
     - name: Verify version regex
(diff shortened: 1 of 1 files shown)


.github/workflows/create-release-branch.yaml
```diff
@@ -38,7 +38,7 @@ jobs:
 
     - name: Bump latest tag variable version
       if: ${{ !github.event.inputs.version }}
-      run: echo "RELEASE_VERSION=$(echo ${{ env.CURRENT_VERSION }} | sed -r 's/([0-9]+)\.([0-9]+)\.([0-9]+)/echo \1.$((\2+1)).\3/' | sh)" >> $GITHUB_ENV
+      run: echo "RELEASE_VERSION=$(echo ${{ env.CURRENT_VERSION }} | sed -r 's/([0-9]+)\.([0-9]+)\.([0-9]+)/echo \1.$((\2+1)).0/' | sh)" >> $GITHUB_ENV
```
tb06904, line 41: Just a thought a simple `awk` script might be a better choice here e.g. something like this should do the job
```bash
cho "RELEASE_VERSION=$(echo ${{ env.CURRENT_VERSION }} | awk -F. -v OFS=. '{$2 += 1 ; $3 = 0; print}')" >> $GITHUB_ENV
```

GCHQDeveloper314 (approved):

.github/workflows/create-release-branch.yaml
t92549: @tb06904 if this sounds good to you, I think for now I will leave this as it isn't crucial to solve the issue and this regex is present on the other repos. Worth maybe thinking about improving though when we come to refactor the release pipeline generally.

GCHQDeveloper314: I would echo this. There are at least as many instances of this `sed` command as there are Gaffer repos, probably a couple more. A change like this should be done as a refactor to keep consistency.

tb06904: That's fair doe we have a ticket for the refactor?

tb06904 (approved):

.github/workflows/create-release-branch.yaml
t92549: Not yet, we should probably make one though. I think the main points I would like to do are:
- removing the `ADMIN_GITHUB_TOKEN`
- removing release branches

PR #3126, opened 2023-11-30
Gh-2988: Add TinkerPop structure standard suite

### LINKED ISSUE #2988: Look into adding core tinkerpop tests
As described in https://github.com/gchq/Gaffer/issues/277#issuecomment-254504980, we should maybe hook the `GafferPopGraph` up to `StructureStandardSuite` and `ProcessStandardSuite` so the standard tests can be run on it. This may flag up issues with our TinkerPop implementation which need fixing. For example, we might claim to implement some [features](https://tinkerpop.apache.org/docs/current/reference/#_features) that we don't.

tb06904:
Basic integration of the standard tinkerpop structure test suite using the method recommended in the [official tinkerpop repo](https://github.com/apache/tinkerpop/blob/master/tinkergraph-gremlin/src/test/java/org/apache/tinkerpop/gremlin/tinkergraph/structure/TinkerGraphStructureStandardTest.java). There are a few bits for consideration which I'm open for suggestions on, but tried to implement a good base to take forward with.

Current results are:
`Tests run: 915, Failures: 95, Errors: 406, Skipped: 207`

**Main Changes**
Basic idea is that there is a GraphProvider class that configures the tinkerpop graph implementation to use, the main thing this means for Gaffer is ensuring correct schemas are loaded based on the expected input data. However, this lead to another required implementation as the standard tinkerpop structure expects an element ID manager to be available e.g. auto assign IDs to elements, so basic ID manager has been added here based on the one in the standard TinkerGraph implementation.

**Other Changes**
- Some of the exceptions our implementation were throwing were not correct based on a standard tinkerpop graph so they have been updated where applicable.
- Use the provided `toString` methods from the tinkerpop `StringFactory` where applicable.
- Support for loading types and element schemas from different directories to reduce duplication of types schemas.
- Basic maven profile for turning on the tests (off by default).

**Further Work**
Running the tests has highlighted some issues with the fundamental mapping of Gaffer to Tinkerpop, TODOs have been added to the relevant code areas for consideration. 
- First main issue is if Tinkerpop `labels` should map to Gaffer `groups` this seems logical and seems to happen in places but not all.
- Tinkerpop expects a traversal mechanism to be available allowing iterating over a graph this is not currently possible in Gaffer so many tests cannot run correctly.
- Tinkerpop is less strict with its elements and can expect elements within the same `label` to be added with different ID types which Gaffer currently can't handle due to its schemas requiring a fix mapping between `group` and `vertex` (id) type.
- The current implementation of read only on all Gaffer elements by default is a problem for Tinkerpop as it expects to be able to modify properties or elements in a Graph.

# Related issue

- Resolve #2988


library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/GafferPopEdge.java
```diff
@@ -46,23 +47,19 @@
...
 
-    public GafferPopEdge(final String label, final Object outVertexId, final Object inVertexId, final GafferPopGraph graph) {
-        this(label, checkVertex(outVertexId, graph), checkVertex(inVertexId, graph), graph);
-    }
-
-    public GafferPopEdge(final String label, final GafferPopVertex outVertex, final GafferPopVertex inVertex, final GafferPopGraph graph) {
-        super(label, Arrays.asList(outVertex.id(), inVertex.id()), graph);
-        this.outVertex = checkVertex(outVertex, graph);
-        this.inVertex = checkVertex(inVertex, graph);
+    public GafferPopEdge(final String label, final Object outVertex, final Object inVertex, final GafferPopGraph graph) {
```
GCHQDeveloper314, line 53: Why does this take `Object` instead of the relevant type?

library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/GafferPopEdge.java
```diff
@@ -127,29 +129,60 @@ public Vertex inVertex() {
...
+    /**
+     * Gets the vertex ID object from the supplied vertex.
+     *
+     * @param vertex The vertex Object or ID.
+     * @return The ID for the vertex.
+     */
+    private static Object getVertexId(final Object vertex) {
+        // Check if we need to pull the ID from the vertex or can use it directly
+        if ((vertex instanceof GafferPopVertex)
+                || (vertex instanceof Vertex)) {
```
GCHQDeveloper314, line 141:
```suggestion
        if (vertex instanceof Vertex) {
```

library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/GafferPopEdge.java
```diff
@@ -127,29 +129,60 @@ public Vertex inVertex() {
...
-            if (vertexId instanceof GafferPopVertex) {
-                gafferPopVertex = checkVertex(((GafferPopVertex) vertexId), graph);
-            } else {
-                gafferPopVertex = new GafferPopVertex(GafferPopGraph.ID_LABEL, ((Vertex) vertexId).id(), graph);
-            }
+    /**
+     * Gets the vertex ID object from the supplied vertex.
+     *
+     * @param vertex The vertex Object or ID.
+     * @return The ID for the vertex.
```
GCHQDeveloper314, line 136: This doesn't seem to reflect what the method does. It doesn't always return the id.

library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/GafferPopGraph.java
```diff
@@ -756,4 +834,27 @@ private GafferPopGraphVariables createVariables() {
         variablesMap.put(GafferPopGraphVariables.SCHEMA, graph.getSchema());
         return new GafferPopGraphVariables(variablesMap);
     }
+
+    /**
+     * Gets the next ID to assign to a supplied vertex based on the currently configured
+     * ID manager.
```
GCHQDeveloper314, line 840: This stuff doesn't seem to be very useful. It's required for the TinkerPop suite?

library/tinkerpop/pom.xml
```diff
@@ -27,6 +27,9 @@
 
     <properties>
         <tinkerpop.version>3.6.4</tinkerpop.version>
+
+        <!-- Don't run any of the standard tinkerpop tests by default-->
+        <test>!GafferPopGraphStructureStandardTest</test>
```
GCHQDeveloper314, line 32: It might be an option to run the tests but not fail on errors. If that can be done for a single test?

tb06904: The best could probably do is set the failure ignore property if the profile is activated [see](https://maven.apache.org/surefire/maven-surefire-plugin/test-mojo.html#testFailureIgnore)

library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/GafferPopGraph.java
```diff
@@ -630,28 +671,33 @@ private Iterator<GafferPopVertex> verticesWithSeedsAndView(final List<ElementSee
...
-                .then(new GenerateObjects.Builder<GafferPopVertex>()
-                        .generator(new GafferPopVertexGenerator(this))
+                .then(new GenerateObjects.Builder<GafferPopElement>()
+                        .generator(new GafferPopElementGenerator(this))
                         .build())
                 .build());
 
-        if (idVertices.isEmpty()) {
-            return new ChainedIterable<GafferPopVertex>(result, idVertices).iterator();
+        final Iterable<GafferPopVertex> resultVertexes = () -> StreamSupport.stream(result.spliterator(), false)
```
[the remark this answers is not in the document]
tb06904: I think for this one it does need filtering I was getting some failed tests as edges were being returned when this method should only give entities, think this is due to the view potentially allowing edges.

library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/GafferPopGraph.java
```diff
@@ -465,12 +500,18 @@ public Iterator<Edge> edges(final Object... edgeIds) {
...
                 .first(getOperation)
-                .then(new GenerateObjects.Builder<GafferPopEdge>()
-                        .generator(new GafferPopEdgeGenerator(this))
+                .then(new GenerateObjects.Builder<GafferPopElement>()
+                        .generator(new GafferPopElementGenerator(this))
                         .build())
-                .build()).iterator();
+                .build());
+
+        final Iterable<Edge> resultEdges = () -> StreamSupport.stream(result.spliterator(), false)
```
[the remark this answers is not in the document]
tb06904: True this probably isn't needed, it was left over from debugging which one of these methods were allowing edges and entities

library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/GafferPopGraph.java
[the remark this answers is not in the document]
t92549: Interesting, it looks like the view only specifies entities, so not sure why it would return edges

library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/GafferPopEdge.java
```diff
@@ -58,15 +62,17 @@ public GafferPopEdge(final String label, final Object outVertex, final Object in
 
     @Override
     public <V> Property<V> property(final String key, final V value) {
-        if (isReadOnly()) {
```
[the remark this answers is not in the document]
tb06904: True could probably move this readonly flag to the configuration so it can be toggled by the user

library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/GafferPopGraph.java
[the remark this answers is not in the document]
tb06904: Think you might be right actually I'm getting this one mixed up with the other one

tb06904:
After integrating cucumber testing and applying minor fixes new results from standard testing are:
`Tests run: 1734, Failures: 208, Errors: 328, Skipped: 276`

library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/GafferPopGraph.java
```diff
@@ -799,70 +790,44 @@ private View createAllEntitiesView() {
...
+        List<ElementSeed> seeds = new LinkedList<>();
+        if (null == ids) {
+            LOGGER.warn("Gaffer does not support unseeded queries, no results will be returned");
+            return seeds;
         }
 
-        return seeds;
-    }
-
-    private List<EntitySeed> getEntitySeeds(final Iterable<Object> vertexIds) {
```
t92549, line 702: What is the reason for moving these into one function? Isn't it technically more performant to have smaller functions? If you are in the `edges` function, all of your seeds should be edgeSeeds? Or is this assumption incorrect with tinkerpop?

tb06904: The assumption appears to be incorrect for tinkerpop as, the interface will request using the `edges()` method which tinkerpop assumes can also accept vertex ID objects as input so that edges on those given vertexes will be returned. Separating the functionality does not all for that use case.

library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/GafferPopEdge.java
GCHQDeveloper314: I see. Could you add a little to the Javadoc to make this clearer.

GCHQDeveloper314 (dismissed):

library/tinkerpop/pom.xml
```diff
@@ -115,4 +132,24 @@
...
         </plugins>
     </build>
+
+    <profiles>
+        <profile>
+            <id>tinkerpop-standard-test</id>
+            <properties>
+                <!-- Gaffer doesn't pass all standard gremlin tests, this will hopefully improve with time -->
+                <maven.test.failure.ignore>true</maven.test.failure.ignore>
+                <test>GafferPopGraphStructureStandardTest,GafferPopFeatureTest</test>
```
rj77259, line 142: Has a ticket been created to address these later on?

library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/GafferPopEdge.java
```diff
@@ -95,6 +97,32 @@ public <V> Iterator<Property<V>> properties(final String... propertyKeys) {
...
+     *
+     * @param <V> Value type
+     * @param key The key
+     * @param value The value
+     * @return The property
+     */
+    public <V> Property<V> propertyWithoutUpdate(final String key, final V value) {
+        ElementHelper.validateProperty(key, value);
+        final Property<V> newProperty = new GafferPopProperty<>(this, key, value);
+        if (null == this.properties) {
```
rj77259, line 118: Incorrect use of this, please remove.

library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/GafferPopEdge.java
```diff
@@ -112,9 +140,14 @@ public Set<String> keys() {
         return null == this.properties ? Collections.emptySet() : this.properties.keySet();
     }
 
+    @Override
+    public void remove() {
+        throw Edge.Exceptions.edgeRemovalNotSupported();
```
rj77259, line 145: Can we put a comment in to say why this is not supported.

library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/GafferPopProperty.java
```diff
@@ -69,6 +69,6 @@ public int hashCode() {
 
     @Override
     public void remove() {
-        throw new UnsupportedOperationException("Properties cannot be removed");
+        throw Property.Exceptions.propertyRemovalNotSupported();
```
rj77259, line 72: Can we add a comment in to say why this is not supported?

library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/GafferPopVertex.java
```diff
@@ -63,26 +69,14 @@ public <V> VertexProperty<V> property(final String key) {
...
+        // Attach the property to this vertex before updating and re adding to the graph
+        VertexProperty<V> vertexProperty = propertyWithoutUpdate(cardinality, key, value, keyValues);
+        LOGGER.warn("Updating Vertex properties via aggregation");
 
-        final List<VertexProperty> list = this.properties.getOrDefault(key, new ArrayList<>());
-        list.add(vertexProperty);
-        this.properties.put(key, list);
-        ElementHelper.attachProperties(vertexProperty, keyValues);
+        // Re add to do a update via aggregation
+        this.graph().addVertex(this);
```
rj77259, line 79: Misuse of 'this'

library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/GafferPopVertex.java
```diff
@@ -110,6 +104,45 @@ public <V> Iterator<VertexProperty<V>> properties(final String... propertyKeys)
...
+
+        final VertexProperty<V> vertexProperty = new GafferPopVertexProperty<>(this, key, value);
+
+        if (null == this.properties) {
+            this.properties = new HashMap<>();
+        }
+
+        final List<VertexProperty> list = this.properties.getOrDefault(key, new ArrayList<>());
+        list.add(vertexProperty);
+        this.properties.put(key, list);
```
rj77259, line 140: Misuse of this keyword

rj77259 (dismissed):

library/tinkerpop/pom.xml
tb06904: No yet, there's probably far too much for one ticket and I'm not sure we will ever be able to make them all pass due to how Gaffer fundamentally works.

library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/GafferPopVertex.java
tb06904: Technically not misuse more not required, the `graph()` method is in the parent class so might not have been obvious where it came from with out a `this`, I can remove though. I was mostly following the style the file already had where `this` was used to refer to anything that came from the parent class e.g. `properties` too.

GCHQDeveloper314 (approved):

t92549 (approved):

PR #3131, opened 2023-12-08
Gh-3120: Improved testing of matchedVertex

### LINKED ISSUE #3120: Many tests try to check matchedVertex incorrectly
**Describe the bug**
There are tests in Gaffer where an assertion is made that a results Edge is the same as an expected Edge, and the `matchedVertex` has an expected result, e.g.:
https://github.com/gchq/Gaffer/blob/1ef3621323064ab6a15c84fa554d35fe1dea003d/integration-test/src/test/java/uk/gov/gchq/gaffer/integration/impl/GeneratorsIT.java#L104-L110

However, they are just relying on the `Edge.equals` method which explicitly ignores `matchedVertex`:
https://github.com/gchq/Gaffer/blob/1ef3621323064ab6a15c84fa554d35fe1dea003d/core/data/src/main/java/uk/gov/gchq/gaffer/data/element/Edge.java#L322-L329

These tests should be fixed to properly check `matchedVertex` is as expected, e.g.:
https://github.com/gchq/Gaffer/blob/1ef3621323064ab6a15c84fa554d35fe1dea003d/core/data/src/test/java/uk/gov/gchq/gaffer/data/element/EdgeTest.java#L72

Or, perhaps a custom condition should be made, similar to:
https://github.com/gchq/Gaffer/blob/1ef3621323064ab6a15c84fa554d35fe1dea003d/store-implementation/accumulo-store/src/test/java/uk/gov/gchq/gaffer/accumulostore/integration/AccumuloMatchedVertexIT.java#L87-L88

t92549:
# Related issue

- Resolve #3120

### DIFF
--- a/core/data/src/test/java/uk/gov/gchq/gaffer/data/util/ElementUtil.java
+++ b/core/data/src/test/java/uk/gov/gchq/gaffer/data/util/ElementUtil.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016-2023 Crown Copyright
+ * Copyright 2016-2024 Crown Copyright
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/integration-test/src/test/java/uk/gov/gchq/gaffer/integration/impl/GetElementsIT.java
+++ b/integration-test/src/test/java/uk/gov/gchq/gaffer/integration/impl/GetElementsIT.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016-2023 Crown Copyright
+ * Copyright 2016-2024 Crown Copyright
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/store-implementation/accumulo-store/src/main/java/uk/gov/gchq/gaffer/accumulostore/retriever/impl/AccumuloElementsRetriever.java
+++ b/store-implementation/accumulo-store/src/main/java/uk/gov/gchq/gaffer/accumulostore/retriever/impl/AccumuloElementsRetriever.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016-2023 Crown Copyright
+ * Copyright 2016-2024 Crown Copyright
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/core/store/src/test/java/uk/gov/gchq/gaffer/store/operation/handler/function/FilterHandlerTest.java
+++ b/core/store/src/test/java/uk/gov/gchq/gaffer/store/operation/handler/function/FilterHandlerTest.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2017-2023 Crown Copyright
+ * Copyright 2017-2024 Crown Copyright
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/core/store/src/test/java/uk/gov/gchq/gaffer/store/operation/handler/function/TransformHandlerTest.java
+++ b/core/store/src/test/java/uk/gov/gchq/gaffer/store/operation/handler/function/TransformHandlerTest.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2017-2023 Crown Copyright
+ * Copyright 2017-2024 Crown Copyright
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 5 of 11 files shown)


integration-test/src/test/java/uk/gov/gchq/gaffer/integration/impl/GetElementsIT.java
```diff
@@ -652,8 +652,13 @@ private void shouldGetElements(final Collection<Element> expectedElements,
         final Iterable<? extends Element> resultsElement = graph.execute(opElement, user);
 
         // Then
-        ElementUtil.assertElementEquals(expectedElements, resultsSeed, true);
-        ElementUtil.assertElementEquals(expectedElements, resultsElement, true);
+        if (includeEdges && inOutType == IncludeIncomingOutgoingType.INCOMING) {
```
t92549, line 655: In other tests, I managed to fix the assertions so they correctly test the MatchedVertex behaviour. However, GetElementsIT is quite complex and the MatchedVertex has been tested incorrectly for a while so this might require a new issue or larger refactor of the test.

core/data/src/test/java/uk/gov/gchq/gaffer/data/util/ElementUtil.java
```diff
@@ -43,4 +43,22 @@ public static void assertElementEquals(final Iterable<? extends ElementId> expec
...
+        final List<ElementId> expectedCache = Lists.newArrayList(expected);
+        final List<ElementId> resultCache = Lists.newArrayList(result);
+        if (ignoreDuplicates) {
+            assertThat(resultCache)
+                .usingRecursiveFieldByFieldElementComparator()
+                .hasSameElementsAs(expectedCache);
+        } else {
+            assertThat(resultCache)
+                .usingRecursiveFieldByFieldElementComparator()
+                .containsExactlyInAnyOrderElementsOf(expectedCache);
```
GCHQDeveloper314, line 61: The same applies to the code further up as well.
```suggestion
        if (ignoreDuplicates) {
            assertThat(result)
                .asList()
                .usingRecursiveFieldByFieldElementComparator()
                .hasSameElementsAs(expected);
        } else {
            assertThat(result)
                .asList()
                .usingRecursiveFieldByFieldElementComparator()
                .containsExactlyInAnyOrderElementsOf(expected);
```

store-implementation/accumulo-store/src/main/java/uk/gov/gchq/gaffer/accumulostore/retriever/impl/AccumuloElementsRetriever.java
```diff
@@ -31,8 +31,8 @@ public AccumuloElementsRetriever(final AccumuloStore store,
                                      final User user)
             throws IteratorSettingException, StoreException {
         super(store, operation, user,
-                // includeMatchedVertex if input only contains EntityIds
-                StreamSupport.stream(operation.getInput().spliterator(), false).noneMatch(input -> EdgeId.class.isInstance(input)),
+                // includeMatchedVertex if any input are EntityIds
+                StreamSupport.stream(operation.getInput().spliterator(), false).anyMatch(EntityId.class::isInstance),
```
GCHQDeveloper314, line 35: Is this related to the issue or just a general improvment?

integration-test/src/test/java/uk/gov/gchq/gaffer/integration/impl/GetElementsIT.java
GCHQDeveloper314: Was there an issue created for this?

store-implementation/accumulo-store/src/main/java/uk/gov/gchq/gaffer/accumulostore/retriever/impl/AccumuloElementsRetriever.java
t92549: Due to adding proper testing for matchedVertex, I realised that this more accurately recreates the Gaffer 1 behaviour.

core/data/src/test/java/uk/gov/gchq/gaffer/data/util/ElementUtil.java
t92549: I tried this but doesn't work in [MapStore tests](https://github.com/gchq/Gaffer/actions/runs/7248721759/job/19745194192?pr=3131)

integration-test/src/test/java/uk/gov/gchq/gaffer/integration/impl/GetElementsIT.java
t92549: Issue created: https://github.com/gchq/Gaffer/issues/3138

core/data/src/test/java/uk/gov/gchq/gaffer/data/util/ElementUtil.java
GCHQDeveloper314: It seems use of the Guava `Lists` library is only to perform a cast without showing any warning. Using `asList` doesn't work because it just casts to a `List` which will fail if the underlying `Iterable` wasn't one already.

The assertion can be used directly by simply casting to remove the `?` wildcard `assertThat((Iterable<ElementId>) result)`. This doesn't fail for those Map Store tests.

GCHQDeveloper314 (approved):

Issue #3125, opened 2023-11-30
GafferPopGraph execute method doesn't catch certain exceptions


GCHQDeveloper314:
**Describe the bug**
When executing an Operation which is not supported, a `java.lang.UnsupportedOperationException` is thrown. This class is unrelated to the `uk.gov.gchq.gaffer.operation.OperationException` class which is caught by the `GafferPopGraph` catch clause.

These are the relevant lines:

https://github.com/gchq/Gaffer/blob/7b5e9ffa95602fb0ab4ae7b04610fd2e4c093d6f/library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/GafferPopGraph.java#L518-L523

This means that `UnsupportedOperationException` is not caught and handled as intended. The handling code could also be improved (e.g. log the error instead of just throwing).

**To Reproduce**
Steps to reproduce the behaviour:
1. Add breakpoint in the above catch.
2. Execute an Operation which is not supported by the configured store.
3. Catch is not invoked as it should be.

**Expected behaviour**
All kinds of Exceptions encountered when executing Operation(s) should be handled.

**Additional context**
The snippet above is not tested. Tests should be added for this to ensure correct behaviour.

t92549:
Generally, the testing of `Store.doUnhandledOperation` isn't great

PR #3133, opened 2023-12-11
Gh-3125: GafferPopGraph should catch certain exceptions

cn337131:
Changed the try/catch to catch all exceptions and added a test that it catches any invalid operation chains.

### DIFF
--- a/library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/GafferPopGraph.java
+++ b/library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/GafferPopGraph.java
@@ -35,8 +35,8 @@ import uk.gov.gchq.gaffer.graph.Graph;
 import uk.gov.gchq.gaffer.graph.GraphConfig;
 import uk.gov.gchq.gaffer.operation.Operation;
 import uk.gov.gchq.gaffer.operation.OperationChain;
-import uk.gov.gchq.gaffer.operation.OperationChain.Builder;
 import uk.gov.gchq.gaffer.operation.OperationException;
+import uk.gov.gchq.gaffer.operation.OperationChain.Builder;
 import uk.gov.gchq.gaffer.operation.data.EdgeSeed;
 import uk.gov.gchq.gaffer.operation.data.ElementSeed;
 import uk.gov.gchq.gaffer.operation.data.EntitySeed;
@@ -518,8 +518,14 @@ public class GafferPopGraph implements org.apache.tinkerpop.gremlin.structure.Gr
         try {
             LOGGER.info("GafferPop operation chain called: " + opChain.toString());
             return graph.execute(opChain, user);
-        } catch (final OperationException e) {
-            throw new RuntimeException(e);
+        } catch (final UnsupportedOperationException | OperationException | IllegalArgumentException e) {
+            // Catch the common exceptions
+            LOGGER.error("Failed to execute operation chain: " + opChain.toString() + " due to " + e.getMessage(), e);
+            throw new UnsupportedOperationException(e);
+        } catch (final Exception e) {
+            // Catch any leftover exceptions
+            LOGGER.error("Failed to execute operation chain: " + opChain.toString() + " due to " + e.getMessage(), e);
+            throw new UnsupportedOperationException(e);
         }
     }
 
--- a/library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/GafferPopGraph.java
+++ b/library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/GafferPopGraph.java
@@ -519,7 +519,7 @@ public class GafferPopGraph implements org.apache.tinkerpop.gremlin.structure.Gr
             return graph.execute(opChain, user);
         } catch (final Exception e) {
             LOGGER.error("Operation chain failed: " + e.getMessage(), e);
-            throw new RuntimeException(e);
+            throw new RuntimeException("Failed to execute GafferPop operation chain", e);
         }
     }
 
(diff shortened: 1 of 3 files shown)


library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/GafferPopGraph.java
```diff
@@ -518,7 +517,8 @@ public <T> T execute(final OperationChain<T> opChain) {
         try {
             LOGGER.info("GafferPop operation chain called: " + opChain.toString());
             return graph.execute(opChain, user);
-        } catch (final OperationException e) {
+        } catch (final Exception e) {
+            LOGGER.error("Operation chain failed: " + e.getMessage(), e);
             throw new RuntimeException(e);
```
GCHQDeveloper314, line 522: We can add a message to this type of exception.
```suggestion
            throw new RuntimeException("Failed to execute GafferPop operation chain", e);
```

library/tinkerpop/src/test/java/uk/gov/gchq/gaffer/tinkerpop/GafferPopGraphIT.java
```diff
@@ -501,7 +504,23 @@ public void shouldThrowExceptionIfGetAdjacentVerticesWithNoSeeds() {
...
+        final Graph gafferGraph = getGafferGraph();
+        final GafferPopGraph graph = GafferPopGraph.open(TEST_CONFIGURATION_1, gafferGraph);
+        final OperationChain<?> invalidOperationChain = new OperationChain.Builder()
+                .first(new AddElementsFromSocket())
+                .then(new GetElements())
+                .build();
+
+        // When / Then
+        assertThatExceptionOfType(RuntimeException.class)
+            .isThrownBy(() -> graph.execute(invalidOperationChain));
```
GCHQDeveloper314, line 523: We can check for our custom message by adding a `hasMessage` assertion.

GCHQDeveloper314 (commented):
Just one minor additon.

GCHQDeveloper314 (approved):

library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/GafferPopGraph.java
```diff
@@ -518,8 +517,9 @@ public <T> T execute(final OperationChain<T> opChain) {
         try {
             LOGGER.info("GafferPop operation chain called: " + opChain.toString());
             return graph.execute(opChain, user);
-        } catch (final OperationException e) {
-            throw new RuntimeException(e);
+        } catch (final Exception e) {
```
rj77259, line 520: I think we should be catching all the specific exceptions and then have the exception catch the others.

rj77259 (changes requested):

library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/GafferPopGraph.java
GCHQDeveloper314: This is the usual practice for Java but not how things have been done in Gaffer. If we want to change to using this approach it would be better to raise as a separate issue.

There's a need to investigate the kinds of exceptions thrown by Operations, ensure they are all children of `OperationException` and enforce this. This is why `Exception` has been directly caught elsewhere - it's not possible to know what specific kinds of exceptions will be thrown and handle them in the usual way.

rj77259: I think if we follow the boy scout rule for this and for other issues then that would be better. As we are in this area of the code lets improve it. I think for now we can catch the specific exceptions that we know about that are going to be thrown and let the exception catch the rest. Ideally, we would catch only the specific ones we know about. There maybe other exceptions that get thrown that we are not aware of but we can't catch everything. Also, on another point it might be worthwhile understanding the root cause of this issue not the symptom. We are catching exceptions here but what is the functionality deeper down. Do we need to address anything in the execute method? Is this expected functionality? Food for thought.

GCHQDeveloper314: The cause of the problem here is the original code not handling the unchecked exception `UnsupportedOperationException` which can be thrown when executing an Operation. This is part of a wider problem of unchecked exception being used instead of checked exceptions.

_In my opinion, this exception shouldn't be thrown or used at all._ As [the official Java guide says,](https://docs.oracle.com/javase/tutorial/essential/exceptions/runtime.html) unchecked exceptions are for when "code cannot reasonably be expected to recover". In Gaffer we always intend to recover from a situation where an Operation is not supported. As such, this should manifest itself with a checked exception. This can be a child of Gaffer's existing `OperationException`. However, this would be a fairly big change and likely API breaking.

If you think this is a good idea to reduce technical debt, then I can raise this as a separate issue to investigate.

If this improvement were made, then we wouldn't need to catch `Exception` at all or catch anything beyond what is thrown by execute - i.e. there would be no problem and the issue this PR is for wouldn't have existed.

t92549 (approved):

tb06904 (approved):

library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/GafferPopGraph.java
rj77259: Agreed I think we need to revisit this then and to sort out the exceptions that shouldn't be thrown in this way.

rj77259 (approved):
This is a temporary fix and we need to create a series of tickets to address the issues around exception handling and throwing.

Issue #3129, opened 2023-12-04
FileGraphLibrary no longer works with FederatedStore


t92549:
**Describe the bug**
You can no longer use a FileGraphLibrary with a FederatedStore. If you add a graph and restart the store, it will crash.

This may be due to changes with how getSchema works. When the FederatedStore is started, a file in the library is made with its empty schema. When a graph is added, the getSchema changes now mean that it will return the merged schema. When the FederatedStore is restarted, this schema now conflicts with the empty schema file, causing a crash.

**To Reproduce**
Steps to reproduce the behaviour:
1. Create a FederatedStore that uses FileGraphLibrary and a persistent cache
2. Add a graph to it
3. Turn off the FederatedStore
4. Turn it back on again, it will crash on startup

**Expected behaviour**
This behaviour should be possible and worked in Gaffer 1.

**Additional context**
I think this problem not being caught indicates we need better tests for turning the FederatedStore on and off, and how this interacts with persistent caches and the FileGraphLibrary.

Also, why isn't there a way to use the cache for the graph library? A file does not seem like the best way top store the graph library.

GCHQDeveloper314:
I've had a look at this and can reproduce it with a simple test within Gaffer which attempts to create a second store instead of restarting. Debugging reveals this is indeed caused by the getSchema changes.

When creating a new `Graph` with the builder, there's [a call on line 841](https://github.com/gchq/Gaffer/blob/b000f72defda0861e86cb70ab89aaaece98711d0/core/graph/src/main/java/uk/gov/gchq/gaffer/graph/Graph.java#L841) which leads to line 997; where if the schema is null or empty it is replaced with a schema fetched from the store with `getSchema()`. _This is the cause of the problem in this issue, because for the Federated Store the schema returned is no longer an empty schema._ The builder then calls the `GraphLibrary` `checkExisting` method which compares the blank schema from the `FileGraphLibrary` file with the schema returned by the Federated Store (which is not blank), this causes the below exception:

```
uk.gov.gchq.gaffer.commonutil.exception.OverwritingException: GraphId federatedGraph already exists with a different schema:
existing schema:
{"types":{}}
new schema:
{....
```

This could be fixed in various ways.

[...]

PR #3140, opened 2023-12-20
Gh-3129: FileGraphLibrary empty Schema fix

GCHQDeveloper314:


GCHQDeveloper314:
In addition to these changes, I also plan to add a test which reproduces the problem identified in the issue. This won't improve coverage but will help ensure future changes don't reintroduce a problem.

core/graph/src/test/java/uk/gov/gchq/gaffer/graph/GraphTest.java
```diff
@@ -1092,6 +1050,23 @@ public void shouldThrowExceptionIfStoreClassPropertyIsNotSet() throws OperationE
         }
     }
 
+    @Test
+    public void shouldThrowExceptionIfGraphIdIsMissing() {
+        final StoreProperties storeProperties = new StoreProperties();
+        storeProperties.setStoreClass(TestStoreImpl.class.getName());
+        try {
```
tb06904, line 1057: Using assertJ would be better here for checking exception is called, as you should only be checking the `build()` method. Doing something like the following would be much better:

```java
Graph.Builder initNoGraphId = new Graph.Builder()
     .config(new GraphConfig.Builder()
         .build())
     .addSchema(new Schema())
     .storeProperties(new StoreProperties());

 assertThatExceptionOfType(IllegalArgumentException.class)
     .isThrownBy(initNoGraphId::build)
     .withMessageContaining("graphId is required");
```

core/graph/src/main/java/uk/gov/gchq/gaffer/graph/Graph.java
```diff
@@ -988,13 +985,13 @@ private void initStore(final GraphConfig config) {
...
 
-            // Use the store's graph Id if we don't have on configured already
+            // Use the store's graph Id if we don't have one configured already
             if (config.getGraphId() == null) {
                 config.setGraphId(store.getGraphId());
             }
 
             // Use the store's schema if not already set
-            if (schema == null || schema.getGroups().isEmpty()) {
+            if (schema == null) {
```
t92549, line 994: Interesting that your fix removes checks, will this not remove crashes with incorrect config?
I was thinking `if (getSchema() == null || schema.getGroups().isEmpty())` might work, did you try that?

GCHQDeveloper314: > will this not remove crashes with incorrect config?

It shouldn't cause a problem because other checks are already made if the schema is non-null. These will already reject an empty schema is that is not valid for the store type. So I think this line would only have been reached when using a schemaless store.

For your other comment, [see my response in the issue](https://github.com/gchq/Gaffer/issues/3129#issuecomment-1863071145).

t92549 (approved):

tb06904 (approved):

rj77259 (approved):

PR #3144, opened 2024-01-05
Gh-2988: Add tinkerpop process standard suite

### LINKED ISSUE #2988: Look into adding core tinkerpop tests

l46978:
Initial integration of the standard TinkerPop process test suite.

Current results are:
`Tests run: 229, Failures: 69, Errors: 83, Skipped: 9`

Further work to be done to improve the broken tests.

# Related issue

- Resolve #2988


library/tinkerpop/src/test/resources/gaffer/map-store.properties
GCHQDeveloper314, line 1: What was the reason for changing this file? Removing the cache services prevents named operation tests from working. This needs to be reverted to fix those tests.

library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/GafferPopGraph.java
```diff
@@ -311,6 +330,19 @@ public void addVertex(final GafferPopVertex vertex) {
...
+                .build())
+        .build();
+
+        final Iterable<? extends GafferPopElement> result = execute(new Builder()
+                .first(getOperation)
+                .then(new GenerateObjects.Builder<GafferPopElement>()
+                        .generator(new GafferPopElementGenerator(this))
+                        .build())
+                .build());
+
```
GCHQDeveloper314, line 345: Are these used? It doesn't look like it based on the other change. I reckon the annotations are the only changes needed in this file.

GCHQDeveloper314 (changes requested):
Tests are failing due to the `map-store.properties` file change.

library/tinkerpop/src/test/resources/gaffer/map-store.properties
GCHQDeveloper314: If these changes are required for the tinkerpop suite, then hopefully that can use a different file instead.

library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/GafferPopGraph.java
l46978: Must have missed this, will remove. Thank you

GCHQDeveloper314 (approved):

PR #3146, opened 2024-01-16
Gh-3095: Log JobTracker errors and still return results

### LINKED ISSUE #3095: Results should still be returned if JobTracker errors
If the job tracker is enabled but the upload to it fails, the results from a Gaffer query will also fail and not be returned. This should not be the case, instead an error should be logged and the results still returned.

If a JobTracker fails, users still want results.

This fail or log could be set via a config option.

l46978:
- Logged error instead of throwing a CacheOperationException within the JobTracker addOrUpdateJob method.
- Adding in testing for the changes within the method.
- Adding in testing within the Store Test to check operations are now able to execute with a broken JobTracker cache.

# Related issue

- Resolve #3095


core/operation/src/test/java/uk/gov/gchq/gaffer/jobtracker/BrokenICache.java
```diff
@@ -0,0 +1,61 @@
...
+
+    @Override
+    public Set getAllKeys() {
+        return emptySet();
+    }
+
+    @Override
+    public void clear() throws CacheOperationException {
+        // TODO Auto-generated method stub
+        // throw new UnsupportedOperationException("Unimplemented method 'clear'");
```
GCHQDeveloper314, line 58: Can be removed

core/operation/src/test/java/uk/gov/gchq/gaffer/jobtracker/BrokenCacheService.java
```diff
@@ -0,0 +1,44 @@
...
+import uk.gov.gchq.gaffer.cache.ICacheService;
+
+import java.util.Properties;
+
+public class BrokenCacheService implements ICacheService {
+
+    @Override
+    public void initialise(Properties properties) {
+        // TODO Auto-generated method stub
+        // throw new UnsupportedOperationException("Unimplemented method 'initialise'");
```
GCHQDeveloper314, line 29: Can be removed

core/operation/src/test/java/uk/gov/gchq/gaffer/jobtracker/BrokenCacheService.java
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    @Override
+    public void shutdown() {
+        // TODO Auto-generated method stub
+        // throw new UnsupportedOperationException("Unimplemented method 'shutdown'");
```
GCHQDeveloper314, line 35: Can be removed

core/operation/src/test/java/uk/gov/gchq/gaffer/jobtracker/BrokenCacheService.java
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    @Override
+    public <K, V> ICache<K, V> getCache(String cacheName) {
+        // TODO Auto-generated method stub
```
GCHQDeveloper314, line 40: Can be removed

core/operation/src/test/java/uk/gov/gchq/gaffer/jobtracker/JobTrackerBrokenCacheTest.java
```diff
@@ -0,0 +1,68 @@
...
+    private Logger loggerMock;
+
+    private static final String CACHE_SERVICE_CLASS_STRING = "uk.gov.gchq.gaffer.jobtracker.BrokenCacheService";
+    private static Properties properties = new Properties();
+
+    @BeforeAll
+    public static void setUp() {
+        properties.setProperty(CacheProperties.CACHE_SERVICE_CLASS, CACHE_SERVICE_CLASS_STRING);
+        CacheServiceLoader.initialise(properties);
+
```
GCHQDeveloper314, line 46: Empty line

core/operation/src/test/java/uk/gov/gchq/gaffer/jobtracker/JobTrackerBrokenCacheTest.java
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    @Test
+    public void jobTrackerDoesNotThrowExceptionWhenCacheIsBroken() {
+        // Given
+        JobTracker jobTracker = new JobTracker("Test");
+        final OperationChain operationChain = new OperationChain.Builder().first(new GetAllElements()).build();
+        final User user = new User("user");
```
GCHQDeveloper314, line 54: This doesn't seem to be used although it could be.

core/store/src/test/java/uk/gov/gchq/gaffer/store/BrokenCacheService.java
GCHQDeveloper314, line 1: This appears to be a duplicate of the class in `operation`.

core/store/src/test/java/uk/gov/gchq/gaffer/store/BrokenICache.java
GCHQDeveloper314, line 1: This appears to be a duplicate of the class in `operation`.

GCHQDeveloper314 (dismissed):
Introducing the 'broken' test classes is not ideal. I think mocking can be used instead, although it won't be as simple to setup. The duplication of these classes should be removed.

core/store/src/test/java/uk/gov/gchq/gaffer/store/StoreTest.java
```diff
@@ -246,6 +246,24 @@ public void after() {
         JSONSerialiser.update();
     }
 
+    @Test
+    public void shouldExecuteOperationWhenJobTrackerCacheIsBroken(@Mock final StoreProperties storeProperties) throws Exception {
```
GCHQDeveloper314, line 250: This test doesn't work because `StoreImpl` overrides the `createJobTracker()` method such that it doesn't create a JobTracker at all. Also, JobTracker is disabled by default and needs to be enabled but this can't be done because mocked store properties are used.

l46978: Added a storeimpl with an enabled JobTracker which isnt mocked.

GCHQDeveloper314 (approved):

core/operation/src/test/java/uk/gov/gchq/gaffer/jobtracker/JobTrackerBrokenCacheTest.java
```diff
@@ -0,0 +1,71 @@
...
+
+public class JobTrackerBrokenCacheTest {
+    @BeforeAll
+    public static void setUp() throws Exception {
+        ICache<Object, Object> mockICache = Mockito.mock(ICache.class);
+        doThrow(new CacheOperationException("Stubbed class")).when(mockICache).put(any(), any());
+        ICacheService mockICacheService = Mockito.spy(ICacheService.class);
+        given(mockICacheService.getCache(any())).willReturn(mockICache);
+
+        Field field = CacheServiceLoader.class.getDeclaredField("service");
```
tb06904, line 47: As an alternative suggestion would you instead be able add an overload to the `CacheServiceLoader`'s `initialise()` method so that accepts an existing `ICacheService` implementation object as a param? e.g.

```java
public static void initialise(final ICacheService service, final Properties properties) {
   ...
}
```

tb06904 (approved):

core/operation/src/test/java/uk/gov/gchq/gaffer/jobtracker/JobTrackerBrokenCacheTest.java
GCHQDeveloper314: I've made various changes to `CacheServiceLoader` in #3147, this could be added there. Although in general I'm not a fan of classes which have a bunch of methods that are only used for tests.

PR #3147, opened 2024-01-17
Gh-3094: Allow services to use different cache implementations

### LINKED ISSUE #3094: Allow different services to use different cache implementations
Currently, in Gaffer's store properties you define what cache implementation you want to use. This is then used for all aspects of Gaffer that use a cache, namely the FederatedStore, NamedOperations and the JobTracker. It would be better if users could specify a different cache implementation for each of these services. 

The different services each have different size and frequency of calls. This means for optimal storage, different implementations should be used.

**Additional context**
Currently, the cache implementation is defined like so:
```
gaffer.cache.service.class=uk.gov.gchq.gaffer.cache.impl.JcsCacheService
```
This should still be possible to not make a breaking change, but should be overridable, like this for example:
```
gaffer.cache.service.class=uk.gov.gchq.gaffer.cache.impl.JcsCacheService
gaffer.jobtracker.cache.service.class=uk.gov.gchq.gaffer.cache.impl.HashMapCache
```
The `CacheServiceLoader` would probably need changing:
https://github.com/gchq/Gaffer/blob/a9d3347928f0d81efead60d23002967013304de7/core/cache/src/main/java/uk/gov/gchq/gaffer/cache/CacheServiceLoader.java#L66

GCHQDeveloper314:
# Related issue

- Resolve #3094


GCHQDeveloper314:
Further work is required to add testing for the new multi-cache implementation and to remove as much use of deprecated code as possible, while also adding tests to ensure external code using deprecated methods can still work to some degree.

core/store/src/main/java/uk/gov/gchq/gaffer/store/Store.java
```diff
@@ -1101,6 +1108,25 @@ private void addConfiguredOperationHandlers() {
...
+        final String namedViewCacheClass = properties.getNamedViewCacheServiceClass();
+        final String namedOperationCacheClass = properties.getNamedOperationCacheServiceClass();
+        final String defaultCacheClass = properties.getDefaultCacheServiceClass();
+        final boolean defaultServiceRequired = !((jobTrackerCacheClass != null) && (namedViewCacheClass != null) && (namedOperationCacheClass != null));
+
+        if (jobTrackerCacheClass != null) {
+            CacheServiceLoader.initialise(JOB_TRACKER_CACHE_SERVICE_NAME, jobTrackerCacheClass, properties.getProperties());
+        }
+        if (namedViewCacheClass != null) {
+            CacheServiceLoader.initialise(NAMED_VIEW_CACHE_SERVICE_NAME, jobTrackerCacheClass, properties.getProperties());
```
cn337131, line 1121:
```suggestion
            CacheServiceLoader.initialise(NAMED_VIEW_CACHE_SERVICE_NAME, jobTrackerCacheClass, properties.getProperties());
```

Not sure this is meant to be 'jobTrackerCacheClass'? If it should be namedViewCacheClass, then the issue is repeated in the if statement after this as well.

GCHQDeveloper314: Good catch, now fixed

core/cache/src/test/java/uk/gov/gchq/gaffer/cache/CacheTest.java
```diff
@@ -76,8 +68,8 @@ public void shouldAddAndGetCacheNoOverwrite() throws CacheOperationException {
     }
 
     @Test
-    public void shouldGetCacheServiceName() {
-        assertEquals("serviceName1", cache.getCacheName());
+    public void shouldGetCacheName() {
+        assertEquals("cacheName1", cache.getCacheName());
```
cn337131, line 72: Elsewhere you have updated tests from junit to assertJ but you haven't done so in this file. Since you have changed these tests it might make sense to update them for consistency.

GCHQDeveloper314: I've now done this

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/FederatedGraphStorage.java
```diff
@@ -201,25 +194,23 @@ private boolean remove(final String graphId, final Predicate<FederatedAccess> ac
...
-                    }
+        try {
+            final GraphSerialisable graphFromCache = federatedStoreCache.getGraphFromCache(graphId);
+            final StoreProperties storeProperties = graphFromCache.getStoreProperties();
+            final List<Cache<?, ?>> caches = Lists.newArrayList(
+                    new NamedViewCache(storeProperties.getCacheServiceNamedViewSuffix(graphId)),
+                    new NamedOperationCache(storeProperties.getCacheServiceNamedOperationSuffix(graphId)),
+                    new JobTracker(storeProperties.getCacheServiceJobTrackerSuffix(graphId)));
+            for (final Cache<?, ?> cacheInstance : caches) {
+                final ICache<?, ?> cache = cacheInstance.getCache();
```
tb06904, line 205: Doesnt the `Cache` class have a `clearCache` method already? Why does it need to be used to then get the `ICache` first?

store-implementation/federated-store/src/test/java/uk/gov/gchq/gaffer/federatedstore/FederatedStorePropertiesTest.java
```diff
@@ -0,0 +1,113 @@
...
+    final String value = FederatedStore.class.getName();
+    static Path propertiesFilePath;
+
+    @BeforeAll
+    public static void init() throws URISyntaxException {
+        propertiesFilePath = Paths.get(FederatedStorePropertiesTest.class.getResource("/properties/federatedStore.properties").toURI());
+    }
+
+    @Test
+    public void shouldGetIsPublicAccessAllowed() {
```
tb06904, line 43: Just a tip, don't need `public` modifiers on test methods or classes in Junit

GCHQDeveloper314: I tend to keep consistent with this for existing classes. But as this is a new class, I'll fix it to not have it. I think it's there as I copy pasted the existing StoreProperties unit test to use as a template...

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/FederatedGraphStorage.java
GCHQDeveloper314: I think you're right. I'd refactored the existing logic which did this, but it does seem that you could clear the cache directly without these extra steps. I'll look at making this change.

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/FederatedGraphStorage.java
```diff
@@ -201,25 +194,23 @@ private boolean remove(final String graphId, final Predicate<FederatedAccess> ac
...
-                    if (nonNull(cache)) {
-                        cache.clear();
-                    } else {
-                        LOGGER.debug(String.format("No cache found graphId:%s with cacheName:%s", graphId, cacheName));
-                    }
+        try {
+            final GraphSerialisable graphFromCache = federatedStoreCache.getGraphFromCache(graphId);
+            final StoreProperties storeProperties = graphFromCache.getStoreProperties();
+            final List<Cache<?, ?>> caches = Lists.newArrayList(
+                    new NamedViewCache(storeProperties.getCacheServiceNamedViewSuffix(graphId)),
```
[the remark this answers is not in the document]
GCHQDeveloper314: Thanks for spotting this, we can use that directly. I thought I could make it neater by creating the objects but there's no advantage given your other comment on clearing the cache.
As written, this code will fail if the default service name wasn't used when initialising the cache. I will try and replicate this is tests and then refactor this code.

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/FederatedGraphStorage.java
GCHQDeveloper314: Turns out that this wouldn't give much of an advantage as we need to do the null check and `clearCache` doesn't allow for that and would throw and exception. Could do something like this but I don't think it's better:
```
            for (final Cache<?, ?> cacheInstance : caches) {
                if (nonNull(cacheInstance.getCache())) {
                    cacheInstance.clearCache();
                } else {
                    LOGGER.debug(String.format("No cache found graphId:%s with cacheName:%s", graphId, cacheInstance.getCacheName()));
                }
            }
```

tb06904: I don't think the other method would require a null check as if you follow the logic through it will ultimately use the `ICacheService` implementation for the cache instance to do the clearing which already handles the `getCache()` call.

GCHQDeveloper314: I believe the reason for not using `clearCache` in `Cache` is as follows.

For this the `ICacheService` object is fetched and the `clearCache` method for that is called. This method then proceeds to fetch the `ICache` object (using implementation specifc `getCache`) and immediately call its `clear` method. A problem could occur where the `ICacheService` `getCache` method returns a `null` instance of an `ICache` implementation which _might_ then throw an exception when its `clear` method is called.

Fetching the `ICache` itself and checking it is not null avoids this problem. This exception is only an issue because it seems we want to log when no cache is found, rather than throwing an error.

Further changes could be made, e.g. move the null check to `ICacheService`, but I think the current setup is ok for now.

tb06904 (approved):

cn337131 (approved):

PR #3148, opened 2024-01-24
Gh-3096: JobTracker Performance Improvement

### LINKED ISSUE #3096: JobTracker performance should be improved
Currently, the performance of the JobTracker is not great, it should be improved to scale better.

When a Gaffer rest-api is started, `Cache::getAllKeys` is called. This is a big performance hit which could be very bad if the cache is large. This should not happen, or at least be something that can be disabled. In fact, with some Cache implementations and sizes, users might want to ensure a `getAllKeys` is never called, perhaps this should also be something that can be disabled. Perhaps it could be configured to have a limit like `GetAllElements`.

Additionally, calls to add to the JobTracker (and probably other caches) are synchronous, results would come back faster in most cases is this were asynchronous. Often, when things are added to the cache they are immediately returned as well, this is also a performance problem and not always needed.

tb06904:
Various improvements to the Job Tracker (and subsequently the cache) to hopefully address some of the performance issues. There may be other areas but the main changes and rational behind them are:

- Move the Cache interface to be more generic `Iterator` instead of `Set` this should allow Cache implementations to do lazy gets so process wont block on loading everything into memory first.
- Configurable Job rescheduling on start up, this allows users to control whether they want to reschedule all jobs in the Cache on start up or not as this can have a large impact on start up times (its off by default).
- Adding to the JobTracker cache is now async to not block the main operation, gets are still synced to prevent concurrent modifications.
- Minor fix to add short circuit logic to the `NamedOperationCache` so that it doesn't always do a get all if overwrite is disabled.
- Few minor bits of refactoring to help maintenance.

Once performance tests are in place they should verify if improvements have made substantial difference. 

# Related issue

- Resolve #3096


core/operation/src/main/java/uk/gov/gchq/gaffer/jobtracker/JobTracker.java
```diff
@@ -33,6 +34,8 @@
 public class JobTracker extends Cache<String, JobDetail> {
 
     private static final String CACHE_SERVICE_NAME_PREFIX = "JobTracker";
+    // Executor to queue to allow queuing up async operations on the cache
+    private final ThreadPoolExecutor executor = new ThreadPoolExecutor(1, 1, 1, TimeUnit.MINUTES, new LinkedBlockingQueue<>());
```
GCHQDeveloper314, line 38: I assume this was done to improve performance, but there's nothing to prevent cache race conditions (which would be hard to test for). As such I'd suggest raising this as a separate ticket for a future improvement.
The PR description states only gets are async?

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/FederatedStoreCache.java
```diff
@@ -120,7 +116,7 @@ public void addGraphToCache(final GraphSerialisable graphSerialisable, final byt
     public FederatedAccess getAccessFromCache(final String graphId) {
         try {
             final byte[] accessFromCache = cacheTransient.getAccessFromCache(graphId);
-            return (isNull(accessFromCache)) ? null : JSONSerialiser.deserialise(accessFromCache, FederatedAccess.class);
+            return (Objects.isNull(accessFromCache)) ? null : JSONSerialiser.deserialise(accessFromCache, FederatedAccess.class);
```
GCHQDeveloper314, line 119: It was already here and as I believe that this use of `Objects.isNull` is poor practice, `== null` should be used instead.

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/FederatedStoreCacheTransient.java
```diff
@@ -98,7 +92,7 @@ public void deleteGraphFromCache(final String graphId) {
      */
     public GraphSerialisable getGraphFromCache(final String graphId) throws CacheOperationException {
         final GraphSerialisable graphSerialisable = getGraphSerialisableFromCache(graphId);
-        return (isNull(graphSerialisable)) ? null : graphSerialisable;
+        return (Objects.isNull(graphSerialisable)) ? null : graphSerialisable;
```
GCHQDeveloper314, line 95: See other comment on `Objects.isNull`

GCHQDeveloper314 (changes requested):
> Adds to the Cache are now not checked exceptions by default, gets are. This is the assumption that a failure to add a value is not that bad so a runtime exception should suffice but if a value cannot be returned then that is more exceptional.

This is not in scope for this issue and will cause breaking changes to 3rd party cache implementations. Adding to the cache is essential in most cases and not recoverable. This should be reverted.

core/graph/src/main/java/uk/gov/gchq/gaffer/graph/hook/NamedOperationResolver.java
```diff
@@ -26,32 +26,30 @@
...
 
-import static uk.gov.gchq.gaffer.store.operation.handler.util.OperationHandlerUtil.updateOperationInput;
-
 /**
  * A {@link GraphHook} to resolve named operations.
  */
 @JsonPropertyOrder(alphabetic = true)
 public class NamedOperationResolver implements GetFromCacheHook {
     private static final Logger LOGGER = LoggerFactory.getLogger(NamedOperationResolver.class);
+    // TODO: Possibly implement a depth limit rather than a hard timeout
```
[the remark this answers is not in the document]
tb06904: True I was intending to add an issue in for it as I suspect this recursion may be hurting some performance too.

core/operation/src/main/java/uk/gov/gchq/gaffer/jobtracker/JobTracker.java
tb06904: So the race condition should be mitigated as only additions to the job tracker cache are async and link queued up to preserve order. Gets from cache are synced as if requested it will wait on the queue to finish so any additions in the executor queue are completed first, this is done via the `get()` on the executor submit.

core/cache/src/main/java/uk/gov/gchq/gaffer/cache/ICacheService.java
```diff
@@ -73,11 +73,13 @@ default <K, V> V getFromCache(final String cacheName, final K key) {
...
-     * @throws CacheOperationException if there is an error adding the new key-value pair to the cache
      */
-    default <K, V> void putInCache(final String cacheName, final K key, final V value) throws CacheOperationException {
-        final ICache<K, V> cache = getCache(cacheName);
-        cache.put(key, value);
+    default <K, V> void putInCache(final String cacheName, final K key, final V value) {
+        try {
+            getCache(cacheName).put(key, value);
+        } catch (final CacheOperationException e) {
+            throw new GafferRuntimeException("Failed to add key: " +  key + " value: " + value + " to cache: " + cacheName, e);
```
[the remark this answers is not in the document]
tb06904: Ahh meant this to be in the Job Tracker specific implementation, which I actually think has been fixed in another ticket will change.

core/cache/src/main/java/uk/gov/gchq/gaffer/cache/ICacheService.java
```diff
@@ -59,8 +57,9 @@ public interface ICacheService {
      * @param <K>       The object type that acts as the key for the cache
      * @param <V>       The value that is stored in the cache
      * @return the requested cache object
+     * @throws CacheOperationException if issue getting from cache.
      */
-    default <K, V> V getFromCache(final String cacheName, final K key) {
+    default <K, V> V getFromCache(final String cacheName, final K key) throws CacheOperationException {
         final ICache<K, V> cache = getCache(cacheName);
         return cache.get(key);
```
GCHQDeveloper314, line 64: Is this throwing an exception now, or did you mean to roll this back. Also the two lines can be condensed into one as you've done with the other methods.

core/operation/src/main/java/uk/gov/gchq/gaffer/jobtracker/JobTracker.java
GCHQDeveloper314: I see; am I right in thinking that the benefit from this is that calls to these methods will now return right away?
I'd suggest expanding the comment a little to explain the motivation and make clear there is no parallel execution.

core/cache/src/main/java/uk/gov/gchq/gaffer/cache/ICache.java
```diff
@@ -36,8 +35,10 @@ public interface ICache<K, V> {
      *
      * @param key the key to lookup in the cache
      * @return the value associated with the key
+     * @throws CacheOperationException if there is an error getting the key-value
+     *                                 pair from the cache
      */
-    V get(final K key);
+    V get(final K key) throws CacheOperationException;
```
cn337131, line 41: can we use unchecked instead of checked exceptions here?

core/store/src/main/java/uk/gov/gchq/gaffer/store/operation/handler/named/cache/NamedViewCache.java
```diff
@@ -155,9 +151,12 @@ public Iterable<NamedViewDetail> getAllNamedViews(final User user, final String
...
      */
-    public void addToCache(final NamedViewDetail namedView, final boolean overwrite)
-            throws CacheOperationException {
-        super.addToCache(namedView.getName(), namedView, overwrite);
+    public void addToCache(final NamedViewDetail namedView, final boolean overwrite) throws CacheOperationException {
+        try {
+            super.addToCache(namedView.getName(), namedView, overwrite);
+        } catch (final GafferRuntimeException e) {
+            throw new CacheOperationException(e);
+        }
```
cn337131, line 159: can we not use unchecked again? If too much of a minefield for this ticket maybe look at exceptions in another ticket

core/cache/src/main/java/uk/gov/gchq/gaffer/cache/ICacheService.java
tb06904: So I think this throwing an exception is possibly correct as the `getFromCache` method in the `Cache` class already had this declaration on which actually just calls this method e.g. the services `getFromCache` by default. Adding it here makes it consistent across the classes and is probably correct that it is an exception.

core/store/src/main/java/uk/gov/gchq/gaffer/store/operation/handler/named/cache/NamedViewCache.java
tb06904: Good spot I was meant to back this change out following the other comment about adds to cache should be checked exceptions

core/operation/src/main/java/uk/gov/gchq/gaffer/jobtracker/JobTracker.java
```diff
@@ -38,6 +40,15 @@ public class JobTracker extends Cache<String, JobDetail> {
     private static final Logger LOGGER = LoggerFactory.getLogger(JobTracker.class);
     private static final String CACHE_SERVICE_NAME_PREFIX = "JobTracker";
 
+    /**
+     * Executor to allow queuing up async operations on the cache.
+     * This is largely to allow additions to the cache to not block the calling
+     * thread as it will impact overall operation performance. Gets from the
+     * cache are executed in the pool but as has a thread size of one they wait
```
GCHQDeveloper314, line 47:
```suggestion
     * thread, which would impact overall operation performance. Gets from the
     * cache are executed in the pool, but as it has a thread size of one they wait
```

GCHQDeveloper314 (approved):

cn337131 (approved):

PR #3174, opened 2024-03-18
Gh-3149 Add depth limit to NamedOperationResolver

### LINKED ISSUE #3149: Use depth limit rather than timeout on NamedOperationResolver
The `NamedOperationResolver` uses a timeout to control how long it will try resolve from cache, this is to stop infinite loops due to nested operations. 

A better approach for this would be to implement a depth limit instead that controls how nested into a `NamedOperation` the resolver will go. This approach prevents infinite loops but also ensures the resolver succeeds under more circumstances and won't get stuck or timeout before finishing.

_Originally posted by @GCHQDeveloper314 in https://github.com/gchq/Gaffer/pull/3148#discussion_r1464815898_

tb06904:
Adds a configurable depth limit to the `NamedOperatationResolver` graph hook so that nested named operations are only resolved to a certain depth (set to 3 by default). 

This also tidies and refactors the class to simplify the recursion and logic so all `NamedOperations` will be resolved to their respective `OperationChain`s, with extended debug and logging for if recursion limit is hit so user is aware. If the limit is hit, the `NamedOperation` is simply left as its class in the chain.

# Related issue

- Resolve #3149

### DIFF
--- a/core/graph/src/main/java/uk/gov/gchq/gaffer/graph/hook/NamedOperationResolver.java
+++ b/core/graph/src/main/java/uk/gov/gchq/gaffer/graph/hook/NamedOperationResolver.java
@@ -27,7 +27,6 @@ import uk.gov.gchq.gaffer.cache.exception.CacheOperationException;
 import uk.gov.gchq.gaffer.named.operation.NamedOperation;
 import uk.gov.gchq.gaffer.operation.Operation;
 import uk.gov.gchq.gaffer.operation.OperationChain;
-import uk.gov.gchq.gaffer.operation.Operations;
 import uk.gov.gchq.gaffer.store.Context;
 import uk.gov.gchq.gaffer.store.operation.handler.named.cache.NamedOperationCache;
 import uk.gov.gchq.gaffer.store.operation.handler.util.OperationHandlerUtil;
@@ -35,6 +34,7 @@ import uk.gov.gchq.gaffer.user.User;
 
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.Collections;
 
 /**
  * A {@link GraphHook} to resolve named operations.
@@ -103,7 +103,7 @@ public class NamedOperationResolver implements GetFromCacheHook {
      * Will also run recursively to a given depth limit to ensure any nested
      * {@link NamedOperation}s are also resolved from the cache.
      *
-     * @param operation {@link NamedOperation} or {@link Operations} list to act on.
+     * @param operation {@link NamedOperation} or {@link OperationChain} to act on.
      * @param user      User for the cache access.
      * @param depth     Current recursive depth, will use limit set in class to
      *                  continue or not.
@@ -126,26 +126,21 @@ public class NamedOperationResolver implements GetFromCacheHook {
                 // Update the operation inputs and add operation chain to the updated list
                 OperationHandlerUtil.updateOperationInput(namedOperationChain, namedOperation.getInput());
 
-                // Run again to get resolve any nested operations in the chain before adding
+                // Run again to resolve any nested operations in the chain before adding
                 namedOperationChain.updateOperations(resolveNamedOperations(namedOperationChain, user, depth + 1));
                 updatedOperations.add(namedOperationChain);
 
             } catch (final CacheOperationException e) {
                 LOGGER.error("Exception resolving NamedOperation within the cache: {}", e.getMessage());
+                return Collections.singletonList(namedOperation);
             }
-        // If given a list of operations then resolve them
-        } else if (operation instanceof Operations) {
-            LOGGER.debug("Resolving Operation List: {}", ((Operations<?>) operation).getOperations());
-            for (final Operation op : ((Operations<?>) operation).getOperations()) {
-                // If just a plain operation then nothing to resolve
-                if (!(op instanceof NamedOperation) && !(op instanceof Operations)) {
-                    updatedOperations.add(op);
-                    continue;
-                }
-
+        // If given an operationChain resolve Operations inside it
+        } else if (operation instanceof OperationChain) {
+            LOGGER.debug("Resolving Operations in Operation Chain: {}", ((OperationChain<?>) operation).getOperations());
+            for (final Operation op : ((OperationChain<?>) operation).getOperations()) {
                 // Resolve if haven't hit the depth limit for named operations yet
                 // Note only need to check the depth here as when checking for a nested named operation it will always be
-                // passed as an operation chain first
+                // recursively passed as an operation chain
                 if (depth < depthLimit) {
                     updatedOperations.addAll(resolveNamedOperations(op, user, depth));
                 } else {
@@ -154,6 +149,9 @@ public class NamedOperationResolver implements GetFromCacheHook {
                 }
 
             }
+        // If just a plain operation then nothing to resolve
+        } else {
+            updatedOperations.add(operation);
         }
         return updatedOperations;
     }
--- a/core/graph/src/main/java/uk/gov/gchq/gaffer/graph/hook/NamedOperationResolver.java
+++ b/core/graph/src/main/java/uk/gov/gchq/gaffer/graph/hook/NamedOperationResolver.java
@@ -24,6 +24,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import uk.gov.gchq.gaffer.cache.exception.CacheOperationException;
+import uk.gov.gchq.gaffer.core.exception.GafferRuntimeException;
 import uk.gov.gchq.gaffer.named.operation.NamedOperation;
 import uk.gov.gchq.gaffer.operation.Operation;
 import uk.gov.gchq.gaffer.operation.OperationChain;
@@ -82,6 +83,8 @@ public class NamedOperationResolver implements GetFromCacheHook {
 
     @Override
     public void preExecute(final OperationChain<?> opChain, final Context context) {
+        // Resolve the named operations in the chain
+        // (depth is set to zero as this is first level of recursion when checking for nested named ops)
         opChain.updateOperations(resolveNamedOperations(opChain, context.getUser(), 0));
     }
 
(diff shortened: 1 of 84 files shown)


core/graph/src/main/java/uk/gov/gchq/gaffer/graph/hook/NamedOperationResolver.java
```diff
@@ -37,79 +35,54 @@
...
-            CompletableFuture.runAsync(() -> resolveNamedOperations(opChain, context.getUser())).get(timeout, timeUnit);
-            LOGGER.info("Finished Named Operation resolver");
-        } catch (final ExecutionException | TimeoutException e) {
-            throw new GafferRuntimeException("ResolverTask did not complete: " + e.getMessage(), e);
-        } catch (final InterruptedException e) {
-            LOGGER.warn("Thread interrupted", e);
-            Thread.currentThread().interrupt();
-        }
-
+        opChain.updateOperations(resolveNamedOperations(opChain, context.getUser(), 0));
```
GCHQDev404, line 85: depth hard coded to zero ?
I this correct, If this is correct can you stick a comment to explain why its zero.

core/graph/src/main/java/uk/gov/gchq/gaffer/graph/hook/NamedOperationResolver.java
```diff
@@ -122,42 +95,64 @@ public <T> T onFailure(final T result, final OperationChain<?> opChain, final Co
...
+     * Resolves any named operations from the cache. What is meant
+     * by 'resolved' is turning the named operations into their respective
+     * {@link OperationChain}s. Ensures any supplied {@link NamedOperation}s
+     * actually exist in the cache and contain their correct {@link Operation}s.
+     * Will also run recursively to a given depth limit to ensure any nested
+     * {@link NamedOperation}s are also resolved from the cache.
+     *
+     * @param operation {@link NamedOperation} or {@link OperationChain} to act on.
+     * @param user      User for the cache access.
+     * @param depth     Current recursive depth, will use limit set in class to
```
GCHQDev404, line 108: inclusive or exclusive of the number, does zero resolve nothing?

core/graph/src/main/java/uk/gov/gchq/gaffer/graph/hook/NamedOperationResolver.java
```diff
@@ -122,42 +95,64 @@ public <T> T onFailure(final T result, final OperationChain<?> opChain, final Co
...
+        } else if (operation instanceof OperationChain) {
+            LOGGER.debug("Resolving Operations in Operation Chain: {}", ((OperationChain<?>) operation).getOperations());
+            for (final Operation op : ((OperationChain<?>) operation).getOperations()) {
+                // Resolve if haven't hit the depth limit for named operations yet
+                // Note only need to check the depth here as when checking for a nested named operation it will always be
+                // recursively passed as an operation chain
+                if (depth < depthLimit) {
+                    updatedOperations.addAll(resolveNamedOperations(op, user, depth));
+                } else {
+                    LOGGER.warn("Nested depth limit hit resolving NamedOperations, operation maybe unresolved {}", op);
```
GCHQDev404, line 147: So this information is not given to the user?
Will what happens next be:

1.  that the operation chain is then executed
2. `NamedOperationHandler` will error to the user and not explain its depth resolving issue?

I don't think I have seen any testing for this situation.

GCHQDev404 (changes requested):

core/graph/src/main/java/uk/gov/gchq/gaffer/graph/hook/NamedOperationResolver.java
GCHQDev404: In the event of ProxyStore or FederatedStore would this unresolved operation get sent to sub-graphs?
could this result in unexpected behaviour?

core/graph/src/main/java/uk/gov/gchq/gaffer/graph/hook/NamedOperationResolver.java
GCHQDev404: I believe this might show that there are no tests to check/prove that the depth limited is working or changed with fail and pass depths.

tb06904: This is correct if you follow the logic, the depth needs to be initialised somewhere this is the outer most and first call to the method so depth is currently zero. There is unit test to ensure the limit is working see line 140 in the `NamedOperationResolverTest`

core/graph/src/main/java/uk/gov/gchq/gaffer/graph/hook/NamedOperationResolver.java
tb06904: Not quite sure what you mean, the limit is set on a class level this is just the current depth of recursion of the method.

core/graph/src/main/java/uk/gov/gchq/gaffer/graph/hook/NamedOperationResolver.java
tb06904: The `Operation` will be left untouched in the chain, there's a chance the `Operation` checked here might not be a `NamedOperation` at which point this is not a problem, an additional check could probably be added here that checks if it is a `NamedOperation` and errors accordingly.

core/graph/src/main/java/uk/gov/gchq/gaffer/graph/hook/NamedOperationResolver.java
```diff
@@ -37,79 +35,54 @@
...
-import java.util.concurrent.TimeUnit;
-import java.util.concurrent.TimeoutException;
 
 /**
  * A {@link GraphHook} to resolve named operations.
  */
 @JsonPropertyOrder(alphabetic = true)
 public class NamedOperationResolver implements GetFromCacheHook {
     private static final Logger LOGGER = LoggerFactory.getLogger(NamedOperationResolver.class);
-    // TODO: Possibly implement a depth limit rather than a hard timeout
```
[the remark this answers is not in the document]
tb06904: This was obviously missed but a ticket was raised and it has now been removed.

core/graph/src/main/java/uk/gov/gchq/gaffer/graph/hook/NamedOperationResolver.java
tb06904: This has been tweaked now to throw exception if limit is hit and there are still unresolved `NamedOperations` in the chain.

core/graph/src/main/java/uk/gov/gchq/gaffer/graph/hook/NamedOperationResolver.java
GCHQDev404: Foolish me, I miss read "depth" and "depthLimit"

core/graph/src/main/java/uk/gov/gchq/gaffer/graph/hook/NamedOperationResolver.java
GCHQDev404: Foolish me, I miss read "depth" and "depthLimit"

core/graph/src/main/java/uk/gov/gchq/gaffer/graph/hook/NamedOperationResolver.java
```diff
@@ -122,42 +98,67 @@ public <T> T onFailure(final T result, final OperationChain<?> opChain, final Co
...
+            }
+        // If given an operationChain resolve Operations inside it
+        } else if (operation instanceof OperationChain) {
+            LOGGER.debug("Resolving Operations in Operation Chain: {}", ((OperationChain<?>) operation).getOperations());
+            for (final Operation op : ((OperationChain<?>) operation).getOperations()) {
+                // Resolve if haven't hit the depth limit for named operations yet
+                // Note only need to check the depth here as when checking for a nested named operation it will always be
+                // recursively passed as an operation chain
+                if (depth < depthLimit) {
+                    updatedOperations.addAll(resolveNamedOperations(op, user, depth));
```
GCHQDev404, line 148: so:

```
OpChain[OpChain[OpChain[OpChain[OpChain[OpChain[NamedOp]]]]]]
```
will still be at the same depth because the depth/nesting of opchains is irrelevant, correct?

core/graph/src/main/java/uk/gov/gchq/gaffer/graph/hook/NamedOperationResolver.java
GCHQDev404: Now that it errors, nothing should get to FedStore or ProxyStore.

GCHQDev404 (approved):

core/graph/src/main/java/uk/gov/gchq/gaffer/graph/hook/NamedOperationResolver.java
tb06904: That's correct it is only the depth of the Named Operation nesting that is of concern

core/graph/src/test/java/uk/gov/gchq/gaffer/graph/hook/NamedOperationResolverTest.java
```diff
@@ -139,13 +138,82 @@ public void shouldResolveNestedNamedOperation(@Mock final User user,
...
+                                                     @Mock final GetElements op2,
+                                                     @Mock final GetElements op3,
+                                                     @Mock final Iterable<? extends EntityId> input)
+            throws CacheOperationException {
+        // Given
+        final String namedOp1Name = "namedOp1";
+        final String namedOp2Name = "namedOp2";
+        final String namedOp3Name = "namedOp3";
+        final String namedOp4Name = "namedOp4";
+        final NamedOperationResolver resolver = new NamedOperationResolver(cache);
```
GCHQDev404, line 156: Where is the depth value being set?
Is this only testing the default value?

GCHQDev404 (changes requested):

core/graph/src/test/java/uk/gov/gchq/gaffer/graph/hook/NamedOperationResolverTest.java
tb06904: Yeah this is only testing the default depth of 3 (set in the resolver class) its not massively easy to programmatically create these nested operations so thought would be fine to just test depth of 3

GCHQDev404: Can you change it to 2 then.
or copy and paste and change it 4.
It is to prove the you can change from the default value.

tb06904: Good point probably not a bad idea to have a test around this, though the previous implementation was missing a JSON config test too

GCHQDeveloper314 (approved):

cn337131 (approved):

PR #3180, opened 2024-03-26
Gh-2994: Make gremlin logging configurable

### LINKED ISSUE #2994: Make tinkerpop Gaffer operation logging configurable
As discussed in https://github.com/gchq/Gaffer/pull/2984#discussion_r1238652217, we should make the logging of Gaffer Operations in tinkerpop configurable and not active by default.

It would clear up the logs in normal usage.

cn337131:
Costly logging should only display now if the SLF4J log level is enabled. 

# Related issue

- Resolve #2994

### DIFF
--- a/library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/GafferPopGraph.java
+++ b/library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/GafferPopGraph.java
@@ -637,8 +637,8 @@ public class GafferPopGraph implements org.apache.tinkerpop.gremlin.structure.Gr
                 Object input = ((Input) operation).getInput();
                 if (input != null) {
                    if (input instanceof MappedIterable) {
-                        ((MappedIterable) input).forEach(item -> { 
-                            LOGGER.info("GafferPop operation input: {}", item); 
+                        ((MappedIterable) input).forEach(item -> {
+                            LOGGER.info("GafferPop operation input: {}", item);
                         });
                     } else {
                         LOGGER.info("GafferPop operation input: {}", input);
--- a/library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/GafferPopGraph.java
+++ b/library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/GafferPopGraph.java
@@ -633,7 +633,7 @@ public class GafferPopGraph implements org.apache.tinkerpop.gremlin.structure.Gr
         for (final Operation operation : opChain.getOperations()) {
             operation.setOptions(opOptions);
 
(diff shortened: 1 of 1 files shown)


GCHQDeveloper314 (approved):

library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/GafferPopGraph.java
```diff
@@ -633,21 +633,22 @@ public <T> T execute(final OperationChain<T> opChain) {
...
+                        LOGGER.debug("GafferPop operation input: {}", input);
                     }
                 }
             }
-
         }
 
         try {
-            LOGGER.info("GafferPop operation chain called: " + opChain.toString());
+            LOGGER.info("GafferPop operation chain called: {}", opChain);
```
tb06904, line 651: Would maybe the `toOverviewString()`  method for the operation chain give nicer output here?

library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/GafferPopGraph.java
```diff
@@ -633,21 +633,22 @@ public <T> T execute(final OperationChain<T> opChain) {
         for (final Operation operation : opChain.getOperations()) {
             operation.setOptions(opOptions);
 
-            if (operation instanceof Input) {
+            if (LOGGER.isDebugEnabled() && operation instanceof Input) {
                 Object input = ((Input) operation).getInput();
                 if (input != null) {
```
tb06904, line 638: Is this null check actually required?

cn337131: Potentially not now that it has moved from info to debug - so can be removed I think :)

tb06904: I was more thinking that as the only path that uses the `input` variable now checks if it is an instance of `MappedIterable` there is now no need for a null check (because you have removed the `toString()` call on the else branch).

GCHQDeveloper314: I think `LOGGER.isDebugEnabled()` should remain as this prevents needlessly iterating though the operation when nothing will ever be logged.

GCHQDeveloper314 (approved):
I note that because the debug log level is not enabled for tests the test coverage is shown as declining in this PR. Usually this wouldn't be a problem, it's only an issue here because the code is not run at all. This is one of the last things I'd change to improve coverage, so I don't think any further changes are required.

PR #3181, opened 2024-03-26
Gh-2495: FederatedGraphStorage has Accumulo specific code

### LINKED ISSUE #2495: FederatedStore FederatedGraphStorage has Accumulo specific code
The Accumulo specific code in FederatedGraphStorage fixes mismatch in tables and graph name.

1. Does a similar problem effect other stores?
2. Does this code belong somewhere else

Possible Likely solution. There should be a mapping of GraphIds to TableID names so map can be updated without needing to rename tables.

link: gh-2435

GCHQDeveloper314:
# Related issue

- Resolve #2495

### DIFF
--- a/store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/FederatedGraphStorage.java
+++ b/store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/FederatedGraphStorage.java
@@ -407,11 +407,12 @@ public class FederatedGraphStorage {
         final GraphSerialisable graphToUpdate = getGraphToUpdate(graphId, accessPredicate);
 
         if (nonNull(graphToUpdate)) {
-            //get access before removing old graphId.
+            // Get access before removing old graphId.
             FederatedAccess access = federatedStoreCache.getAccessFromCache(graphId);
-            //Removed first, to stop a sync issue when sharing the cache with another store.
+            // Remove first, to stop a sync issue when sharing the cache with another store.
             remove(graphId, federatedAccess -> true, false);
-
+            // For Accumulo derived stores the table name must be updated to match the graph's new id. These stores are:
+            // uk.gov.gchq.gaffer.accumulostore.[AccumuloStore, SingleUseAccumuloStore, MiniAccumuloStore, SingleUseMiniAccumuloStore]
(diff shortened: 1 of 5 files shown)


store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/FederatedGraphStorage.java
```diff
@@ -439,26 +438,18 @@ private void updateCacheWithNewGraphId(final String newGraphId, final GraphSeria
         }
     }
 
+    /* For Accumulo derived stores the table name must
+     * be updated to match the graph's new id.
+     * These stores are:
+     * uk.gov.gchq.gaffer.accumulostore.[AccumuloStore, SingleUseAccumuloStore,
+     * MiniAccumuloStore, SingleUseMiniAccumuloStore]
+     */
```
tb06904, line 446: Is it worth making this a proper javadoc and moving this comment to the actual `renameTable()` call?

tb06904 (approved):

store-implementation/federated-store/src/main/java/uk/gov/gchq/gaffer/federatedstore/FederatedGraphStorage.java
GCHQDeveloper314: I'd considered making it proper Javadoc but as this is a private method I didn't see much reason to. Do you mean moving this to where it was before or somewhere else? This comment relates to the Federated Store so it shouldn't be in the Accumulo module.

tb06904: I'm in the camp believing that all methods even private ones should have javadocs on as they can help developers understand the code but it is bit of a debated topic, I more meant this comment isn't really generic enough to relate to the method it should be where the method is called.

GCHQDeveloper314: I've moved the comment to where the method is called. This fits in with other comments in that area - I agree this is a better place for it.

tb06904 (approved):

cn337131 (approved):

PR #3184, opened 2024-04-02
Gh-3183 Remove gafferpop specifics from operation chains

### LINKED ISSUE #3183: Gafferpop includes library specific classes in operation chains
**Describe the bug**
This is relevant when using a proxy store to connect gafferpop to an existing a graph, the gremlin functions will fail. This is due to a slight implementation oversight where the gafferpop element generators are included in the operation chains that are used. This is an issue when using the proxy store as the whole chain is sent to the remote graph which may not have tinkerpop library and serialisation is currently missing. 

**Expected behaviour**
A better solution would be to have this element translation happen locally instead. Essentially only the 'plain' Gaffer operations should be executed on the graph via gafferpop, all gafferpop specific bits should happen before or on the results outside of the chain sent to the remote graph.

tb06904:
Fixes issue with the current Gafferpop library where the resulting operation chain that would be ran on the graph included the element generators specific to Gafferpop. They are now called locally on the results of a chain or before running a chain so that the target graph does not need to be aware of the Gafferpop library. This also avoids any serialisation errors when using the proxy store to enable gremlin on an existing graph.

*Other fixes*
- Fixed issue with using Edge IDs as seeds
- Stopped the `View` being added twice to `GetAdjacentIds` call as this was removing all returned vertexes.

Tested some of this manually spinning up a Map Store REST and attaching a gremlin server via a proxy store to allow querying.

# Related issue

- Resolve #3183

### DIFF
--- a/library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/GafferPopGraph.java
+++ b/library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/GafferPopGraph.java
@@ -719,16 +719,14 @@ public class GafferPopGraph implements org.apache.tinkerpop.gremlin.structure.Gr
         }
 
         View processedView = view == null ? createAllEntitiesView() : view;
-
         final Iterable<? extends Element> result = execute(new OperationChain.Builder()
                 .first(new GetAdjacentIds.Builder()
                         .input(seeds)
                         .view(processedView)
                         .inOutType(getInOutType(direction))
                         .build())
-                .then(new GetElements.Builder()
-                        .view(processedView)
-                        .build())
+                // GetAdjacentIds provides list of entity seeds so run a GetElements to get the actual Entities
+                .then(new GetElements())
                 .build());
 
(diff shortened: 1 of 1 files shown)


GCHQDeveloper314 (commented):
This is an interesting fix as it essentially bypasses the usual method of using Gaffer Generators. I think this is ok, though I've never been a big fan of the verbose nature of Generators. I'd like to ask @t92549 if this is a good idea.

t92549:
It's definitely a different approach that will work better for proxy stores but worse for Accumulo stores. This probably just points to a problem in Gaffer where proxy stores fail to execute operations that aren't present on the remote store. I can't remember if there is a way to make them execute locally without using a FederatedStore though. If not, that could perhaps be a separate improvement.

GCHQDeveloper314 (approved):
I will approve this with the caveat that using the generators in this way is a bit of a workaround. Once this is merged, an issue should be raised to review and reimplement once proxy stores have been enhanced and can be configured to handle certain operations locally.

cn337131 (approved):

PR #3188, opened 2024-04-09
Gh-2989: Add configurable GetAllElements Limit to Tinkerpop

### LINKED ISSUE #2989: Add configurable GetAllElements Limit to tinkerpop
There should be a configurable limit to the amount of elements returned in a GetAllElements by tinkerpop. This should use the Limit operation and be configurable. Perhaps it could use a [TraversalStrategy](https://tinkerpop.apache.org/docs/current/reference/#traversalstrategy)

So gremlin users don't accidentally use GetAllElements and create a performance issue

If Gaffer had a property index, then a `TraversalStrategy` could be added to use that instead, as described in https://github.com/gchq/Gaffer/issues/277#issuecomment-254504980

p29876:
# Related issue

- Resolve #2989

### DIFF
--- a/library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/GafferPopGraph.java
+++ b/library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/GafferPopGraph.java
@@ -413,7 +413,7 @@ public class GafferPopGraph implements org.apache.tinkerpop.gremlin.structure.Gr
 
         final int resultSize = IterableUtils.size(translatedResults);
         if (getAll && resultSize >= getAllElementsLimit) {
-            LOGGER.warn("Result size is equal to  configured limit (" + getAllElementsLimit + "). Results may have been truncated");
+            LOGGER.warn("Result size is equal to configured limit ({}). Results may have been truncated", getAllElementsLimit);
         }
 
 
@@ -574,7 +574,7 @@ public class GafferPopGraph implements org.apache.tinkerpop.gremlin.structure.Gr
 
         final int resultSize = IterableUtils.size(translatedResults);
         if (getAll && resultSize >= getAllElementsLimit) {
-            LOGGER.warn("Result size is equal to  configured limit (" + getAllElementsLimit + "). Results may have been truncated");
+            LOGGER.warn("Result size is equal to configured limit ({}). Results may have been truncated", getAllElementsLimit);
         }
 
         return translatedResults.iterator();
--- a/library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/GafferPopGraph.java
+++ b/library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/GafferPopGraph.java
@@ -896,7 +896,9 @@ public class GafferPopGraph implements org.apache.tinkerpop.gremlin.structure.Gr
         variablesMap.put(GafferPopGraphVariables.OP_OPTIONS, Collections.unmodifiableMap(opOptions));
         variablesMap.put(GafferPopGraphVariables.USER, user);
         variablesMap.put(GafferPopGraphVariables.SCHEMA, graph.getSchema());
-        variablesMap.put(GafferPopGraphVariables.GET_ALL_ELEMENTS_LIMIT, getAllElementsLimit);
+        if(null != getAllElementsLimit) {
+            variablesMap.put(GafferPopGraphVariables.GET_ALL_ELEMENTS_LIMIT, getAllElementsLimit);
+        }
         return new GafferPopGraphVariables(variablesMap);
     }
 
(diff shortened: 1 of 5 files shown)


library/tinkerpop/src/test/java/uk/gov/gchq/gaffer/tinkerpop/GafferPopGraphIT.java
```diff
@@ -257,6 +260,37 @@ public void shouldGetAllVertices() {
...
+        final GafferPopGraph graph = GafferPopGraph.open(TEST_CONFIGURATION_2, gafferGraph);
+
+        // When
+        addSoftwareVertex(graph);
+        graph.addVertex(T.label, PERSON_GROUP, T.id, VERTEX_2, NAME_PROPERTY, "Gaffer");
+        final Iterator<Vertex> vertices = graph.vertices();
+
+        // Then
+        final List<Vertex> verticesList = new ArrayList<>();
+        while (vertices.hasNext()) {
```
tb06904, line 276: `assertThat` can probably be simplified to remove need for an extra list:

```java
assertThat(vertices)
    .toIterable()
    .contains(
        new GafferPopVertex(SOFTWARE_NAME_GROUP, VERTEX_1, graph),
        new GafferPopVertex(SOFTWARE_NAME_GROUP, VERTEX_2, graph));
```

library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/GafferPopGraph.java
```diff
@@ -234,6 +241,8 @@ public GafferPopGraph(final Configuration configuration, final Graph graph) {
                 .dataAuths(configuration().getStringArray(DATA_AUTHS))
                 .build();
 
+        getAllElementsLimit = configuration.getInteger(GET_ALL_ELEMENTS_LIMIT, null);
```
GCHQDeveloper314, line 244: I think it would be useful to have a set default, instead of defaulting to `null`. With no default we'll need to make changes to various properties files and it will raise the barrier to entry when setting up - compared to using a sensible default. Gaffer also uses defaults everywhere else in these situations.

library/tinkerpop/src/test/java/uk/gov/gchq/gaffer/tinkerpop/GafferPopGraphIT.java
```diff
@@ -73,10 +73,11 @@ public void shouldConstructGafferPopGraphWithOnlyConfig() {
         // Then
         final Map<String, Object> variables = graph.variables().asMap();
         assertThat(variables.get(GafferPopGraphVariables.USER)).isEqualTo(expectedUser);
+        assertThat(variables.get(GafferPopGraphVariables.GET_ALL_ELEMENTS_LIMIT)).isEqualTo(1);
 
         final Map<String, String> opOptions = (Map<String, String>) variables.get(GafferPopGraphVariables.OP_OPTIONS);
         assertThat(opOptions).containsEntry("key1", "value1").containsEntry("key2", "value2").hasSize(2);
-        assertThat(variables.size()).isEqualTo(3);
+        assertThat(variables.size()).isEqualTo(4);
```
GCHQDeveloper314, line 80: I don't understand the reason for this change.

GCHQDeveloper314 (changes requested):

library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/GafferPopGraph.java
p29876: Can you suggest a sensible default?

GCHQDeveloper314: Probably fairly high, say 1000. Although I'd also like others input. What do you think @tb06904?

tb06904: I'm slightly on the fence, my only worry with a default limit is awareness in that it would be very easy for a user to be unaware there is a default limit. The config would only be relevant to the gremlin side which requires a separate config anyways so no changes for existing graphs.

l46978: As long as this is well documented, I think it should be fine.

tb06904: Fair point, I'd go a little higher as the default though maybe more like 5000? As a best of both worlds if possible could maybe look into throwing an exception if the limit is reached so the user is at least aware when it happens.

rj77259 (approved):

GCHQDeveloper314 (approved):

tb06904 (approved):

PR #3189, opened 2024-04-09
Gh-3187: Add Open Telemetry proof of concept

### LINKED ISSUE #3187: Add a Open Telemetry proof of concept for Operation calls
Add Open Telemetry hooks into the code, starting with a basic proof of concept prototype.

If added to correct locations this could be a much better Job tracker to log and track operations and information. Also would provide better fault data on errors etc.

tb06904:
Adds a basic implementation of Open Telemetry that will record Spans across executions of Operations/OperationChains. Can be toggled on via a `otelActive` option in the `graphConfig.json` (off by default).  Hopefully is a useful start point to be expanded on if required.

Is difficult one to demonstrate via unit testing but been manually tested to ensure Spans look useful with relevant data via a `Jaeger` UI. Uses the auto configure feature of Open Telemetry so can be easily configured where and how to export data to via environment variables e.g.
- `OTEL_SERVICE_NAME=gaffer-rest` - sets the service name to `gaffer-rest`
- `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT=http://jaeger:4317` - sets the endpoint to export traces to e.g. `Jaeger`
- `OTEL_TRACES_EXPORTER=console` - sets the exporter to console rather than the default of OTEL (needs to be OTEL to work with Jaeger).

# Related issue

- Resolve #3187

### DIFF
--- a/core/graph/src/main/java/uk/gov/gchq/gaffer/graph/Graph.java
+++ b/core/graph/src/main/java/uk/gov/gchq/gaffer/graph/Graph.java
@@ -17,6 +17,13 @@
 package uk.gov.gchq.gaffer.graph;
 
 import com.google.common.collect.Lists;
+
+import io.opentelemetry.api.GlobalOpenTelemetry;
+import io.opentelemetry.api.trace.Span;
+import io.opentelemetry.api.trace.StatusCode;
+import io.opentelemetry.context.Scope;
+import io.opentelemetry.sdk.autoconfigure.AutoConfiguredOpenTelemetrySdk;
+
 import org.apache.commons.io.IOUtils;
 import org.apache.commons.lang3.builder.EqualsBuilder;
 import org.apache.commons.lang3.builder.HashCodeBuilder;
@@ -309,8 +316,15 @@ public final class Graph {
 
         final Context clonedContext = request.getContext().shallowClone();
         final OperationChain clonedOpChain = request.getOperationChain().shallowClone();
+
+        Span span = GlobalOpenTelemetry
+            .getTracer("Execute on graph: " + this.getGraphId())
+            .spanBuilder(clonedOpChain.toOverviewString())
+            .setAttribute("gaffer.user", clonedContext.getUser().getUserId())
+            .startSpan();
+        clonedContext.setVariable("GRAPHSPAN", io.opentelemetry.context.Context.current().with(span));
         O result = null;
-        try {
+        try (Scope scope = span.makeCurrent()) {
             updateOperationChainView(clonedOpChain);
             for (final GraphHook graphHook : config.getHooks()) {
                 graphHook.preExecute(clonedOpChain, clonedContext);
@@ -333,10 +347,13 @@ public final class Graph {
             for (final UpdateViewHook hook : hookInstances) {
                 hook.preExecute(clonedOpChain, clonedContext);
             }
+            span.addEvent("Finish Pre-execute");
             result = (O) storeExecuter.execute(clonedOpChain, clonedContext);
+            span.addEvent("Operation Chain Complete");
             for (final GraphHook graphHook : config.getHooks()) {
                 result = graphHook.postExecute(result, clonedOpChain, clonedContext);
             }
+            span.addEvent("Finish Post-execute");
         } catch (final Exception e) {
             for (final GraphHook graphHook : config.getHooks()) {
                 try {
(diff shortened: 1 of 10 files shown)


core/graph/src/main/java/uk/gov/gchq/gaffer/graph/Graph.java
```diff
@@ -309,8 +316,17 @@ private <O> GraphResult<O> _execute(final StoreExecuter<O> storeExecuter, final
...
+        Span span = OtelUtil.startSpan(
+            this.getClass().getName(),
+            "Graph Request: " + clonedOpChain.toOverviewString());
+        span.setAttribute("gaffer.graphId", getGraphId());
+        span.setAttribute("gaffer.jobId", clonedContext.getJobId());
+        span.setAttribute("gaffer.user", clonedContext.getUser().getUserId());
+
         O result = null;
-        try {
+        try (Scope scope = span.makeCurrent()) {
```
GCHQDeveloper314, line 329: This `scope` object doesn't seem to be used inside the try?

pom.xml
```diff
@@ -793,6 +801,27 @@
             <artifactId>slf4j-reload4j</artifactId>
             <version>${slf4j.api.version}</version>
         </dependency>
+        <!-- Opentelementry libraries -->
```
GCHQDeveloper314, line 804: It might be better to specify these as requirements in the POMs for `store`/`graph`, rather than for the whole project.

core/graph/src/main/java/uk/gov/gchq/gaffer/graph/Graph.java
tb06904: It in the try with resources block so that it is auto closed, the scope simply sets the span to be current so that it correctly makes the links between parent and child spans: https://opentelemetry.io/docs/languages/java/instrumentation/#create-spans

pom.xml
tb06904: Possibly but the intention is that this could be added to any part of the project as we see fit its not specific to any one part.

core/graph/src/main/java/uk/gov/gchq/gaffer/graph/Graph.java
GCHQDeveloper314: I thought it would be something like that, thanks for the link. It's an interesting use for the auto close functionality, I'm not sure I like it though. Could you add a single line comment to briefly cover what's happening?

pom.xml
GCHQDeveloper314: True, although every part which would want to use them will probably have a dependency on graph or store, which would bring them in as transient dependencies.

tb06904: Not sure I quite agree with relying on transients in this case, if a module uses a library directly it should be defined in the `dependencies` section. Both the `graph` and `store` modules currently use this library directly but the only dependency between them is `store` -> `graph, so the dependency would have to be in the `store` module's pom or both. Having it at a project level removes any duplication or need for transient reliance.

core/graph/src/main/java/uk/gov/gchq/gaffer/graph/Graph.java
tb06904: I agree does look slightly odd but this auto links the spans, the alternative is passing and linking them manually through the Opentelementry `Context`which is lot more effort. I've added comments above relevant lines to clarify.

pom.xml
GCHQDeveloper314: That makes sense, although currently this parent POM is used for dependencies which are required by all or almost all modules. While this could be used by other modules outside `store`/`graph`, it's not currently. If in future it is used in many places then it certainly would make sense to put them in here.

It's unfortunate that 5 separate dependencies are required as this does magnify the duplication issue.

tb06904: True the 5 different modules isn't ideal but guess its as its quite modulised if it was just one would be easier to just add directly to `graph` and `store` poms

GCHQDeveloper314: Looking at it again, both `graph` and `store` import the new class `OtelUtil.java` from `common-util`, so would it work to put the dependencies in there only?

tb06904: Good point, probably not a bad place to put it as the intention is all span creation goes through that class (so we have control over if its turned on or off) so any modules that implement opentelementry will use the `common-util` module

GCHQDeveloper314 (approved):

cn337131 (approved):

l46978 (approved):

PR #3202, opened 2024-05-01
Gh-2991: Improve User Authorisation in GafferPop

### LINKED ISSUE #2991: Improve user auth in gafferpop
There is basic auth in Gaffer's tinkerpop graph provided via config:
https://github.com/gchq/Gaffer/blob/920bbc3bb3950c015138d91aa3d951bd8415bd30/library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/GafferPopGraph.java#L141
We should look into how this might actually work with real auth, for example using a provided user factory

If connecting to a proxy store, could auth be handled by the target store?

tb06904:
This adds the relevant hooks and framework for full user auth for gremlin server connections to GafferPop. It utilises the existing frameworks provided by Tinkerpop to ensure the authenticated user ID is passed on to the graph to use for the query. The way this works is by leveraging the custom `GafferPopGraphStep` to inject the `userId` via a `with()` step on the requested traversal, this is then passed onto the graph variables and used in the query. There are checks in place to prevent manually adding a `with()` step that sets the `userId` so that only the currently authorised user ID is used.

There is a example/default `Authenticator` class provided but this is intended to be used as a template for a specific implementation for a production environment's auth mechanism (for example tinkerpop provide a kerberos version [here](https://github.com/apache/tinkerpop/blob/master/gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/auth/Krb5Authenticator.java)).

The way the authentication classes are activated is by adding the following config to the gremlin server yaml:

```yaml
authentication: {
  authenticator: uk.gov.gchq.gaffer.tinkerpop.server.auth.DefaultGafferPopAuthenticator
}
authorization: {
  authorizer: uk.gov.gchq.gaffer.tinkerpop.server.auth.GafferPopAuthoriser
}
```

# Related issue

- Resolve #2991

### DIFF
--- a/library/tinkerpop/pom.xml
+++ b/library/tinkerpop/pom.xml
@@ -48,6 +48,11 @@
             <artifactId>gremlin-core</artifactId>
             <version>${tinkerpop.version}</version>
         </dependency>
+        <dependency>
+            <groupId>org.apache.tinkerpop</groupId>
+            <artifactId>gremlin-server</artifactId>
+            <version>${tinkerpop.version}</version>
+        </dependency>
 
         <!-- Runtime dependencies -->
         <dependency>
--- a/library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/server/auth/DefaultGafferPopAuthenticator.java
+++ b/library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/server/auth/DefaultGafferPopAuthenticator.java
@@ -16,17 +16,17 @@
 
 package uk.gov.gchq.gaffer.tinkerpop.server.auth;
 
-import java.net.InetAddress;
-import java.nio.charset.StandardCharsets;
-import java.util.Arrays;
-import java.util.HashMap;
-import java.util.Map;
-
 import org.apache.tinkerpop.gremlin.server.auth.AuthenticatedUser;
 import org.apache.tinkerpop.gremlin.server.auth.AuthenticationException;
 import org.apache.tinkerpop.gremlin.server.auth.Authenticator;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+
+import java.net.InetAddress;
+import java.nio.charset.StandardCharsets;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.Map;
 
 /**
  * The default authenticator class for GafferPop, this should not
@@ -43,18 +43,18 @@ public class DefaultGafferPopAuthenticator implements Authenticator {
     }
 
     @Override
-    public void setup(Map<String, Object> config) {
+    public void setup(final Map<String, Object> config) {
         // Nothing to do
     }
 
     @Override
-    public SaslNegotiator newSaslNegotiator(InetAddress remoteAddress) {
+    public SaslNegotiator newSaslNegotiator(final InetAddress remoteAddress) {
         return new PlainTextSaslAuthenticator();
     }
 
 
     @Override
-    public AuthenticatedUser authenticate(Map<String, String> credentials) throws AuthenticationException {
+    public AuthenticatedUser authenticate(final Map<String, String> credentials) throws AuthenticationException {
         // Get the username and password from the credentials set by the SASL negotiator
         final String username = credentials.get("username");
         final String password = credentials.get("password");
@@ -86,8 +86,10 @@ public class DefaultGafferPopAuthenticator implements Authenticator {
 
         @Override
         public AuthenticatedUser getAuthenticatedUser() throws AuthenticationException {
-            if (!complete) throw new AuthenticationException("SASL negotiation not complete");
-            final Map<String,String> credentials = new HashMap<>();
+            if (!complete) {
+                throw new AuthenticationException("SASL negotiation not complete");
+            }
+            final Map<String, String> credentials = new HashMap<>();
             credentials.put("username", username);
             credentials.put("password", password);
             return authenticate(credentials);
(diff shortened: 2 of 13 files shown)


library/tinkerpop/pom.xml
```diff
@@ -48,6 +48,11 @@
             <artifactId>gremlin-core</artifactId>
             <version>${tinkerpop.version}</version>
         </dependency>
+        <dependency>
+            <groupId>org.apache.tinkerpop</groupId>
+            <artifactId>gremlin-server</artifactId>
+            <version>${tinkerpop.version}</version>
+        </dependency>
```
GCHQDeveloper314, line 55: This is required because of importing the auth classes from tinkerpop?

library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/server/auth/DefaultGafferPopAuthenticator.java
```diff
@@ -0,0 +1,135 @@
...
+import java.net.InetAddress;
+import java.nio.charset.StandardCharsets;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * The default authenticator class for GafferPop, this should not
+ * be used in production as it allows all user and password combinations.
+ * The class is intended as a template for an instance specific class
```
GCHQDeveloper314, line 32: Does "instance" mean deployment? As in a specific instance of Gaffer?

library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/server/auth/DefaultGafferPopAuthenticator.java
```diff
@@ -0,0 +1,135 @@
...
+ * that hooks into a proper authorisation mechanism such as LDAP etc.
+ */
+public class DefaultGafferPopAuthenticator implements Authenticator {
```
GCHQDeveloper314, line 35: I'm wondering if this should be called "default" if we don't want it to be used except for testing.

GCHQDeveloper314 (commented):
To check my understanding, when these new classes are not used, the default user for GafferPop is used as the Gaffer user, and when they are used, the username is added using `with`.

I'm assuming a username still be supplied using `with` (as before) if these classes are not enabled in config?

library/tinkerpop/pom.xml
tb06904: That's correct its for the auth classes as they are specifically for authenticating with a gremlin server

library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/server/auth/DefaultGafferPopAuthenticator.java
tb06904: Yeah probably could be changed to deployment, wasnt sure on best terminology

library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/server/auth/DefaultGafferPopAuthenticator.java
tb06904: That is probably true maybe could be changed to `ExampleGafferPopAuthenticator` as its not really a production ready class

tb06904:
> To check my understanding, when these new classes are not used, the default user for GafferPop is used as the Gaffer user, and when they are used, the username is added using `with`.
> 
> I'm assuming a username still be supplied using `with` (as before) if these classes are not enabled in config?

Correct, when you don't specify any auth classes the user ID in the `gafferpop.properties` file will be used to construct the Gaffer user and you can arbitrarily set it via a `with("userId", "id")` to anything for a given query. 

When the `GafferPopAuthoriser` is used it will block any attempts at setting the ID via a `with()` and instead pass on the ID of the current authorised user (from the authenticator class) by injecting its own `with()` into the query.

GCHQDeveloper314 (approved):

cn337131 (approved):

library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/server/auth/ExampleGafferPopAuthenticator.java
```diff
@@ -0,0 +1,135 @@
...
+        }
+
+        @Override
+        public AuthenticatedUser getAuthenticatedUser() throws AuthenticationException {
+            if (!complete) {
+                throw new AuthenticationException("SASL negotiation not complete");
+            }
+            final Map<String, String> credentials = new HashMap<>();
+            credentials.put("username", username);
+            credentials.put("password", password);
```
rj77259, line 89: Just thinking will this be plain text in memory? If so, just wondering how we would go about ensuring that this is secure through the journey of the authentication?

tb06904: As mentioned in previous comment and in the javadocs this is not a production ready class it is just an example for plain text authentication for us to test with each deployment will have to implement their own `Authenticator` that hooks into their own authentication method e.g. LDAP, Kerberos etc.

rj77259 (approved):

PR #3205, opened 2024-05-02
Gh 3203 gafferpop filter edges by label

### LINKED ISSUE #3203: GafferPop not filtering edges by label
**Describe the bug**
The following gremlin query fails in the Cucumber test suite
```
g.V().choose(__.out('knows').count().is(P.gt(0)),
             __.out("knows"),
             __.identity())
      .values('name')
```
This is due to out('knows') not filtering the edges

p29876:
# Related issue

- Resolve #3203

### DIFF
--- a/library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/GafferPopGraph.java
+++ b/library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/GafferPopGraph.java
@@ -566,8 +566,11 @@ public class GafferPopGraph implements org.apache.tinkerpop.gremlin.structure.Gr
                     .map(e -> Arrays.asList((String[]) e))
                     .forEach(edgeLables::addAll);
 
-            // Assume everything that isn't a String[] is an element seed
-            List<Object> elements = partitions.get(false);
+            // Assume everything none-null that isn't a String[] is an element seed
+            List<Object> elements = partitions.get(false)
+                .stream()
+                .filter(e -> e != null)
+                .collect(Collectors.toList());
 
             if (edgeLables.isEmpty()) {
                 // Get all edges
@@ -927,16 +930,6 @@ public class GafferPopGraph implements org.apache.tinkerpop.gremlin.structure.Gr
             // Extract source and destination from ID list
             } else if (id instanceof Iterable) {
                 ((Iterable<?>) id).forEach(edgeIdList::add);
-            // } else if (id instanceof Object[]) {
-            //     // This is probably a label .. 'knows'
-            //     // how do we turn that into an EdgeSeed??
-
-
-            //     Object[] arr = (Object[]) id;
-            //     for(Object o : arr) {
-            //         EdgeSeed seed = new EdgeSeed();
-            //         edgeIdList.add(o);
-            //     }
             // Attempt to extract source and destination IDs from a string form of an array/list
             } else if ((id instanceof String) && (((String) id).matches("^\\[.*,.*\\]$"))) {
                 edgeIdList = Arrays.asList(((String) id)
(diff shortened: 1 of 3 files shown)


library/tinkerpop/src/test/java/uk/gov/gchq/gaffer/tinkerpop/GafferPopGraphIT.java
```diff
@@ -485,7 +486,56 @@ public void shouldGetAllEdgesInGroup() {
         final Iterator<Edge> edges = graph.edges(null, Direction.OUT, CREATED_EDGE_GROUP);
 
         // Then
-        assertThat(edges).toIterable().contains(edgeToAdd1);
+        assertThat(edges).toIterable()
+            .hasSize(1)
+            .contains(edgeToAdd1);
```
GCHQDeveloper314, line 491: It should be possible to use `containsExactly`

library/tinkerpop/src/test/java/uk/gov/gchq/gaffer/tinkerpop/GafferPopGraphIT.java
```diff
@@ -485,7 +486,56 @@ public void shouldGetAllEdgesInGroup() {
...
+        String[] group = {CREATED_EDGE_GROUP};
+
+        // When - get all 'created' edges
+        final Iterator<Edge> edges = graph.edges(null, group);
+
+        // Then - edge 2 (dependsOn) filtered out
+        assertThat(edges).toIterable()
+            .hasSize(2)
+            .contains(edgeToAdd1)
+            .contains(edgeToAdd3);
```
GCHQDeveloper314, line 515: see other comment

library/tinkerpop/src/test/java/uk/gov/gchq/gaffer/tinkerpop/GafferPopGraphIT.java
```diff
@@ -485,7 +486,56 @@ public void shouldGetAllEdgesInGroup() {
...
+
+        String[] group = {CREATED_EDGE_GROUP};
+
+        // When - get all of v1's created edges
+        final Iterator<Edge> edges = graph.edges(VERTEX_1, group);
+
+        // Then - edge 3 (v3 'created') and edge 2 ('dependsOn') filtered out
+        assertThat(edges).toIterable()
+            .hasSize(1)
+            .contains(edgeToAdd1);
```
GCHQDeveloper314, line 538: see other comment

library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/GafferPopGraph.java
```diff
@@ -551,12 +554,38 @@ public Iterator<Edge> edges(final Object... elementIds) {
                 .then(new Limit<>(getAllElementsLimit, true))
                 .build();
         } else {
-            getOperation = new GetElements.Builder()
-                    .input(getElementSeeds(Arrays.asList(elementIds)))
-                    .view(new View.Builder()
+            // Partition elementIds by String[] or Object
+            // The assumption is that the String[] items are edge labels
+            // e.g. elementIds = {"1", ['knows']}
```
tb06904, line 559: I don't know if this assumption is quite right, edge IDs will usually always be in the form of `"[source, dest]"` this would show as a plain string not a string array as the gremlin query would be something like `g.E("[source, dest]")` or `g.E(["[source1, dest1]","[source2, dest2]"])`. Wouldn't if it was a single edge ID it be mistaken for a element seed? I think you might be confusing IDs with labels here?

p29876: Yeah I wasn't sure about this either ... but when I debug it the params passed are `{"1", ["knows"]}`
Easiest example is the federated query that I changed that used `.outE('knows')`
Unless the root of the whole problem is that the params passed are incorrect?

PR #3214, opened 2024-05-15
Gh-3209: GetAll Limit fails on Proxy Store

### LINKED ISSUE #3209: Issues with GafferPop's GetAllElements limitting on proxy stores
**Describe the bug**
This was found testing some of the standard gremlin queries but via the proxy store, there is essentially three smaller but related issues.

- Proxy store can't accept operation chains that have unmodifiable options or null options this needs making more robust.
- The  `GafferPopGraphStep` doesnt actually run the right `vertex()` method on the graph class due to it wrapping the array in a list which pushes it to an override method.
The apache utils method for checking the size of the returned iterator seems to have casting issues from LinkedList to Element when using proxy store this could be some deserialisation issue

tb06904:
Fixes most of the issues in the original issue, two minor fixes:
1. ensures the correct method is called in the `GafferPopGraphStep` 
2. Prevents `null` operation chain options.

Added work around to main serialisation error in the proxy store so that it attempts to deserialise element iterators. This will be fixed in a larger rework of the proxy store so that it does not need to double deserialise.

# Related issue

- Resolve #3209

### DIFF
--- a/core/operation/src/main/java/uk/gov/gchq/gaffer/operation/OperationChain.java
+++ b/core/operation/src/main/java/uk/gov/gchq/gaffer/operation/OperationChain.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016-2023 Crown Copyright
+ * Copyright 2016-2024 Crown Copyright
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/core/operation/src/main/java/uk/gov/gchq/gaffer/operation/OperationChain.java
+++ b/core/operation/src/main/java/uk/gov/gchq/gaffer/operation/OperationChain.java
@@ -66,7 +66,7 @@ import java.util.stream.Collectors;
 public class OperationChain<OUT> implements Output<OUT>,
         Operations<Operation> {
     private List<Operation> operations;
-    private Map<String, String> options;
+    private Map<String, String> options = new HashMap<>();
 
     public OperationChain() {
         this(new ArrayList<>());
@@ -186,7 +186,7 @@ public class OperationChain<OUT> implements Output<OUT>,
     public void setOptions(final Map<String, String> options) {
         if (options == null) {
             this.options = new HashMap<>();
-        } else{
+        } else {
             this.options = options;
         }
     }
--- a/core/operation/src/main/java/uk/gov/gchq/gaffer/operation/OperationChain.java
+++ b/core/operation/src/main/java/uk/gov/gchq/gaffer/operation/OperationChain.java
@@ -184,7 +184,11 @@ public class OperationChain<OUT> implements Output<OUT>,
 
     @Override
     public void setOptions(final Map<String, String> options) {
-        this.options = options;
+        if (options == null) {
+            this.options = new HashMap<>();
+        } else{
+            this.options = options;
+        }
     }
 
     @Override
--- a/library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/GafferPopGraph.java
+++ b/library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/GafferPopGraph.java
@@ -383,6 +383,7 @@ public class GafferPopGraph implements org.apache.tinkerpop.gremlin.structure.Gr
         final OperationChain<Iterable<? extends Element>> getOperation;
 
         if (getAll) {
+            LOGGER.debug("Requested a GetAllElements, results will be truncated to: {}.", getAllElementsLimit);
             getOperation = new Builder()
                     .first(new GetAllElements.Builder()
                             .view(new View.Builder()
@@ -412,9 +413,9 @@ public class GafferPopGraph implements org.apache.tinkerpop.gremlin.structure.Gr
                 .map(e -> (Vertex) e)
                 .iterator();
 
-        if (getAll && IterableUtils.size(translatedResults) > getAllElementsLimit) {
+        if (getAll && IterableUtils.size(translatedResults) == getAllElementsLimit) {
             LOGGER.warn(
-                "Result size is larger than configured limit ({}). Results may have been truncated",
+                "Result size is equal to configured limit ({}). Results may have been truncated",
                  getAllElementsLimit);
         }
         return translatedResults.iterator();
@@ -542,7 +543,7 @@ public class GafferPopGraph implements org.apache.tinkerpop.gremlin.structure.Gr
 
         final OperationChain<Iterable<? extends Element>> getOperation;
         if (getAll) {
-            LOGGER.warn("Requested a GetAllElements results will be truncated to: {}.", getAllElementsLimit);
+            LOGGER.debug("Requested a GetAllElements, results will be truncated to: {}.", getAllElementsLimit);
             getOperation = new Builder()
                 .first(new GetAllElements.Builder()
                         .view(new View.Builder()
@@ -574,9 +575,9 @@ public class GafferPopGraph implements org.apache.tinkerpop.gremlin.structure.Gr
                 .map(e -> (Edge) e)
                 .iterator();
 
-        if (getAll && IterableUtils.size(translatedResults) >= getAllElementsLimit) {
+        if (getAll && IterableUtils.size(translatedResults) == getAllElementsLimit) {
             LOGGER.warn(
-                "Result size is larger than configured limit ({}). Results may have been truncated",
+                "Result size is equal to configured limit ({}). Results may have been truncated",
                 getAllElementsLimit);
         }
         return translatedResults.iterator();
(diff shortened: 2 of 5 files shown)


core/operation/src/main/java/uk/gov/gchq/gaffer/operation/OperationChain.java
```diff
@@ -184,7 +184,11 @@ public Map<String, String> getOptions() {
 
     @Override
     public void setOptions(final Map<String, String> options) {
-        this.options = options;
+        if (options == null) {
+            this.options = new HashMap<>();
```
GCHQDeveloper314, line 188: Why do this if it is now set to a `HashMap` by default; is this in case it's set to something else already?

I'm guessing the new default `HashMap` is because of interaction with another class?

library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/GafferPopGraph.java
```diff
@@ -540,8 +540,9 @@ public Iterator<Vertex> adjVerticesWithView(final Iterable<Object> vertexIds, fi
     public Iterator<Edge> edges(final Object... elementIds) {
         final boolean getAll = null == elementIds || 0 == elementIds.length;
 
-        final Output<Iterable<? extends Element>> getOperation;
+        final OperationChain<Iterable<? extends Element>> getOperation;
         if (getAll) {
+            LOGGER.warn("Requested a GetAllElements results will be truncated to: {}.", getAllElementsLimit);
```
GCHQDeveloper314, line 545:
```suggestion
            LOGGER.warn("Requested a GetAllElements, results will be truncated to: {}.", getAllElementsLimit);
```

store-implementation/proxy-store/src/main/java/uk/gov/gchq/gaffer/proxystore/response/deserialiser/impl/DefaultResponseDeserialiser.java
```diff
@@ -36,8 +42,21 @@ public O deserialise(final String jsonString) throws SerialisationException {
...
-            return JSONSerialiser.deserialise(encodeString(jsonString), typeReference);
+            // For some operations like 'Limit' we may not know the output type up front so it
+            // will be left as a generic '?', this means some objects will not be deserialised correctly.
+            try {
+                // Check if we can deserialise to common outputs like list of elements.
+                // This is a workaround until proxy store has been refactored
+                if (typeReference.getType().getTypeName().equals(Iterable.class.getName() + "<?>")) {
+                    return (O) JSONSerialiser.deserialise(encodeString(jsonString), new TypeReferenceImpl.IterableElement());
+                } else {
+                    return JSONSerialiser.deserialise(encodeString(jsonString), typeReference);
```
GCHQDeveloper314, line 53: This line is duplicated in the catch, I'm wondering if it could be used in a `finally` instead?

library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/GafferPopGraph.java
```diff
@@ -411,12 +412,11 @@ public Iterator<Vertex> vertices(final Object... vertexIds) {
                 .map(e -> (Vertex) e)
                 .iterator();
 
-        final int resultSize = IterableUtils.size(translatedResults);
-        if (getAll && resultSize >= getAllElementsLimit) {
-            LOGGER.warn("Result size is equal to configured limit ({}). Results may have been truncated", getAllElementsLimit);
+        if (getAll && IterableUtils.size(translatedResults) > getAllElementsLimit) {
+            LOGGER.warn(
```
p29876, line 416: Should this ever happen? Is the size > limit the results haven't been truncated

core/operation/src/main/java/uk/gov/gchq/gaffer/operation/OperationChain.java
tb06904: Its mostly due to the proxy store as it was asking for the current options (which were null) then attempting to add to them which hit a null pointer. It's really its more a bug here as shouldn't be able to set the options to null they should just be blank/empty.

library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/GafferPopGraph.java
tb06904: Good spot, I removed this section when debugging so just been added back slightly wrong

store-implementation/proxy-store/src/main/java/uk/gov/gchq/gaffer/proxystore/response/deserialiser/impl/DefaultResponseDeserialiser.java
tb06904: Think that would be incorrect use of a finally block as you are saying you always want to run whatever is in the finally block whereas, in this case we only want to run it conditionally as a fallback.

GCHQDeveloper314: I was thinking of something like this:
```
O rtn = null;
try {
    if (case) {
        rtn = ...
    }
} catch (e) {
    LOGGER.error(...)
} finally {
    if (rtn == null) {
        rtn == fallback
    }
    return rtn;
}
```
Though this might not be a great idea just to save some duplication.

tb06904: Fair I see what you mean, though I'm not sure if its really worth changing especially as this will all be made obsolete under proposed redesign of proxy store

GCHQDeveloper314 (approved):

p29876 (approved):

store-implementation/proxy-store/src/main/java/uk/gov/gchq/gaffer/proxystore/response/deserialiser/impl/DefaultResponseDeserialiser.java
```diff
@@ -36,8 +42,21 @@ public O deserialise(final String jsonString) throws SerialisationException {
...
+                }
+            } catch (final SerialisationException e) {
+                // The input is likely a valid JSON value type, so process using the deserialiser
+                LOGGER.error("Unable to deserialse Iterable<?> as Iterable<Elements> using default deserialisation", e);
+                return JSONSerialiser.deserialise(encodeString(jsonString), typeReference);
```
GCHQDev404, line 58: "try" at line 53 "return JSONSerialiser.deserialise(encodeString(jsonString), typeReference);"
"catch" line 58 "return JSONSerialiser.deserialise(encodeString(jsonString), typeReference);"

the scoping of this try catch does not seem correct.

GCHQDev404 (changes requested):

core/operation/src/main/java/uk/gov/gchq/gaffer/operation/OperationChain.java
```diff
@@ -66,7 +66,7 @@
 public class OperationChain<OUT> implements Output<OUT>,
         Operations<Operation> {
     private List<Operation> operations;
-    private Map<String, String> options;
+    private Map<String, String> options = new HashMap<>();
```
GCHQDev404, line 69: This might change the string representation of JSON of operation chains. It might always display 
`"options":{}` which is verbose.

store-implementation/proxy-store/src/main/java/uk/gov/gchq/gaffer/proxystore/response/deserialiser/impl/DefaultResponseDeserialiser.java
GCHQDeveloper314: Was discussed in [a previous comment here](https://github.com/gchq/Gaffer/pull/3214#discussion_r1603322513). Ideally could be better but is only temporary until a better proxystore is done.

core/operation/src/main/java/uk/gov/gchq/gaffer/operation/OperationChain.java
GCHQDeveloper314: See the discussion [in this comment](https://github.com/gchq/Gaffer/pull/3214#discussion_r1603314029), this is a good point though. It would likely be better to make further changes in the Proxy Store instead.

tb06904: Surely that would be better to have the options field displayed as you would hit the same issue where the key would be missing if trying to parse the JSON. It is arguably safer to prevent the null pointer from ever occurring than to expect every thing else to handle it.

PR #3221, opened 2024-05-22
Gh-3219: Fix inV() and outV() missing results

### LINKED ISSUE #3219: inV() and outV() not supported on GafferPopEdges
**Describe the bug**
Currently a default label will be used if a vertex IDs are used to construct a new edge or if asking for the in/out vertex on the edge. This may result in an incorrect mapping of a Tinkerpop 'label' to a Gaffer 'group', meaning the GafferPopVertex returned by the 'getOutVertex()' and 'getInVertex()' methods likely does not have the correct 'label'.

This is likely due to Gaffer allowing multiple entities/vertexes to be associated with the source and destination vertices of an Edge whereas, Tinkerpop only expects one out vertex and one in vertex on the edge, and hence the 'getOutVertex()' and 'getInVertex()' methods are expected to be used and are part of the interface.

One solution is that a search could be carried out to determine the correct Entity to use when a vertex ID is supplied to an edge, his may hamper performance though, this ticket is to investigate if this is a feature we want to support and the impact of not supporting it.

cn337131:
Implemented a fix which will use an Operation Chain to search the graph for the vertex by ID to get the right results. Fixes 32 cucumber tests. 

# Related issue

- Resolve #3219

### DIFF
--- a/library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/GafferPopEdge.java
+++ b/library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/GafferPopEdge.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016-2023 Crown Copyright
+ * Copyright 2016-2024 Crown Copyright
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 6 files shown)


library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/GafferPopEdge.java
```diff
@@ -26,20 +26,30 @@
...
  * A <code>GafferPopEdge</code> is an {@link GafferPopElement} and {@link Edge}.
  * <p>
  * inVertex() and outVertex() methods are not supported as it is possible for a
- * edge to have multiple in vertices and multiple out vertices (due to the mapping
- * to a TinkerPop vertex to Gaffer {@link uk.gov.gchq.gaffer.data.element.Entity}.
+ * edge to have multiple in vertices and multiple out vertices (due to the
+ * mapping
+ * to a TinkerPop vertex to Gaffer
+ * {@link uk.gov.gchq.gaffer.data.element.Entity}.
  * Use vertices(Direction) instead.
```
p29876, line 53: Do we want to update this comment now?

p29876 (approved):

PR #3225, opened 2024-06-03
Gh-3222: Add OpenCypher translator to GafferPop

### LINKED ISSUE #3222: Add opencypher translator to GafferPop
There exists a opencypher -> gremlin translator in the opencypher project that could be integrated into our tinkerpop implementation so that you can run cypher queries via the gremlin interface.

Would give a best of both worlds allowing users more familiar with cypher a way to query the graph.

Writing our own cypher layer for Gaffer would take a long time and be bespoke this would give us a quick implementation hopefully without much work.

**Additional context**
https://github.com/opencypher/cypher-for-gremlin

tb06904:
This integrates the [opencypher translator](https://github.com/opencypher/cypher-for-gremlin/tree/master/translation) into GafferPop to allow passing a cypher query in a `with()` step. This takes the query and translates it into a Gremlin traversal to then run as normal in GafferPop. Added some limited testing to ensure it is functional but more extensive testing of the translator is already carried out in the opencypher repo.

Also fixes small bug where the limit for `GetAllElements` was missed on some methods in GafferPop.

# Related issue

- Resolve #3222


library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/process/traversal/strategy/optimisation/GafferPopGraphStepStrategy.java
```diff
@@ -24,28 +24,60 @@
...
     }
 
     @Override
     public void apply(final Admin<?, ?> traversal) {
+        // Check for any options on the traversal
+        Optional<OptionsStrategy> optionsStrategy = traversal.getStrategies().getStrategy(OptionsStrategy.class);
+        if (optionsStrategy.isPresent()) {
+            Map<String, Object> options = optionsStrategy.get().getOptions();
+            // Translate and add a cypher traversal in if that key has been set
+            if (options.containsKey(CYPHER_KEY)) {
```
GCHQDeveloper314, line 71: It feels like this could be more concise and avoid the nested if.
```suggestion
        // Translate and add a cypher traversal in if that key has been set
        if (optionsStrategy.isPresent() && optionsStrategy.get().getOptions().containsKey(CYPHER_KEY)) {
```

GCHQDeveloper314 (changes requested):

library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/process/traversal/strategy/optimisation/GafferPopGraphStepStrategy.java
tb06904: This was done like this in case more options were added in future but I think they will probably be handled in the `GafferPopGraphStep` class instead so happy to change.

library/tinkerpop/pom.xml
```diff
@@ -29,6 +29,8 @@
         <tinkerpop.version>3.7.1</tinkerpop.version>
         <google-guice.version>4.2.3</google-guice.version>
         <cucumber.version>7.15.0</cucumber.version>
+        <!-- To use a later version it requires us to update our scala version to v2.12.x -->
```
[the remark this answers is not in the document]
tb06904: Fair point this was more a reminder in case someone updated the version here as it will only crash at runtime not build time and the error message is not overly obvious. I could edit the message but I think its still worth noting here about the dependency.

library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/process/traversal/strategy/optimisation/GafferPopGraphStepStrategy.java
```diff
@@ -24,28 +24,56 @@
...
     }
 
     @Override
     public void apply(final Admin<?, ?> traversal) {
+        // Check for any options on the traversal
+        Optional<OptionsStrategy> optionsStrategy = traversal.getStrategies().getStrategy(OptionsStrategy.class);
+        // Translate and add a cypher traversal in if that key has been set
+        if (optionsStrategy.isPresent() && optionsStrategy.get().getOptions().containsKey(CYPHER_KEY)) {
+            LOGGER.info("Replacing traversal with translated Cypher query");
+            CypherAst ast = CypherAst.parse((String) optionsStrategy.get().getOptions().get(CYPHER_KEY));
```
GCHQDeveloper314, line 70: What happens if the Cypher query is malformed/invalid? It doesn't seem that this method throws an exception, at least not a checked one anyway.

library/tinkerpop/src/test/java/uk/gov/gchq/gaffer/tinkerpop/process/traversal/strategy/GafferPopCypherIT.java
GCHQDeveloper314, line 1: Could you add a further test to cover what happens if someone submits an invalid Cypher query?

library/tinkerpop/pom.xml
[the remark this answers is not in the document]
GCHQDeveloper314: Could also use the enforcer plugin to prevent any accidental upgrades. This is more verbose than a comment but it can't be ignored.

library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/process/traversal/strategy/optimisation/GafferPopGraphStepStrategy.java
tb06904: So you will get a `SyntaxException` and the query will be rejected (gives a reason too) I agree with your other comment probably good to add a test in for it for reference.

library/tinkerpop/pom.xml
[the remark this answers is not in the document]
tb06904: Probably not critical enough for the enforcer plugin just a small comment is likely enough, the latest version is only 1.0.4 anyways and development has mostly stopped so we're not likely to update anytime soon.

GCHQDeveloper314 (approved):

library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/process/traversal/strategy/optimisation/GafferPopGraphStepStrategy.java
GCHQDeveloper314: Any info on this seems to be entirely absent [from their Javadocs](https://www.javadoc.io/doc/org.opencypher.gremlin/translation/1.0.0/index.html) unfortunately. At least the code handles it ok.

tb06904: The javadocs are unfortunate, its probably as the exception is a runtime one so isn't declared by the methods throwing it but there is at least some info about the exception class itself [here](https://www.javadoc.io/static/org.opencypher.gremlin/translation/1.0.0/org/opencypher/gremlin/translation/exception/SyntaxException.html).

p29876 (approved):

PR #3230, opened 2024-06-06
Gh-3223: Change GafferPopEdge id representation

### LINKED ISSUE #3223: Change GafferPopEdge id representation
Edge ids are currently represented as

```
Arrays.asList(getVertexId(outVertex), getVertexId(inVertex)) 
```

The List representation is causing issues when used as a GraphStep seed, this representation should be changed to a String.

Should also consider including the edge label as part of the ID to force them to be unique in the case that multiple edges exist between that same 2 vertices, e.g. something like the Tinkerpop representation "[source-label->target]"

As part of this ticket, he GafferPopGraph.getElementSeeds method should be improved.
Note: this method should create EdgeSeeds from inputs with and without the label

e.g.
```
g.E("[1-knows->2]")
g.E("[1,2]") 
```

cn337131:
Able to now query edges with the id in format [1, knows, 2] in addition to [1, 2]. If the label is included this is used in the GetElements View.

Label also added to the returned edge id.

# Related issue

- Resolve #3223

### DIFF
--- a/library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/GafferPopGraph.java
+++ b/library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/GafferPopGraph.java
@@ -1009,7 +1009,7 @@ public class GafferPopGraph implements org.apache.tinkerpop.gremlin.structure.Gr
                 List<String> idElements = parseString((String) id);
 
                 // If contains label, extract to use in View as edge group
-                if(idElements.size() == 3){
+                if (idElements.size() == 3) {
                     labels.add(idElements.get(1));
                 } else {
                     // Fallback is to use all schema edge groups
@@ -1027,7 +1027,7 @@ public class GafferPopGraph implements org.apache.tinkerpop.gremlin.structure.Gr
      * @param id The id that is to be parsed
      * @return The edge ids and any labels
      */
-    private List<String> parseString(String id) {
+    private List<String> parseString(final String id) {
         List<String> idElements = new ArrayList<>();
 
         String regex1 = "^\\[(?<first>.*),\\s*(?<middle>[a-zA-Z0-9|-]*)?\\s*,\\s*(?<last>.*)\\]$";
@@ -1037,18 +1037,18 @@ public class GafferPopGraph implements org.apache.tinkerpop.gremlin.structure.Gr
         Pattern pattern2 = Pattern.compile(regex2);
         Matcher matcher2 = pattern2.matcher(id);
 
-        if(matcher1.find()){
+        if (matcher1.find()) {
             // Add source
             idElements.add(matcher1.group("first"));
 
             // Add label
-            if(matcher1.group("middle") != null){
+            if (matcher1.group("middle") != null) {
                 idElements.add(matcher1.group("middle"));
             }
 
             // Add dest
             idElements.add(matcher1.group("last"));
-        } else if (matcher2.find()){
+        } else if (matcher2.find()) {
             // Add source
             idElements.add(matcher2.group("first"));
 
--- a/library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/GafferPopGraph.java
+++ b/library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/GafferPopGraph.java
@@ -249,7 +249,7 @@ public class GafferPopGraph implements org.apache.tinkerpop.gremlin.structure.Gr
 
     private static final Logger LOGGER = LoggerFactory.getLogger(GafferPopGraph.class);
     private static final String GET_ALL_DEBUG_MSG = "Requested a GetAllElements, results will be truncated to: {}.";
-    private static final Pattern EDGE_ID_REGEX = Pattern.compile("^\\[(?<src>[a-zA-Z0-9|-]*)(,\\s*(?<label>[a-zA-Z0-9|-]*))?\\s*,\\s*(?<dest>[a-zA-Z0-9|-]*)\\]$");
+    private static final Pattern EDGE_ID_REGEX = Pattern.compile("^\\[\\s*(?<src>[a-zA-Z0-9|-]*)\\s*(,\\s*(?<label>[a-zA-Z0-9|-]*))?\\s*,\\s*(?<dest>[a-zA-Z0-9|-]*)\\s*\\]$");
 
     public GafferPopGraph(final Configuration configuration) {
         this(configuration, createGraph(configuration));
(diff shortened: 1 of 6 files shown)


library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/GafferPopGraph.java
```diff
@@ -971,27 +970,59 @@ private List<ElementSeed> getElementSeeds(final Iterable<Object> ids) {
...
+    private Set<String> getEdgeViewGroup(final Iterable<Object> ids) {
+        Set<String> labels = new HashSet<>();
+
+        ids.forEach(id -> {
+            if ((id instanceof String) && ((String) id).matches("^\\[.*,[a-zA-Z0-9]*,.*\\]$")) {
+                List<String> edgeIdList = Arrays.asList(((String) id)
+                        .replaceAll("\\s", "")
+                        .replace("[", "")
+                        .replace("]", "")
+                        .split(","));
```
p29876, line 1015: Same thing here regarding regex and capture groups
Can we reduce the duplication?

library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/GafferPopGraph.java
```diff
@@ -637,12 +637,12 @@ public Iterator<Edge> edges(final Object id, final Direction direction, final St
...
      * Supports input as a {@link Vertex}, {@link Edge}, List of Edge IDs or individual Vertex IDs.
      * @param direction {@link Direction} of edges to return.
-     * @param labels    labels of edges. Alternatively you can supply a Gaffer View serialised into JSON.
+     * @param labels labels of edges. Alternatively you can supply a Gaffer View serialised into JSON.
      * @return iterator of {@link GafferPopEdge}s.
      * @see #edges(Object...)
      */
     public Iterator<Edge> edges(final Iterable<Object> ids, final Direction direction, final String... labels) {
-        return edgesWithView(ids, direction, createView(labels));
+        return edgesWithView(getElementSeeds(ids), direction, createView(labels));
```
p29876, line 645: Is this needed? I think `edgesWithView` calls `getElementSeeds` so I think we're doing it twice here

library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/GafferPopGraph.java
```diff
@@ -971,27 +972,97 @@ private List<ElementSeed> getElementSeeds(final Iterable<Object> ids) {
...
-                edgeIdList = Arrays.asList(((String) id)
-                        .replaceAll("\\s", "")
-                        .replace("[", "")
-                        .replace("]", "")
-                        .split(","));
-            // Assume entity ID as fallback
+            // Attempt to extract source and destination IDs from a string form of an
+            // array/list
+            } else if ((id instanceof String) && (((String) id).matches("^\\[.*,.*\\]$")
+                || ((String) id).matches("^\\[.*,[a-zA-Z0-9|-]*,.*\\]$"))) {
```
p29876, line 978: We could just use one regex for both here
```java
// Matches [src,dest] and [src,label,dest]
EDGE_ID_REGEX = "^\\[(?<src>[a-zA-Z0-9|-]*)(,\\s*(?<label>[a-zA-Z0-9|-]*))?\\s*,\\s*(?<dest>[a-zA-Z0-9|-]*)\\]$";`
```
allowing us to just do something like
```java
if (matcher.matches()) {
  seeds.add(new EdgeSeed(matcher.group("src"), matcher.group("dest")))
}
```
so we can get rid of the `edgeIdList` size checks and the `parseString` method

library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/GafferPopGraph.java
```diff
@@ -971,25 +973,49 @@ private List<ElementSeed> getElementSeeds(final Iterable<Object> ids) {
...
                 ((Iterable<?>) id).forEach(edgeIdList::add);
-            // Attempt to extract source and destination IDs from a string form of an array/list
-            } else if ((id instanceof String) && (((String) id).matches("^\\[.*,.*\\]$"))) {
-                edgeIdList = Arrays.asList(((String) id)
-                        .replaceAll("\\s", "")
-                        .replace("[", "")
-                        .replace("]", "")
-                        .split(","));
-            // Assume entity ID as fallback
+            // Attempt to extract source and destination IDs from a string form of an
```
p29876, line 976:
```suggestion
            // Attempt to extract source and destination IDs from a string from of an
```

library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/GafferPopGraph.java
```diff
@@ -971,25 +973,49 @@ private List<ElementSeed> getElementSeeds(final Iterable<Object> ids) {
...
+            // array/list
+            } else if ((id instanceof String) && (EDGE_ID_REGEX.matcher((String) id).matches())) {
+                Matcher edgeIDMatcher = EDGE_ID_REGEX.matcher((String) id);
+
+                if (edgeIDMatcher.matches()) {
```
tb06904, line 981: You've already checked it matches in the `if` on line 978 you probably don't need to do it again.

library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/GafferPopGraph.java
```diff
@@ -971,25 +973,49 @@ private List<ElementSeed> getElementSeeds(final Iterable<Object> ids) {
...
+     */
+    private Set<String> getEdgeViewGroup(final Iterable<Object> ids) {
+        Set<String> labels = new HashSet<>();
+
+        ids.forEach(id -> {
+            if ((id instanceof String) && (EDGE_ID_REGEX.matcher((String) id).matches())) {
+                Matcher edgeIdWithLabelMatcher = EDGE_ID_REGEX.matcher((String) id);
+
+                // If contains label, extract to use in View as edge group
+                if (edgeIdWithLabelMatcher.matches() && edgeIdWithLabelMatcher.group("label") != null) {
```
tb06904, line 1009: Already checked it matches in the other `if`

library/tinkerpop/src/test/java/uk/gov/gchq/gaffer/tinkerpop/GafferPopFederatedIT.java
```diff
@@ -249,13 +249,26 @@ void shouldGetEdgesById() {
...
                 .containsExactly(CREATED_EDGE_GROUP, CREATED_EDGE_GROUP);
     }
 
+    @Test
+    void shouldGetEdgesByIdAndLabel() {
+        // Given
+        GraphTraversalSource g = gafferPopGraph.traversal();
+
+        // When
+        List<Edge> result = g.E("[p1, created, s1]").toList();
```
tb06904, line 265: Maybe want to do some testing with some varying white space etc. just to better stress the regex

p29876 (dismissed):

p29876 (approved):

rj77259 (approved):

PR #3235, opened 2024-06-10
Gh-3234: TypeSubTypeValue Seeded Query Support in GafferPop

### LINKED ISSUE #3234: Support TSTV seeded queries in GafferPop
`TypeSubTypeValue` is a data type that can be used for IDs in Gaffer. We need to ensure that users can use TSTVs via Gremlin using GafferPop.

**Additional context**
The simplest way we could allow it would be treating them as String IDs in the form `myType|mySubType|myValue` and then parse them as `TypeSubTypeValues` once they've got through the server layer. Essentially so a user could do a seeded Gremlin query like: `g.V("myType|mySubType|myValue")`.

tb06904:
Adds basic support for seeding queries in GafferPop with TypeSubTypeValues by parsing specific string formats separated via a `|` to construct the correct object to seed the query with. 

As an example `g.V("t:type|st:sub|v:value")` -> `new TypeSubTypeValue("type", "sub", "value")`

# Related issue

- Resolve #3234

### DIFF
--- a/library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/GafferPopGraph.java
+++ b/library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/GafferPopGraph.java
@@ -981,15 +981,13 @@ public class GafferPopGraph implements org.apache.tinkerpop.gremlin.structure.Gr
                         .split(","));
             // Assume entity ID as fallback
             } else {
-                seeds.add(new EntitySeed((getTSTVFromID(id) != null) ? getTSTVFromID(id) : id));
+                seeds.add(new EntitySeed(getIDAsRelevantType(id)));
             }
 
             // If found a list verify source and destination
             if (edgeIdList.size() == 2) {
-                Object source = (getTSTVFromID(edgeIdList.get(0)) != null) ?
-                    getTSTVFromID(edgeIdList.get(0)) : edgeIdList.get(0);
-                Object dest = (getTSTVFromID(edgeIdList.get(1)) != null) ?
-                    getTSTVFromID(edgeIdList.get(1)) : edgeIdList.get(1);
+                Object source = getIDAsRelevantType(edgeIdList.get(0));
+                Object dest = getIDAsRelevantType(edgeIdList.get(1));
                 seeds.add(new EdgeSeed(source, dest));
             }
         });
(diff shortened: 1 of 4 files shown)


library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/GafferPopGraph.java
```diff
@@ -980,18 +981,40 @@ private List<ElementSeed> getElementSeeds(final Iterable<Object> ids) {
...
+     * @param id The ID.
+     * @return The ID as its relevant type.
+     */
+    private Object getIDAsRelevantType(final Object id) {
+        if (id instanceof String) {
+            // See if can split into a TSTV
+            String[] split = ((String) id).split("\\|");
+            if (split.length == 3) {
+                LOGGER.debug("Parsing ID as a TSTV: {}", id);
+                return new TypeSubTypeValue(split[0], split[1], split[2]);
```
p29876, line 1012: May be worth changing to regex to bring it in line style-wise with #3230 
But this is still very readable so happy either way

p29876 (approved):

library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/GafferPopGraph.java
p29876: assuming we don't have to handle any whitespace here?

cn337131 (dismissed):

library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/GafferPopGraph.java
tb06904: Updated to use a regex instead with a slightly tweaked format. I don't think we really want to handle white space as we can't know if that is meant to be part of the type, subtype or value so safer to take whats between the `|` delimiter as literal.

p29876 (dismissed):

library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/GafferPopGraph.java
```diff
@@ -245,6 +248,7 @@ public enum DefaultIdManager {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(GafferPopGraph.class);
     private static final String GET_ALL_DEBUG_MSG = "Requested a GetAllElements, results will be truncated to: {}.";
+    private static final Pattern TSTV_REGEX =  Pattern.compile("^tstv:(?<type>[A-Za-z0-9-]*)\\|(?<stype>[A-Za-z0-9-]*)\\|(?<val>[A-Za-z0-9-]*)$");
```
cn337131, line 251: just to check that TSTV's can't contain a pipe in the type/subtype/value ?

tb06904: Updated the regex match to allow seeding like `g.V("t:thetype|st:thesubstype|v:theval")` this should hopefully avoid majority of conflicts

cn337131 (approved):

p29876 (approved):

PR #3251, opened 2024-07-15
Gh-3250: GetElementsWithinSet bug fix

### LINKED ISSUE #3250: Bug when using GetElementsWithinSet operation
**Describe the bug**
When a large input seed list is passed to GetElementsWithinSet not all the expected edges are returned. This is particularly evident when the input size is greater than the max entries for the batch scanner.

**To Reproduce**
Set the maxEntriesForBatchScanner to less than the number of input seeds. 
Run a GetElementsWithinSet with the input seeds.

**Expected behaviour**
Some edges will be missing. The edges are missing where the destination value is greater than the batch scanner entry value. 

**Additional context**
It appears as though the batch is effectively setting includeIncomingOutGoing to OUTGOING.

cn337131:
GetElementsWithinSet did not implement the SeededGraphFilters interface which was leading to edges getting lost when the number of entries for the batch scanner was less than the input size as the range was defaulting to outgoing edges only. This is fixed by implementing the interface, this is also in-line with all other Accumulo specific operations.

# Related issue

- Resolve #3250


store-implementation/federated-store/src/test/java/uk/gov/gchq/gaffer/federatedstore/FederatedStoreGetElementsWithinSetTest.java
```diff
@@ -0,0 +1,273 @@
...
+
+    private Graph federatedGraph;
+
+    @AfterAll
+    public static void tearDownCache() {
+        resetForFederatedTests();
+    }
+
+    @BeforeEach
+    public void setUp() throws Exception {
```
tb06904, line 77: Don't need public modifiers on test methods/classes

store-implementation/federated-store/src/test/java/uk/gov/gchq/gaffer/federatedstore/FederatedStoreGetElementsWithinSetTest.java
```diff
@@ -0,0 +1,273 @@
...
+        addEdges("A244", "A87", GRAPH_ID_A);
+        addEdges("A168", "A110", GRAPH_ID_A);
+        addEdges("A56", "A299", GRAPH_ID_A);
+        addEdges("A297", "A193", GRAPH_ID_B);
+        addEdges("A15", "A285", GRAPH_ID_B);
+        addEdges("A1", "A52", GRAPH_ID_B);
+    }
+
+    @Test
+    public void shouldReturnOnlyEdgesWhenViewContainsNoEntitiesFromSubGraphsWithFederatedOperation() throws Exception {
```
tb06904, line 118: Not major but you can probably leave out the `FromSubGraphsWithFederatedOperation` from these test names you're already in a federated test class its probably implied.

p29876 (approved):

tb06904 (approved):

PR #3255, opened 2024-07-19
Gh-3253 Fix custom types in gremlin

### LINKED ISSUE #3253: TSTV can't deserialise when using a Gremlin connection
**Describe the bug**
There is support for type sub type values with Gremlin, however when you use this feature from a Gremlin server or via the integrated Gremlin endpoint in the rest API it cannot deserialise the result.

This is partly an issue because Gremlin will add a type tag to all returned datatypes and client end may not know what a Type subtype value is so can't deserialse it.

**Additional context**
Possible fixes could be to just make sure any custom data types are serialised or converted to Strings before serialising the main full response and sending back the response so they are already in JSON for the client to consume.

This will likely only be doable with the Gremlin endpoint in the REST API as that will have the opportunity to transform the response.

tb06904:
Fixes the usage of custom Gaffer types not being able to be handled by a Gremlin client. 

Works by simply making sure whenever conversion from Gaffer element to Tinkerpop element any custom types are converted to string representation and vice versa so that results can be parsed fine by a Gremlin client or Gaffer. This does mean slight overhead as conversions will happen even if Gremlin does not return the step result to the client but there is no way round this as each Gremlin step in theory can be returned so always needs any custom types converted.

Main type that is always converted is the Gaffer `TypeSubTypeValue` but more could be added, the string format has also been changed to be the same as the `.toString` so can more easily pass back in results, but a slight short hand is also accepted in Gremlin queries like `[type=thetype,subType=thesubtype,value=thevalue]`

# Related issue

- Resolve #3253


p29876:
All looks fine, just need to remember to update the docs with the new TSTV String format.

Am I right in thinking this means you can no longer get a TSTV object from a GafferPop graph, you'll only ever get the String in case you're querying from the server?
If so, assuming actual TSTV objects is something users would want, should we add a helper method to create a TSTV from the String representation to save them from parsing it themselves on the other end? e.g. a new constructor or a static `TypeSubTypeValue.from(tstvString)`

p29876:
> All looks fine, just need to remember to update the docs with the new TSTV String format.
> 
> Am I right in thinking this means you can no longer get a TSTV object from a GafferPop graph, you'll only ever get the String in case you're querying from the server? If so, assuming actual TSTV objects is something users would want, should we add a helper method to create a TSTV from the String representation to save them from parsing it themselves on the other end? e.g. a new constructor or a static `TypeSubTypeValue.from(tstvString)`

Ignore me, you can do that with the new custom types factory
I was thinking we could have something more concrete that would return a TSTV or throw if invalid or something but it would just be a duplication of this method and it's probably fine as is

p29876 (dismissed):

tb06904:
> All looks fine, just need to remember to update the docs with the new TSTV String format.
> 
> Am I right in thinking this means you can no longer get a TSTV object from a GafferPop graph, you'll only ever get the String in case you're querying from the server? If so, assuming actual TSTV objects is something users would want, should we add a helper method to create a TSTV from the String representation to save them from parsing it themselves on the other end? e.g. a new constructor or a static `TypeSubTypeValue.from(tstvString)`

You are right basically, if the results are returned as Tinkerpop a `Vertex` or `Edge` it will have any custom types made safe as strings, you can always pass these back through a Gaffer element generator to make them into a Gaffer `Entity` or `Edge` though which will convert the types back or use the static util method.

cn337131 (approved):

p29876 (approved):

cn337131 (approved):

PR #3257, opened 2024-07-23
Gh-3256: Dependency updates

tb06904:
Various dependency updates for Java 11


pom.xml
GCHQDeveloper314, line 1: With the default Java version set to 11, the `java11` profile defined on line 1703 of this PR is redundant and should be removed. It activates automatically when a JDK 11 environment is found and will now conflict with the defaults.

GCHQDeveloper314:
The "CodeQL test" action is failing because it is still using Java 8 (with sonarcloud now being used this action could be removed - something for a new issue).
I'm not sure why you've changed `StoreTest`, but that is the cause of the other error.

PR #3263, opened 2024-08-01
Gh-3254: Standardise Spring REST user authentication

### LINKED ISSUE #3254: Standarise the User auth in Spring REST
As the spring REST should be our standard API for deployment the user authentication for it needs standardising as the current user factory class is not adequate.

Current problems are that the user factory has no way of passing HTTP headers needed for proper authorisation other than via a javax context which is not used in spring boot. The general idea with current method is that implementations of user factories can be used to 'create' the user e.g. based on the HTTP headers.

**Additional context**
Some work to address this has already been started under the GafferPop integration work in #3248 this should be expanded on so all endpoints that require it have proper user authentication hooks.

tb06904:
Some standarisation to the spring boot based REST API that enures any endpoints that execute operations on the graph have the HTTP headers correctly extracted from the request and set in the user factory before user creation. 

Took the liberty to simplify the controller classes too removing the redundant controller interfaces which are not required and is really not recommended in spring boot. This means all the controller classes are fairly consistent now.

# Related issue

- Resolve #3254


p29876 (approved):

rest-api/spring-rest/src/main/java/uk/gov/gchq/gaffer/rest/controller/JobController.java
```diff
@@ -63,8 +78,12 @@ public ResponseEntity<JobDetail> startJob(@RequestBody final Operation operation
                 .body(jobDetail);
     }
 
-    @Override
-    public ResponseEntity<JobDetail> scheduleJob(@RequestBody final Job job) throws OperationException {
+    @PostMapping(path = "/schedule", consumes = APPLICATION_JSON_VALUE, produces = APPLICATION_JSON_VALUE)
+    @io.swagger.v3.oas.annotations.Operation(summary = "schedules an asynchronous job")
```
cn337131, line 82:
```suggestion
    @io.swagger.v3.oas.annotations.Operation(summary = "Schedules an asynchronous job")
```

cn337131 (approved):

p29876 (approved):

PR #3269, opened 2024-08-20
Gh 2839 track creation date of sub graphs

### LINKED ISSUE #2839: Track creation date of sub graphs
The federated store should be able to track the creation date of new sub graphs

So old graphs can be aged off

**Additional context**
For implementation, a date field could be added to [GraphSerialisable](https://github.com/gchq/Gaffer/blob/6ce0a4e3fceb097943d00d53aeacad1e59a310c5/core/graph/src/main/java/uk/gov/gchq/gaffer/graph/GraphSerialisable.java#L50) that gets populated as current datetime when created

wb36499:
- Adds /config/graphCreatedTime endpoint to retrieve graph creation timestamp associated with REST API
- Adds operation to retrieve list of all graphIds & associated creation timestamps. In a federated store this will return all sub-graph ids and creations times
- Adds custom property to Accumulostore on table creation to save graph creation time
- Saves MapStore based graph creation times as Store variable on build 

# Related issue

- Resolve #2839


store-implementation/accumulo-store/src/test/java/uk/gov/gchq/gaffer/accumulostore/AccumuloStoreTest.java
```diff
@@ -580,4 +582,48 @@ void shouldDoNothingWhenCannotClose() throws Exception {
...
+        final AccumuloStore store = new MiniAccumuloStore();
+        store.initialise("graphId", SCHEMA, PROPERTIES);
+        GafferRuntimeException expectedException = new GafferRuntimeException("Error getting timestamp.");
+
+        // When
+        store.getConnection().tableOperations().delete("graphId");
+        GafferRuntimeException exception = assertThrows(GafferRuntimeException.class, () -> store.getCreatedTime());
+
+        // Then
+        assertThat(exception.getMessage()).isEqualTo(expectedException.getMessage());
```
cn337131, line 612: you could simplify this by removing lines 605 and 609 and changing the test to simply be 
`assertThatExceptionOfType(GafferRuntimeException.class).isThrownBy(() -> store.getCreatedTime()).withMessage("Error getting timestamp"); `

This could then be repeated on the test below :)

core/store/src/main/java/uk/gov/gchq/gaffer/store/Store.java
```diff
@@ -212,6 +215,7 @@ public abstract class Store {
     protected final OperationChainValidator opChainValidator;
     private final SchemaOptimiser schemaOptimiser;
     private final Boolean addCoreOpHandlers;
+    private LocalDateTime createdTime = LocalDateTime.now();
```
p29876, line 218: Do we want to initialise the time here if `Graph` is meant to use `store.setCreatedTime`

p29876: ignore that
Do we want `createdTime` to be mutable? should we have a public `setCreatedTime` at all or should it be the responsibility of the store to set it's created time once when it is initialised

core/store/src/main/java/uk/gov/gchq/gaffer/store/Store.java
```diff
@@ -312,6 +316,22 @@ public void initialise(final String graphId, final Schema schema, final StorePro
...
+     */
+    public String getCreatedTime() {
+        return createdTime.toString();
+    }
+    /**
+     * Sets this Store's creation timestamp.
+     *
+     * @param createdTime the LocalDateTime to store as the graph creation date.
+     */
+    public void setCreatedTime(final LocalDateTime createdTime) {
```
p29876, line 332: As above, just check if we want to be able to do this. `createdTime` feels like it should be immutable but I'm not sure

core/store/src/main/java/uk/gov/gchq/gaffer/store/Store.java
wb36499: You make a good point - I have simplified the `createdTime` setting to once the `Store` initialised and remove from the `Graph`

cn337131 (approved):

tb06904 (approved):

Issue #3277, opened 2024-09-03
GafferPop should 'squash' GetElements Operations


p29876:
At the moment if you run a query like g.V('1').outE(),  for every edge on the vertex '1' GafferPop will run a GetElements operation with only one seed. This can be very problematic for queries that touch a lot of data as each operation has overhead and gets logged so could end up with thousands of operations being called at once.

To fix this we might need to implement custom vertex/edge traversal strategies so that all the seeds are acquired first then added to one big operation rather than lots of smaller ones.

PR #3278, opened 2024-09-03
Gh-3277: Add custom GafferPop Vertex Step

p29876:
Add a fold step before each vertex step to collect all vertex ids so the input to the new VertexStep is a list rather than single ids. This means we can do a single Gaffer query for all ids.

Note: input seeds are essentially deduped so expected output may change

### DIFF
--- a/library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/process/traversal/strategy/optimisation/GafferPopVertexStepStrategy.java
+++ b/library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/process/traversal/strategy/optimisation/GafferPopVertexStepStrategy.java
@@ -1,0 +1,57 @@
+/*
+ * Copyright 2024 Crown Copyright
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package uk.gov.gchq.gaffer.tinkerpop.process.traversal.strategy.optimisation;
+
+import org.apache.tinkerpop.gremlin.process.traversal.Traversal.Admin;
+import org.apache.tinkerpop.gremlin.process.traversal.TraversalStrategy;
+import org.apache.tinkerpop.gremlin.process.traversal.step.map.FoldStep;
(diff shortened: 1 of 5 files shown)


library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/process/traversal/strategy/optimisation/GafferPopVertexStepStrategy.java
```diff
@@ -0,0 +1,69 @@
...
+    public void apply(final Admin<?, ?> traversal) {
+        TraversalHelper.getStepsOfClass(VertexStep.class, traversal).forEach(originalVertexStep -> {
+            LOGGER.debug("Inserting FoldStep and replacing VertexStep");
+
+            // Replace vertex step
+            final GafferPopListVertexStep<? extends Element> listVertexStep = new GafferPopListVertexStep<>(
+                    originalVertexStep);
+            TraversalHelper.replaceStep(originalVertexStep, listVertexStep, traversal);
+
+            // Add in a fold step before the new VertexStep so that the input is the list of
```
tb06904, line 58: Random question but could the input be an iterable rather than a list? Just thinking if this can be made slightly lazy at all for if the step as a lot of vertexes returned?
I guess it will have to be made into a list at some point for the actual `GetElement` operation though.

tb06904 (dismissed):

cn337131 (approved):

wb36499 (dismissed):

library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/process/traversal/step/LazyFoldStep.java
```diff
@@ -0,0 +1,66 @@
...
+
+/**
+ * A Lazy implementation of the FoldStep
+ * Folds to a lazy iterable
+ */
+public class LazyFoldStep<S> extends AbstractStep<S, Iterable<S>> {
+
+    private boolean done = false;
+
+    public LazyFoldStep(final Traversal.Admin traversal) {
```
tb06904, line 35: Can probably not just use the raw type here `Traversal.Admin<S, Iterable<S>>` should work.

library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/process/traversal/step/GafferPopListVertexStep.java
```diff
@@ -0,0 +1,186 @@
...
+ * is returned.
+ * <p>
+ * e.g. for the Tinkerpop Modern Graph:
+ *
+ * <pre>
+ * (Gremlin)   g.V().out() = [v2, v3, v3, v3, v4, v5]
+ * (GafferPop) g.V().out() = [v2, v3, v4, v5]
+ * </pre>
+ */
+public class GafferPopListVertexStep<E extends Element> extends FlatMapStep<Iterable<Vertex>, E>
```
tb06904, line 65: Is `ListVertexStep` still the best name?

cn337131 (approved):

tb06904 (approved):

PR #3286, opened 2024-09-10
Gh-3276: Add ability to seed with Vertex only in Gremlin

### LINKED ISSUE #3276: Gremlin does not return vertex without Gaffer entity
**Describe the bug**
In Gaffer you can have an edge that has a SRC and DEST where the vertex does not have an actual entity associated with it in the Graph itself. The vertex still exists on the edge but there is no actual Entity. Currently in Gremlin this means if you seed with that vertex (g.V("vertex")) then nothing is returned. In standard Gaffer the associated edge should be returned and able to be used. 

**To Reproduce**
Steps to reproduce the behaviour:
1. Create an edge with no entity associated. e.g. a -> b where b is not an entity
2. Seed a query with b in Gremlin - e.g. g.V("b) 
3. Find that nothing is returned.

**Expected behaviour**
Expect the graph to be able to use the vertex as a seed and to be able to traverse the edge.

cn337131:
This should allow users to seed a Gremlin query with a vertex that exists only on an edge and not as an individual entity. 

# Related issue

- Resolve #3276


library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/process/traversal/util/GafferVertexFromEdgeFactory.java
```diff
@@ -0,0 +1,69 @@
...
+            .map(e -> ((Entity) e).getVertex())
+            .anyMatch(e -> Arrays.asList(vertexIds).contains(e));
+    }
+
+    private static Iterable<Vertex> extractSeedMatchFromEdges(final Iterable<? extends Element> result, final GafferPopGraph graph, final Object... vertexIds) {
+        return StreamSupport.stream(result.spliterator(), false)
+            .filter(Edge.class::isInstance)
+            .map(e -> ((Edge) e))
+            .map(e -> {
+                if (Arrays.asList(vertexIds).contains(e.getDestination())) {
```
tb06904, line 58: Won't this potentially be converting vertexes with an entity too? It looks like its not checking if the vertexId is already associated with a returned entity. The logic probably needs to be to gather a list of vertex Ids that don't already have an entity associated with them and use that as the input to this.

library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/process/traversal/util/GafferVertexFromEdgeFactory.java
```diff
@@ -0,0 +1,69 @@
...
+            nonEntityVertices = extractSeedMatchFromEdges(result, graph, vertexIds);
+        }
+        return nonEntityVertices;
+    }
+
+    private static boolean checkIfSeedInEntityResults(final Iterable<? extends Element> result, final Object... vertexIds) {
+        return StreamSupport.stream(result.spliterator(), false)
+            .filter(Entity.class::isInstance)
+            .map(e -> ((Entity) e).getVertex())
+            .anyMatch(e -> Arrays.asList(vertexIds).contains(e));
```
p29876, line 50: I'm not sure this will work when we seed with a correct seed and an orphan seed
e.g. g.V("7", "1")

library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/process/traversal/util/GafferVertexFromEdgeFactory.java
```diff
@@ -0,0 +1,69 @@
...
+import uk.gov.gchq.gaffer.tinkerpop.GafferPopGraph;
+import uk.gov.gchq.gaffer.tinkerpop.GafferPopVertex;
+
+import java.util.Arrays;
+import java.util.Objects;
+import java.util.stream.Collectors;
+import java.util.stream.StreamSupport;
+
+public final class GafferVertexFromEdgeFactory {
+    private static Iterable<Vertex> nonEntityVertices;
```
p29876, line 33: Do we need this as global?
If you ran checkForSeed twice and the first time there were results and the second there weren't you could still get results returned from the previous run

library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/process/traversal/util/GafferVertexFromEdgeFactory.java
```diff
@@ -0,0 +1,69 @@
...
+
+    private GafferVertexFromEdgeFactory() {
+        // Utility class
+    }
+
+    public static Iterable<Vertex> checkForSeed(final Iterable<? extends Element> result, final GafferPopGraph graph, final Object... vertexIds) {
+        if (!checkIfSeedInEntityResults(result, vertexIds)) {
+            nonEntityVertices = extractSeedMatchFromEdges(result, graph, vertexIds);
+        }
+        return nonEntityVertices;
```
p29876, line 43: can this be `Collections.emptyList()`?

tb06904 (dismissed):

library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/process/traversal/util/GafferVertexUtils.java
```diff
@@ -0,0 +1,89 @@
...
+     */
+
+    public static Iterable<Vertex> getOrphanVertices(final Iterable<? extends Element> result, final GafferPopGraph graph, final Object... vertexIds) {
+        // Check for any vertex ID seeds that are not returned as Entities
+        List<Object> orphanVertexIds = Arrays.stream(vertexIds)
+            .filter(id -> StreamSupport.stream(result.spliterator(), false)
+                .filter(Entity.class::isInstance)
+                .map(e -> ((Entity) e).getVertex())
+                .noneMatch(e -> e.equals(id)))
+                .collect(Collectors.toList());
```
p29876, line 59: Can we un-indent this line one level to make it really clear which stream it's collecting?

library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/process/traversal/util/GafferVertexUtils.java
```diff
@@ -0,0 +1,89 @@
...
+
+        return (orphanVertexIds != null) ? extractOrphanVerticesFromEdges(result, graph, orphanVertexIds) : Collections.emptyList();
```
p29876, line 61: Can this actually be null? not sure but might be worth add a check for if it's empty too so we don't have to iterate the results again

library/tinkerpop/src/main/java/uk/gov/gchq/gaffer/tinkerpop/process/traversal/util/GafferVertexUtils.java
```diff
@@ -0,0 +1,89 @@
...
+     * @param graph The GafferPop graph being queried
+     * @param vertexIds The vertexIds that have been used as seeds in the query
+     * @return Iterable of 'orphan' {@link Vertex}'s
+     */
+
+    public static Iterable<Vertex> getOrphanVertices(final Iterable<? extends Element> result, final GafferPopGraph graph, final Object... vertexIds) {
+        // Check for any vertex ID seeds that are not returned as Entities
+        List<Object> orphanVertexIds = Arrays.stream(vertexIds)
+            .filter(id -> StreamSupport.stream(result.spliterator(), false)
+                .filter(Entity.class::isInstance)
```
tb06904, line 56: Think these need to all be aligned

tb06904: Ahh nice I didn't spot that!

PR #3330, opened 2024-10-31
2478: JobTracker tracks all operations not just jobs

ms9698:
Changes Made:
Removed addOrUpdateJobDetail from the execute method. 
Non-job operations will no longer be tracked, preventing unnecessary performance degradation and cache pollution.
Removed test for checking if operation runs when jobtracker cache is broken.


core/store/src/main/java/uk/gov/gchq/gaffer/store/Store.java
```diff
@@ -385,14 +385,24 @@ public <O> O execute(final Output<O> operation, final Context context) throws Op
         return execute(OperationChain.wrap(operation), context);
     }
 
+    private boolean isJobOperation(Class<?> operationClass) {
```
tb06904, line 388: Not sure this is solving the issue, I think you want to basically remove all the job tracker stuff from the `execute` method so that it only updates things submitted to the `executeJob` method

ms9698: fixed

p29876:
Can you fix the code smell?

core/store/src/main/java/uk/gov/gchq/gaffer/store/Store.java
```diff
@@ -978,21 +970,19 @@ public OperationHandler<Operation> getOperationHandler(final Class<? extends Ope
         return operationHandlers.get(opClass);
     }
 
-    private JobDetail addOrUpdateJobDetail(final OperationChain<?> operationChain, final Context context,
-                                           final String msg, final JobStatus jobStatus) {
+    private JobDetail addOrUpdateJobDetail(final OperationChain<?> operationChain, final Context context, final String msg, final JobStatus jobStatus) {
         final JobDetail newJobDetail = new JobDetail(context.getJobId(), context.getUser(), operationChain, jobStatus, msg);
-        if (nonNull(jobTracker)) {
+        if (null != jobTracker) {
```
p29876, line 975: nitpick but we're transitioning away from yoda conditions so ideally the changes to this method would be reverted (which will also fix the code coverage issues), or reverse the conditions

core/store/src/test/java/uk/gov/gchq/gaffer/store/StoreTest.java
```diff
@@ -256,32 +249,6 @@ public void after() {
...
-        field.setAccessible(true);
-        java.util.Map<String, ICacheService> mockCacheServices = (java.util.Map<String, ICacheService>) field.get(new HashMap<>());
-        mockCacheServices.put(JOB_TRACKER_CACHE_SERVICE_NAME, mockICacheService);
-
-        final AddElements addElements = new AddElements();
-        final StoreImpl3 store = new StoreImpl3();
-        store.initialise("graphId", createSchemaMock(), storeProperties);
-
-        // When
-        store.execute(addElements, context);
```
p29876, line 277: This test is worth keeping as it tests job functionality
We no longer handle jobs in the `execute` method, but we do in the `executeJob` method.  
Swapping this line should pass the test

core/store/src/main/java/uk/gov/gchq/gaffer/store/Store.java
ms9698: reverted :)

rest-api/core-rest/src/test/java/uk/gov/gchq/gaffer/rest/service/v2/OperationServiceV2IT.java
```diff
@@ -106,24 +106,6 @@ public void shouldPropagateStatusInformationContainedInOperationExceptionsThrown
         assertEquals(SERVICE_UNAVAILABLE.getStatusCode(), response.getStatus());
     }
 
-    @Test
-    public void shouldReturnSameJobIdInHeaderAsGetAllJobDetailsOperation() throws IOException {
```
p29876, line 110: Agree that this test can be deleted now as it fails due to no jobs running which is correct.
However, I'm not a fan of just deleting tests with no explanation. It's better to change the test to document the new behavior as a bit of self documenting code. I think we need a new test here to show that we shouldn't get a job-id header returned anymore for the executeOperation endpoint as it doesn't make sense.

However, we are still getting the header returned so you will need to implement this

p29876 (changes requested):
Can we stop the execute endpoint from returning a job-id header as it no longer makes sense to return one?

PR #3336, opened 2024-11-19
2478 Jobtracker tracks all operations not just jobs

ms9698:


p29876:
original review https://github.com/gchq/Gaffer/pull/3330

PR #3343, opened 2024-12-02
Gh-3342: Handle different groups in View federated poc

### LINKED ISSUE #3342: Be able to handle different groups in View Federated POC
The simple federated POC should be able to handle a View with groups that would normally fail validation on sub graphs. Groups in a View might not be in every sub graph so need to be able to handle this use case.

tb06904:
Adds ability to handle groups in a View that might not be relevant to a selected sub graph.

# Related issue

- Resolve #3342

### DIFF
--- a/store-implementation/simple-federated-store/src/main/java/uk/gov/gchq/gaffer/federated/simple/FederatedUtils.java
+++ b/store-implementation/simple-federated-store/src/main/java/uk/gov/gchq/gaffer/federated/simple/FederatedUtils.java
@@ -157,7 +157,7 @@ public final class FederatedUtils {
                     graphSerialisable.getGraphId(),
                     FederatedOperationHandler.OPT_SKIP_FAILED_EXECUTE));
             }
-            return builder.build();
+            return newView;
         }
 
         // Nothing to do return unmodified view
--- a/store-implementation/simple-federated-store/src/main/java/uk/gov/gchq/gaffer/federated/simple/FederatedUtils.java
+++ b/store-implementation/simple-federated-store/src/main/java/uk/gov/gchq/gaffer/federated/simple/FederatedUtils.java
@@ -1,0 +1,137 @@
+/*
+ * Copyright 2024 Crown Copyright
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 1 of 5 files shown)


store-implementation/simple-federated-store/src/main/java/uk/gov/gchq/gaffer/federated/simple/FederatedUtils.java
```diff
@@ -0,0 +1,153 @@
...
+            ((OperationView) operation).setView(
+                getValidViewForGraph(((OperationView) operation).getView(), graphSerialisable));
+
+            updatedOperations.add(operation);
+
+        // Recursively go into operation chains to make sure everything is fixed
+        } else if (operation instanceof OperationChain) {
+            for (final Operation op : ((OperationChain<?>) operation).getOperations()) {
+                // Resolve if haven't hit the depth limit for validation
+                if (depth < 5) {
```
p29876, line 101: Shall we make this magic number a constant at least? or maybe configurable

store-implementation/simple-federated-store/src/main/java/uk/gov/gchq/gaffer/federated/simple/FederatedUtils.java
```diff
@@ -0,0 +1,153 @@
...
+        if (!validEntities.equals(view.getEntityGroups()) || !validEdges.equals(view.getEdgeGroups())) {
+            // Need to make changes to the view so start by cloning the view
+            // and clearing all the edges and entities
+            final View.Builder builder = new View.Builder()
+                    .merge(view)
+                    .entities(Collections.emptyMap())
+                    .edges(Collections.emptyMap());
+            validEntities.forEach(e -> builder.entity(e, view.getEntity(e)));
+            validEdges.forEach(e -> builder.edge(e, view.getEdge(e)));
+            return builder.build();
```
p29876, line 146: What happens here if the corrected view is empty? Can we end up in a situation where we forward an empty view onto a graph which may retrieve unexpected results? E.g. if I did GetAll 'created' edges from a federated graph of the Modern dataset, would the 'knows' graph receive an empty view and then return its knows edges?

store-implementation/simple-federated-store/src/main/java/uk/gov/gchq/gaffer/federated/simple/FederatedUtils.java
tb06904: Good point could make it an operation option with a default

store-implementation/simple-federated-store/src/main/java/uk/gov/gchq/gaffer/federated/simple/FederatedUtils.java
tb06904: That is a good point I don't think that situation was handled on old federation either I guess in that situation you would probably want to skip execution on that graph all together. Maybe could throw an exception to state that nothing specified in the View is relevant to that graph then the user can choose whether to resubmit the chain but with a skip on fail option or more targeted graph selection.

p29876 (approved):

cn337131 (approved):

PR #3347, opened 2025-01-06
Gh-2478: JobTracker tracks all operations, not just jobs

### LINKED ISSUE #2478: JobTracker tracks all operations, not just jobs
The documentation of Job, and the general title/JavaDoc of JobTracker indicates that, when enabled on a store, the JobTracker should track only Jobs.

However, inspection of Store.java (see https://github.com/gchq/Gaffer/blob/1fb6823082ef9dd88f65fab33ac1541e8a2a76cc/core/store/src/main/java/uk/gov/gchq/gaffer/store/Store.java#L389  and https://github.com/gchq/Gaffer/blob/1fb6823082ef9dd88f65fab33ac1541e8a2a76cc/core/store/src/main/java/uk/gov/gchq/gaffer/store/Store.java#L924) indicates that, when the JobTracker is enabled, any and every operation executed on the store is tracked as if it was a Job.

I suggest this is a defect, and that regular non-job operations should not be tracked by the JobTracker.  

If this _is_, in fact, intended behaviour, then I suggest the documentation and Javadoc should be updated, and maybe the name of JobTracker changed to OperationTracker, as I'd argue the existing behaviour is surprising given the name of the class and documentation.

Impacts of this defect include:

1. Performance degradation when running multiple fast operations caused by multiple writes to the cache for every operation
2. Performance degradation when instantiating a Store (see the initialise method in Store.java), as initialising a store with the job tracker enabled requires iterating through everything tracked by the JobTracker (even jobs for other stores)
3. In an environment where the JobTracker cache has a maximum size; early expiry of 'real' jobs as they are forced out of the cache early
4. Data being persisted to the cache that the user might not expect

As always, happy to submit a PR (I think the solution to this issue is quite straightforward) if the maintainers agree that this is indeed a defect.

Thanks,

P.S.  This issue has partially blocked the testing of gh-2457

cn337131:
Removes the job tracker from the main execute and any returned headers.

# Related issue

- Resolve #2478

### DIFF
--- a/rest-api/core-rest/src/test/java/uk/gov/gchq/gaffer/rest/service/v2/OperationServiceV2IT.java
+++ b/rest-api/core-rest/src/test/java/uk/gov/gchq/gaffer/rest/service/v2/OperationServiceV2IT.java
@@ -217,7 +217,7 @@ public class OperationServiceV2IT extends OperationServiceIT {
         // When
         final Response response = ((RestApiV2TestClient) client).executeOperationChainChunkedWithHeaders(opChain, "ListUser");
 
-        // Then]
+        // Then
         assertThat(response.getStatus()).isEqualTo(200);
     }
 
--- a/rest-api/core-rest/src/test/java/uk/gov/gchq/gaffer/rest/service/v2/OperationServiceV2IT.java
+++ b/rest-api/core-rest/src/test/java/uk/gov/gchq/gaffer/rest/service/v2/OperationServiceV2IT.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015-2021 Crown Copyright
+ * Copyright 2015-2025 Crown Copyright
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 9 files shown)


rest-api/core-rest/src/test/java/uk/gov/gchq/gaffer/rest/service/v2/OperationServiceV2IT.java
```diff
@@ -220,12 +217,12 @@ public void shouldAllowUserWithAuthThroughHeaders() throws IOException {
         // When
         final Response response = ((RestApiV2TestClient) client).executeOperationChainChunkedWithHeaders(opChain, "ListUser");
 
-        // Then
-        assertEquals(200, response.getStatus());
+        // Then]
```
tb06904, line 220: little typo
```suggestion
        // Then
```

p29876 (approved):

tb06904 (approved):

PR #3356, opened 2025-01-21
DeleteElements to return count

### LINKED ISSUE #3355: DeleteElements should return number of elements deleted
Currently, users have no feedback on what is deleted when they run a DeleteElements. It would probably be useful to log what is deleted and to also allow users to have some feedback on how many elements have been deleted in the operation.

cn337131:
DeleteElements will now return a count to reflect the number of elements deleted. Additional debug logging added as well to allow users to check what is deleted.

# Related issue

- Resolve #3355


store-implementation/accumulo-store/src/main/java/uk/gov/gchq/gaffer/accumulostore/operation/handler/DeleteElementsHandler.java
```diff
@@ -46,9 +54,19 @@ private void deleteElements(final DeleteElements operation, final AccumuloStore
...
+                elementCount++;
+            }
+            LOGGER.debug("Deleting elements: {}", elementsToDelete);
+
             store.deleteElements(validatedElements);
         } catch (final StoreException e) {
             throw new OperationException("Failed to delete elements", e);
         }
+
+        return elementCount;
```
tb06904, line 70: Can you not just return `elementsToDelete.size()`

tb06904 (approved):

p29876 (approved):

Issue #3350, opened 2025-01-09
Investigate optimisations for federated POC


tb06904:
**Describe the new feature you'd like**
Now the store is mostly feature complete can investigate if there are areas of optimisation.

PR #3357, opened 2025-01-21
Gh-3350: Optimisations for federated POC

tb06904:
Some minor optimisations and helpful functions for new federated store, changes include:
- Minimising stream API on small collections to save initialisation overhead.
- Using array/list data types where iteration performance is more important.
- Helpful operation to allow changing default graph IDs.
- Some tweaks hit the cache fewer times.
- Better shortcut logic on checks.


p29876:
Going to close this PR and split up the changes so we can progress
Some of the fixes are in #3358
The optimisations are to follow

